PR #30, opened 2022-11-12
Allow customizing the procfile handler and add example using overmind

kwbauson:
Hello! Super excited to see this project and I'm looking forward to following its development :)

I'm a big fan of [overmind](https://github.com/DarthSim/overmind) for process management. It runs processes with `tmux` and it allows features like ad-hoc restarting and killing of processes etc.

I've attempted to allow customizing the process runner and given an example of using overmind. However I'm not sure if I've chosen a good option name, and this does incur an extra evaluation when doing `devenv up`


domenkozar:
Nice! 

I'm hoping we can provide this out of the box, so people can just switch between different process implementations.

Something like `processes.implementation = "overman";`. 

``devenv up`` should then just execute ``config.processScript`` (and save it for a gc root) based on the option chosen.

domenkozar:
@kwbauson let me know if I can help further :)

kwbauson:
Sorry, I was out of town for a few days :) I couldn't figure out how to do `processes.implementation` since `processes` is already an option, so I went with `processesImplementation` instead

src/modules/top-level.nix
```diff
@@ -37,6 +37,13 @@ in
       description = "Processes can be started with ``devenv up`` and run in foreground mode.";
     };
 
+    processesImplementation = lib.mkOption {
```
domenkozar, line 40:
```suggestion
    process.implementation = lib.mkOption {
```

src/devenv.nix
```diff
@@ -78,9 +78,9 @@ pkgs.writeScriptBin "devenv" ''
         echo "No 'processes' option defined."  
         exit 1
       else
-        echo "Starting processes ..." 1>&2
-        echo "" 1>&2
-        ${pkgs.honcho}/bin/honcho start -f $procfile --env $procfileenv
+        procfilescript=$($CUSTOM_NIX/bin/nix $NIX_FLAGS build --no-link --print-out-paths --impure '.#procfileScript')
```
domenkozar, line 81: This should be done at line 73 and we don't need procfile & procfileenv anymore

domenkozar (changes requested):
Really close now! :)

src/devenv.nix
kwbauson: Is my latest push what you were thinking? I ended up having to remove the check for if processes were defined since it uses `$procfile`

src/flake.nix
```diff
@@ -45,9 +45,7 @@
       in {
         packages."${pkgs.system}" = {
           ci = pkgs.runCommand "ci" {} ("ls " + toString config.ci + " && touch $out");
-          procfile = config.procfile;
-          procfileEnv = config.procfileEnv;
-          info = config.info;
+          inherit (config) info procfileScript;
```
domenkozar, line 48: We need to keep these for backwards compatibility.

```suggestion
          inherit (config) info procfileScript procfileEnv procfile;
```

PR #33, opened 2022-11-13
Init Erlang language

ankhers:
Adding support for Erlang.

### DIFF
--- a/src/modules/languages/erlang.nix
+++ b/src/modules/languages/erlang.nix
@@ -1,0 +1,28 @@
+{ pkgs, config, lib, ... }:
+
+let
+  cfg = config.languages.erlang;
+in
+{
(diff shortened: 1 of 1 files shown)


src/modules/languages/erlang.nix
```diff
@@ -0,0 +1,28 @@
+{ pkgs, config, lib, ... }:
+
+let
+  cfg = config.languages.erlang;
+in
+{
+  options.languages.erlang = {
+    enable = lib.mkEnableOption "Enable tools for Erlang development.";
+
+    package = mkOption {
```
domenkozar, line 10: Needs `lib` and same for `types`

ankhers:
I will update #34 once I make sure this one works.

PR #56, opened 2022-11-14
init: support target directory argument

bobvanderlinden:
I started using devenv and intuitively used `devenv init myproject`, thinking it would create a new directory. It creates the files in my current directory.

This PR should add such support:

```console
$ devenv init myproject
$ cd myproject
```

I have not done full argument parsing/validation, as I wasn't sure you wanted to bother with that yet.


domenkozar:
Nice! line 174 should also list the new optional TARGET parameter :)

PR #63, opened 2022-11-15
java: add package options

bobvanderlinden:
Currently it is not possible to change the JDK package for Java.

When doing so it helps to run Maven and Gradle in the same JDK as defined. This is because Maven plugins may be run that sometimes need to run in the same JDK as your application. This is why it is needed to also allow redefining the Maven/Gradle packages. By default, the package is overridden to use the same JDK as the application.

Enable options were added for maven and gradle to avoid unnecessary rebuilding of these packages.

`JAVA_HOME` environment variable is defined. This helps IDEs and tools to pick the right JDK.

Lastly I took the liberty to remove the `enterShell`. It didn't seem useful to see the maven version specifically each time the shell is entered. The JDK version would be more useful.

If this is acceptable in a general sense, then I'd like to add some of the documentation to the different options.

### DIFF
--- a/src/modules/languages/java.nix
+++ b/src/modules/languages/java.nix
@@ -2,24 +2,46 @@
 
 let
   cfg = config.languages.java;
-  inherit (lib) types mkEnableOption mkOption mkDefault mkIf optional;
+  inherit (lib) types mkEnableOption mkOption mkDefault mkIf mdDoc optional literalExpression;
 in
 {
   options.languages.java = {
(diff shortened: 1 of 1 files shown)


domenkozar:
Seems like there's an unresolved rebase?

domenkozar (approved):

bobvanderlinden:
Should be good now. I added documentation. I also saw that gradle was overriding JAVA_HOME in its package as well, so I made sure gradle's JDK is also overridden by default.

PR #81, opened 2022-11-18
Add psql-devenv to postgres integration

JanLikar:


domenkozar:
Nice! We should just put this into the postgres module itself and call it something like `psql-devenv` as a script name.

JanLikar:
@domenkozar sure, that's even better.

src/modules/postgres.nix
```diff
@@ -61,6 +61,8 @@ in
 
     env.PGDATA = config.env.DEVENV_STATE + "/postgres";
 
+    scripts."psql-devenv".exec = "${cfg.package}/bin/psql -h $(realpath .devenv/state/postgres/) $@";
```
domenkozar, line 64: This should use $PGDATA

PR #83, opened 2022-11-18
Add background processes

JanLikar:
Use `devenv up -d` to run honcho with `nohup`.

`devenv down` and `devenv logs` are also included and are analog to their `docker-compose` counterparts.

### DIFF
--- a/src/devenv.nix
+++ b/src/devenv.nix
@@ -98,6 +98,9 @@ pkgs.writeScriptBin "devenv" ''
         echo "" 1>&2
         kill $(cat "$DEVENV_DIR/honcho.pid")
         rm "$DEVENV_DIR/honcho.pid"
+      else
+        echo "No running processes were found." 1>&2
+        echo "" 1>&2
       fi;
     ;;
     logs)
--- a/src/devenv.nix
+++ b/src/devenv.nix
@@ -77,12 +77,32 @@ pkgs.writeScriptBin "devenv" ''
       if [ "$(cat $procfile)" = "" ]; then
         echo "No 'processes' option defined."  
         exit 1
+      fi;
+
+      honcho_command="${pkgs.honcho}/bin/honcho start -f $procfile --env $procfileenv"
+
+      if [[ $# -eq 1 ]] && [[ $1 == "-d" ]]; then
+        echo "Starting background processes ..." 1>&2
+        echo "" 1>&2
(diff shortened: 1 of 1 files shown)


JanLikar:
This is first and foremost a PoC, so feel free to voice any concerns you might have, I'm more than happy to address them :)

src/devenv.nix
```diff
@@ -77,12 +77,32 @@ pkgs.writeScriptBin "devenv" ''
...
       fi
-      ;;
+    ;;
+    down)
+      if [ -r "$DEVENV_DIR/honcho.pid" ]; then
+        echo "Stopping processes ..." 1>&2
+        echo "" 1>&2
+        kill $(cat "$DEVENV_DIR/honcho.pid")
+        rm "$DEVENV_DIR/honcho.pid"
+      fi;
```
shyim, line 101: Maybe info that no background process could be found?

JanLikar: That's a good point, I will add it.

shyim (approved):

domenkozar:
I'm still not sure about this one. 

Could you explain what kind of workflow do you have that requires this interface?

shyim:
I would like to call that also in Gitpod to have one shell.

devenv up -d
devenv shell

and let the user work in the shell and don't have to care about the running services

JanLikar:
I guess so, yes

thenonameguy:
I recommend creating a prototype using https://github.com/NixOS/nixpkgs/blob/e66309ddcce8fc316b9addeabd11bf2187fa2eb9/pkgs/tools/system/daemonize/default.nix. 
This works by wrapping arbitrary programs and turns them into daemon processes.
Should be what we need without relying on the programs to do this themselves.

thenonameguy:
Started discussion about upstreaming this feature to `process-compose` so we don't need the above comment:
https://github.com/F1bonacc1/process-compose/discussions/35

This way not only the honcho implementation could fulfill the new interface.

Nebucatnetzer:
Another use case is bootstrapping the project.
E.g. I start Postgres and Django in the background and then I run some commands to run migrations and insert fixtures.
Afterwards I open the URL for the dev server in the browser.
The dev then can start developing and if he needs to reattach to the log view.
I'm currently doing this with overmind.

domenkozar:
This is now implemented in #745

PR #89, opened 2022-11-19
Rename deadnix.fix to deadnix.edit

jocelynthode:
Update doc to match changes in PR https://github.com/cachix/pre-commit-hooks.nix/pull/189


domenkozar:
That file is auto generated, so we need to run ``devenv update`` instead

jocelynthode:
Ah hadn't seen that. Sorry for the noise

PR #91, opened 2022-11-20
Javascript: use different node package 

samjwillis97:
This adds the ability to use a different node package with the javascript language enabled. I have used the same implementation as is used for Ruby.

I am not sure if this is the direction you are looking to go in to add support for different versions of packages but thought it was worth raising a PR and starting a conversation.


domenkozar:
This is great, thanks :)

PR #95, opened 2022-11-20
language: robotframework

datakurre:
Well... robotframework is based on Python, so it come with Python interpreter and could conflict with ``languages.python.enable = true`` if it has different interpreter.

For a second, I thought if robotframework should be a Python application instead of package, but it is usual for custom Robot Framework Python libraries to import some API modules from robot modules.

Of course, this gets useful in the real world only with additional packages, and Nixpkgs seems to already have a few (including geckodriver and chromedriver to be used with seleniumlibrary):

https://search.nixos.org/packages?channel=22.05type=packages&query=robotframework

### DIFF
--- a/src/modules/languages/robotframework.nix
+++ b/src/modules/languages/robotframework.nix
@@ -18,7 +18,7 @@ in
 
   config = lib.mkIf cfg.enable {
     packages = with pkgs; [
-      (cfg.python.withPackages(ps: [
+      (cfg.python.withPackages (ps: [
         ps.robotframework
       ]))
     ];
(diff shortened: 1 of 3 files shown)


src/modules/languages/robotframework.nix
```diff
@@ -0,0 +1,30 @@
...
+      default = pkgs.python3;
+      defaultText = "pkgs.python3";
+      description = "The Python package to use.";
+    };
+
+  };
+
+  config = lib.mkIf cfg.enable {
+    packages = with pkgs; [
+      (cfg.python.withPackages(ps: [
```
domenkozar, line 21:
```suggestion
      (cfg.python.withPackages (ps: [
```

PR #112, opened 2022-11-22
Pass `input` as a whole argument to avoid confusion

sayanarijit:
Also, update docs and some minor formatting fixes.

### DIFF
--- a/devenv.nix
+++ b/devenv.nix
@@ -1,8 +1,8 @@
-{ pkgs, nix, lib, config, ... }:
+{ inputs, pkgs, lib, config, ... }:
 
 {
   packages = [
-    (import ./src/devenv.nix { inherit pkgs nix; })
+    (import ./src/devenv.nix { inherit pkgs; nix = inputs.nix; })
     pkgs.python3Packages.virtualenv
     pkgs.python3Packages.cairocffi
     pkgs.yaml2json
--- a/devenv.nix
+++ b/devenv.nix
@@ -1,5 +1,11 @@
 { inputs, pkgs, lib, config, ... }:
 
+let
+  pre-commit-check = inputs.pre-commit-hooks.run {
+    src = ./.;
+    hooks.shellcheck.enable = true;
+  };
+in
 {
   packages = [
     (import ./src/devenv.nix { inherit pkgs; nix = inputs.nix; })
@@ -13,6 +19,8 @@
   processes.build.exec = "${pkgs.watchexec}/bin/watchexec -e nix nix build";
 
   enterShell = ''
+    ${pre-commit-check.shellHook}
+
     echo "To Install:"
     echo
     echo "virtualenv ."
@@ -61,7 +69,7 @@
 
       # Enable all languages tooling!
       ${lib.concatStringsSep "\n  " (map (lang: "languages.${lang}.enable = true;") (builtins.attrNames config.languages))}
-    
+
       # If you're missing a language, please contribute it by following examples of other languages <3
     }
     EOF
--- a/devenv.nix
+++ b/devenv.nix
@@ -1,11 +1,5 @@
 { inputs, pkgs, lib, config, ... }:
 
-let
-  pre-commit-check = inputs.pre-commit-hooks.run {
-    src = ./.;
-    hooks.shellcheck.enable = true;
-  };
-in
 {
   packages = [
     (import ./src/devenv.nix { inherit pkgs; nix = inputs.nix; })
@@ -19,8 +13,6 @@ in
   processes.build.exec = "${pkgs.watchexec}/bin/watchexec -e nix nix build";
 
   enterShell = ''
-    ${pre-commit-check.shellHook}
-
     echo "To Install:"
     echo
     echo "virtualenv ."
--- a/docs/inputs.md
+++ b/docs/inputs.md
@@ -13,11 +13,19 @@ inputs:
     url: github:cachix/pre-commit-hooks.nix
 ```
 
-Input names like ``nixpkgs`` and ``pre-commit-hooks`` are identifiers for what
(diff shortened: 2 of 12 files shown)


docs/inputs.md
```diff
@@ -13,11 +13,19 @@ inputs:
...
 
-Input names like ``nixpkgs`` and ``pre-commit-hooks`` are identifiers for what
- is passsed in the first line of the function:
+The dependencies you mention as `inputs` are passsed as an argument to the function.
 
 ```nix title="devenv.nix"
-{ pkgs, lib, nixpkgs, pre-commit-hooks, config, ... }:
+{ inputs, ... }:
+
+{
```
domenkozar, line 21: Let's show an example how to reference for example pre-commit-hooks from the inputs

docs/inputs.md
```diff
@@ -27,6 +35,13 @@ See [basics](basics.md) for more about ``devenv.nix``.
 
 There are a few special inputs passed into ``devenv.nix``:
 
+```nix title="devenv.nix"
+{ pkgs, lib, config, pre-commit-hooks, ... }:
```
domenkozar, line 39: that should be inputs now

docs/inputs.md
```diff
@@ -27,6 +35,13 @@ See [basics](basics.md) for more about ``devenv.nix``.
...
+
+{
+}
+```
+
 - ``pkgs`` is a ``nixpkgs`` input containing all of the available packages for your system.
 - ``lib`` is [a collection of functions for working with Nix data structures](https://nixos.org/manual/nixpkgs/stable/#sec-functions-library).
 - ``config`` is the resolved configuration for your developer environment, which you can use to reference any other options set in ``devenv.nix``.
```
domenkozar, line 47: `inputs.pre-commit-hooks`

docs/pre-commit-hooks.md
```diff
@@ -5,7 +5,7 @@ We recommend a two-step approach for integrating your linters and formatters.
 ## 1) At commit time to make sure that commits are well-formatted
 
 ```nix title="devenv.nix"
-{ pkgs, ... }:
+{ inputs, pkgs, ... }:
```
domenkozar, line 8: Are inputs needed here?

docs/inputs.md
sayanarijit: Getting

```
error: The option `inputs' does not exist. Definition values:
       - In `/home/sayanarijit/Documents/GitHub/devenv/devenv.nix':
           {
             pre-commit = {
               hooks = {
                 nixpkgs-fmt = {
                   enable = true;
           ...

       … while evaluating the attribute 'config'

       at https://github.com/NixOS/nixpkgs/blob/69335c46c48a73f291d5c6f332fb9fe8b8e22b30/lib/modules.nix:330:9:

          329|         options = checked options;
          330|         config = checked (removeAttrs config [ "_module" ]);
             |         ^
          331|         _module = checked (config._module);
```

sayanarijit: You mean something like https://github.com/cachix/devenv/pull/112/files#diff-cef540503c67fa4a36d1b9fe95a62f979c4b8b2bc43a6e023e250b7154341afcR3-R23?

It's not working btw.

flake.nix
```diff
@@ -14,7 +14,7 @@
   };
   inputs.nix.url = "github:domenkozar/nix/relaxed-flakes";
 
-  outputs = { self, nixpkgs, pre-commit-hooks, nix, ... }:
+  outputs = { self, nixpkgs, pre-commit-hooks, nix, ... }@inputs:
```
domenkozar, line 17:
```suggestion
  outputs = { self, nixpkgs, pre-commit-hooks, nix, ... }:
```
This is not needed :)

devenv.nix
```diff
@@ -1,8 +1,14 @@
-{ pkgs, nix, lib, config, ... }:
+{ inputs, pkgs, lib, config, ... }:
 
+let
```
domenkozar, line 3: Why do you think we need shellcheck as part of the repo?

docs/inputs.md
domenkozar: That's probably not the best example, maybe we can use an example of using git from a stable version of Nix:

```yaml
inputs:
  nixpkgs-stable:
    url: github:NixOS/nixpkgs/nixpkgs-22.05
```

And then:

```nix
{ inputs, system, ... }:

let 
  nixpkgs-stable = import inputs.nixpkgs-stable { inherit system; };
in {
  packages = [ pkgs-stable.git ];
}
```

domenkozar (changes requested):

docs/inputs.md
sayanarijit: `github:NixOS/nixpkgs/nixpkgs-22.05` doesn't resolve. I think you meant `github:NixOS/nixpkgs/nixos-22.05`. Also, `system` isn't a standalone argument. `pkgs.system` works.

This brings another scope for improvement:

How about we could declare it like:

```yaml
inputs:
  nixpkgs-unstable:  # Defaults to `url: github:NixOS/nixpkgs/nixpkgs-unstable`
  nixos-22.05:  # Defaults to `url: github:NixOS/nixpkgs/nixos-22.05`
```

devenv.nix
sayanarijit: Thought it was serving as an example. Removing.

docs/inputs.md
domenkozar: Once we introduce registry this is quite possible

sayanarijit: I meant how about letting `url: github:NixOS/nixpkgs` be the default registry? Ultil we have a separate registry field, we have to be specific when adding non github:NixOS/nixpkgs deps.

PR #115, opened 2022-11-22
feat(processes): add process-compose as new process.implementation option

### LINKED ISSUE #78: Consider cross-platform userspace process orchestrator for process management
To avoid reimplementing common service requirements (wait until postgres is ready to launch dependent services) I recommend taking a look at generating the data structure of https://github.com/F1bonacc1/process-compose and use it provide an alternative `devenv up` command.

Nix is perfect for supplying the binaries, the startup/shutdown scripts can be added as devenv scripts and the execution/scheduling can be handed off to this new opt-in(?) `process-compose` binary.

I've been developing with it via the classic `Nix Flake + direnv + process-compose` and it is the perfect `docker-compose` replacement.
I would be open to contributing to this, if the idea seems acceptable to the maintainers.

thenonameguy:
fixes #78

### DIFF
--- a/examples/process-compose/devenv.nix
+++ b/examples/process-compose/devenv.nix
@@ -1,0 +1,30 @@
+{ pkgs, ... }:
+
+{
+  process.implementation = "process-compose";
+
+  processes.foo.exec = "echo foo; sleep 5";
+
+  # implicitly creates a postgres process
+  # consider https://github.com/cachix/devenv/issues/75
+  postgres.enable = true;
+
+  processes.bar = {
+    exec = (pkgs.writeShellScript "complex-process" ''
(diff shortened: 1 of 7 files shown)


examples/process-compose/devenv.nix
```diff
@@ -0,0 +1,30 @@
+{ pkgs, ... }:
+
+{
+  process.implementation = "process-compose";
+
+  processes.foo.exec = "echo foo; sleep 5";
+
+  # implicitly creates a postgres process
+  # consider https://github.com/cachix/devenv/issues/75
```
domenkozar, line 9: I left a comment on that issue, let's leave this discussion there and not pollute the example :)

PR #120, opened 2022-11-24
MySQL improvements

bobvanderlinden:
Currently whenever mysql is called with arguments that contain spaces, it will result in errors.

Example:

```console
$ mysql -e "select @@socket"
... Unknown database '@@socket'
```

This is fixed by using `"$@"` instead of `$@`.

In addition, the environment variables that were used do not seem to be mentioned in the documentation of mysql. The env vars that are mentioned are `MYSQL_HOME` and `MYSQL_UNIX_SOCKET`. These are automatically picked up by the cli tools.
This also makes it unnecessary to use `--socket=$MYSQLSOCKET`, so I have removed these.

Lastly, `MYSQL_TCP_PORT` is another documented environment variable. Since the applications I have worked on use TCP instead of unix sockets I thought it was a good idea to set this variable based on the mysqld port from the settings.


domenkozar:
@shyim

src/modules/mysql.nix
```diff
@@ -120,19 +120,20 @@ in
...
 
-    env.MYSQLDATA = config.env.DEVENV_STATE + "/mysql";
-    env.MYSQLSOCKET = config.env.DEVENV_STATE + "/mysql.sock";
+    env.MYSQL_HOME = config.env.DEVENV_STATE + "/mysql";
+    env.MYSQL_UNIX_PORT = config.env.DEVENV_STATE + "/mysql.sock";
+    env.MYSQL_TCP_PORT = mkIf (cfg.settings.mysqld ? port) (toString cfg.settings.mysqld.port);
 
     scripts.mysql.exec = ''
-      exec ${cfg.package}/bin/mysql ${mysqlOptions} --socket=$MYSQLSOCKET $@
+      exec ${cfg.package}/bin/mysql ${mysqlOptions} "$@"
```
shyim, line 128: If we don't need to set `--socket` we can remove all scripts related to mysql

shyim (changes requested):
Looks good. Small change

src/modules/mysql.nix
bobvanderlinden: `my.cnf` is not in `MYSQL_HOME` yet. If we copy or link it to `MYSQL_HOME` then the wrappers can be removed.

Maybe copy/linking it right before starting mysqld could be an option?

src/modules/mysql.nix
```diff
@@ -120,19 +120,20 @@ in
       cfg.package
     ];
 
-    env.MYSQLDATA = config.env.DEVENV_STATE + "/mysql";
-    env.MYSQLSOCKET = config.env.DEVENV_STATE + "/mysql.sock";
+    env.MYSQL_HOME = config.env.DEVENV_STATE + "/mysql";
+    env.MYSQL_UNIX_PORT = config.env.DEVENV_STATE + "/mysql.sock";
+    env.MYSQL_TCP_PORT = mkIf (cfg.settings.mysqld ? port) (toString cfg.settings.mysqld.port);
```
shyim, line 125: image

with just
```
mysql.enable = true;
```

src/modules/mysql.nix
shyim: Yea we could link it while starting the MySQL server. 👍

domenkozar:
Let me know when it's ready to be merged :)

src/modules/mysql.nix
bobvanderlinden: I'll look into doing that in a separate PR.

src/modules/mysql.nix
bobvanderlinden: Ah, oops. I have fixed this issue in the latest version :+1:

src/modules/mysql.nix
```diff
@@ -120,20 +121,16 @@ in
       cfg.package
     ];
 
-    env.MYSQLDATA = config.env.DEVENV_STATE + "/mysql";
-    env.MYSQLSOCKET = config.env.DEVENV_STATE + "/mysql.sock";
+    mysql.settings.mysql.user = mkDefault "root";
+    mysql.settings.client.user = mkDefault config.mysql.settings.mysql.user;
+    mysql.settings.mysqldump.user = mkDefault config.mysql.settings.mysql.user;
```
bobvanderlinden, line 126: This sets the user that these tools will use by default.

The tools internally still use `-u root`, as this has the correct rights to setup databases and do ping.

bobvanderlinden:
@shyim Could you have another look?

src/modules/mysql.nix
```diff
@@ -7,28 +7,29 @@ let
...
   startScript = pkgs.writeShellScriptBin "start-mysql" ''
     set -euo pipefail
 
-    if [[ ! -d "$MYSQLDATA" ]]; then
-      mkdir -p "$MYSQLDATA"
-      ${if isMariaDB then "${cfg.package}/bin/mysql_install_db" else "${cfg.package}/bin/mysqld"} ${mysqldOptions} ${optionalString (!isMariaDB) "--initialize-insecure"}
+    if [[ ! -d "$MYSQL_HOME" ]]; then
+      mkdir -p "$MYSQL_HOME"
+      ln -sf "${configFile}" "$MYSQL_HOME"/my.cnf
+      ${if isMariaDB then "${cfg.package}/bin/mysql_install_db" else "${cfg.package}/bin/mysqld"} ${optionalString (!isMariaDB) "--initialize-insecure"}
```
shyim, line 16: image

datadir is not passed anymore to `mysql_install_db`

shyim (changes requested):

src/modules/mysql.nix
bobvanderlinden: Hmm, I probably didn't properly clear all of the mysql directory. I've removed the commit that removes the wrapper scripts. I'll keep that for a potentially later PR.

shyim (approved):

shyim:
Thanks! :)

bobvanderlinden:
> Thanks! :)

Thank you for testing 👌

> Let me know when it's ready to be merged :)

It is now 👍

PR #142, opened 2022-11-29
WIP: Creating files

domenkozar:
The idea is that you can create files like:

```nix
{
  files."devcontainer.json".json = { name = "myenv"; };
  files.".editorconfig".ini = { "*" = { end_of_line = "lf"; }; };
  files."README.md".text = "Here we go!";
}
```

And running `devenv shell` would print:

```
$ devenv shell
...
Creating devcontainer.json
Creating README.md
Creating .editorconfig
```

## TODO

- [ ] write documentation introduction
- [ ] implement different format generation
- [ ] assert that only one format can be used per file

### DIFF
--- a/src/modules/files.nix
+++ b/src/modules/files.nix
@@ -1,0 +1,58 @@
+{ pkgs, lib, ... }:
+
+let
+  types = lib.types;
+  ini = pkgs.formats.ini { };
+  json = pkgs.formats.json { };
+  yaml = pkgs.formats.yaml { };
+  toml = pkgs.formats.toml { };
+  fileType = types.submodule ({ config, ... }: {
+    ini = lib.mkOption {
+      type = ini.type;
+      default = null;
+      description = "TODO";
+    };
+
+    json = lib.mkOption {
+      type = json.type;
+      default = null;
+      description = "";
+    };
+    yaml = lib.mkOption {
+      type = yaml.type;
+      default = null;
+      description = "";
+    };
+    toml = lib.mkOption {
+      type = toml.type;
+      default = null;
+      description = "";
+    };
+    text = lib.mkOption {
+      type = types.str;
+      default = null;
+      description = "";
+    };
+    executable = lib.mkOption {
+      type = types.bool;
+      default = false;
+      description = "";
+    };
+  });
+  createFile = filename: fileOption: ''
+    echo "Creating ${filename}"
+    mkdir -p $(dirname ${filename})
+    ${lib.optionalString fileOption.executable "chmod +x ${filename}"}
+  '';
+in
+{
+  options.files = lib.mkOption {
+    type = types.attrsOf fileType;
+    default = { };
+    description = "";
+  };
+
+  config = {
+    enterShell = lib.mapAttrsToList createFile config.files;
+  };
+}
--- a/src/modules/info.nix
+++ b/src/modules/info.nix
@@ -23,5 +23,8 @@
 
     # processes
     ${lib.concatStringsSep "\n" (lib.mapAttrsToList (name: process: "- ${name}: ${process.exec}") config.processes)}
+
+    # files
+    ${lib.concatStringsSep "\n" (lib.mapAttrsToList (filename: file: filename) config.files)}
   '';
 }
--- a/src/modules/top-level.nix
+++ b/src/modules/top-level.nix
@@ -39,6 +39,7 @@ in
     ./mongodb.nix
     ./pre-commit.nix
     ./info.nix
+    ./files.nix
     ./scripts.nix
     ./processes.nix
     ./update-check.nix


domenkozar:
> Questions:
> 
> > implement different format generation
> 
> Would better make sure our design is extensible since the number of output formats [isn't defined](https://github.com/cruel-intentions/stylops#projects)?

That's a good point, how can we make the design extensible without sacrificing the simplicity of the user api?

> 
> > assert that only one format can be used per file
> 
> If we resolve all other types to `files."fileName.ext".text` it would automatically reject config if two files point to same the path?

What would reject the config of two files in such case?

> On other side, should we change type of `text` to be lines (that auto merge)?

For some files you would want to append, for some you would want to override. Not sure.

hugosenari:
> That's a good point, how can we make the design extensible without sacrificing the simplicity of the user api?

:shrug: 

> What would reject the config of two files in such case?
Ie:
```nix
{
  files."devcontainer.json".json = { name = "myenv"; };
  files."devcontainer.json".ini  = { "*" = { end_of_line = "lf"; }; };  
}
```
If both resolve to `files."devcontainer.json".text` (in [my code](https://github.com/cruel-intentions/devshell-files/blob/master/modules/file-format.nix#L27) is `file."devcontainer.json".text` because of infinity recursion), eval modules will complain that value was settled twice, unless we use without `mkForce` (`mkDefault`, etc..) or type is `lines` (in this case they will be appended to the end)

hugosenari:
> That's a good point, how can we make the design extensible without sacrificing the simplicity of the user api?

@bew once told me that we could add only text option and user does whatever to convert into the text.

I don't agree, but is good seen things through a different lens

hugosenari:
> I'm interested whether this should be inside enterShell or whether some kind of setup script should be introduced. Mostly because these files are 'conflict-prone'. What should happen upon conflicts? Should it error? Should it warn? Should it silently overwrite?

I'm favoring of overwriting, as we need to set a single source of truth. Ideally, files should add header banner 'This files was created by X Y Z do not edit it', but isn't always viable (JSON, no comments),

Other problem would be updates, we can't diff existing files with previous generation to check if changes are updates or someone edited it directly. Any change would'be a conflict.

But we should not add any option that creates files by default, only when enabled. ie: Devenv shouldn't control gitignore file, unless someone configure it to do. #178 

> An approach I took on my devenv config is to have a `setup` script that you usually run once. It includes some authentication assertions (`.npmrc`), copy some (project specific) config.example.json to config.json file and run npm ci. Doing this each enterShell is too much, which is why I placed it in a script. If anything fails, you attempt to fix it and rerun the script again. Not ideal, but it served my purpose for now. Might be something to consider here as well?

The only problem of this is to have files out of sync more often, ie:
- edit a nix file stage.nix (cfg for stage environment)
- activate the devenv
- ~run setup~
- create a zip file for stage deployment.

In steps above, user forgot to run setup, deploying stage with old conf file.
He can forget to activate either, but we minimized the problem.
With direnv the problem almost disappear.

domenkozar:
> Nice! This can really come in handy!
> 
> I'm interested whether this should be inside enterShell or whether some kind of setup script should be introduced. Mostly because these files are 'conflict-prone'. What should happen upon conflicts? Should it error? Should it warn? Should it silently overwrite?
> 
> If erroring/warning, should it not load the shell? How can you make sure it reattempts to apply the files?
> 
> An approach I took on my devenv config is to have a `setup` script that you usually run once. It includes some authentication assertions (`.npmrc`), copy some (project specific) config.example.json to config.json file and run npm ci. Doing this each enterShell is too much, which is why I placed it in a script. If anything fails, you attempt to fix it and rerun the script again. Not ideal, but it served my purpose for now. Might be something to consider here as well?

My thinking is that conflicts should be resolved at Nix level, if you define two of the same files. Otherwise Nix assumes it's the source of truth what should be the latest version of the declared file: it will override.

I'm thinking to split `enterShell` into `motd` (which would be more about printing into the shell` and `setup` that would take care of setting up the environment.

I'm not sure creating files will slow things down that much given that everyone has an SSD these days. But for slower operations we might consider some command to run things manually.

src/modules/files.nix
```diff
@@ -0,0 +1,58 @@
+{ pkgs, lib, ... }:
+
+let
+  types = lib.types;
+  ini = pkgs.formats.ini { };
```
blaggacao, line 5: For editorconfig, you'll need https://github.com/nix-community/nixpkgs.lib/blob/master/lib/generators.nix#L164

adrian-gierakowski:
> > I'm not sure creating files will slow things down that much given that everyone has an SSD these days. But for slower operations we might consider some command to run things manually.
> 
> I'd say there should be an option to disable auto-creation and provide a command\script which could be used to recreate the files manually. A script which checks that all files are in expected state would also be useful for running checks in CI.

An idea of how disabling of auto-creation could be implemented: 
https://github.com/cachix/devenv/issues/209#issuecomment-1435671727

For manual execution we could add `create-all` option of package type (like [this one](https://github.com/rhinofi/devshell-files/commit/017d6c90029a23074887305d5bcc23630896d7cf#diff-4f03e2c32c9953b21249d9c1b672e24d29d35b99b8c7c9f83a62c44ed1e10851R37-R45)), and then have it added to the devenv `packages`

adrian-gierakowski:
@hugosenari 

> On other side, should we change type of `text` to be lines (that auto merge)?

Personally I think the default should be not to merge as auto-merging could lead to surprising bugs. But we could let users choose with options like (pseudocode):

```nix
{
  options.merge.enable = { type = bool; default = false; };
  options.merge.strategy = { type = optionType; default = lines; };
}
```

And then set `text.type = config.merge.strategy`. Apart form `lines`, one might want to use [commas or envVars](https://github.com/NixOS/nixpkgs/blob/7fac81ab46cb6bb4d859217893ffa61f011c1b72/lib/types.nix#L434-L435), or some custom strategy.

adrian-gierakowski:
@domenkozar as mentioned on the bottom of [this post](https://github.com/cruel-intentions/devshell-files/issues/8#issuecomment-1435785862) I believe it would be beneficial to have this implemented in a "platform" agnostic way, so that anyone can re-use this regardless if they use `devenv`, `devshell`, or whatever else. The standalone module should contain option definitions and logic for creating self contained scripts which could be executed from anywhere to create files from given root dir.  The only thing to be done on `devenv` side would be to import the module and add [devenv specific config](https://github.com/cachix/devenv/pull/142/files#diff-f9827f6d19c5d322f8a81feef284182f1ec0d3745f29133af1326cac10d75a6bR55-R57). 

I'm happy to work on this if we can agree on an interface.

domenkozar:
> text.type = config.merge.strategy

I don't think it's possible for options to depend on the config - that creates an infinity error.

> I believe it would be beneficial to have this implemented in a "platform" agnostic way

I agree, if we can unite the design goals :)

adrian-gierakowski:
> > text.type = config.merge.strategy
> 
> I don't think it's possible for options to depend on the config - that creates an infinity error.

I used to think that but it seems to work [here](https://github.com/adrian-gierakowski/nix-microservice-monorepo/blob/2bd8e0c8de066dd911e428f4bf4bd41368188e23/nix/modules/service.nix#L55), and [here](https://github.com/adrian-gierakowski/nix-microservice-monorepo/blob/2bd8e0c8de066dd911e428f4bf4bd41368188e23/services/random-service/default.nix#L10).

adrian-gierakowski:
> > I believe it would be beneficial to have this implemented in a "platform" agnostic way
> 
> I agree, if we can unite the design goals :)

Sure!

@hugosenari do you have any further comments apart from the feedback you’ve provided here so far?

domenkozar:
if someone wants to pick this up that would be great!

bobvanderlinden:
Just today I ran into a use-case for this. I'll have a look cherry-picking this PR and brining it up-to-date :+1:

bobvanderlinden:
I've created https://github.com/cachix/devenv/pull/1590 against the latest main. Mostly in line with this PR. Feedback is welcome.

PR #144, opened 2022-11-29
Add a comment into gitignore

### LINKED ISSUE #129: devenv init: auto adjust gitignore
The `.devenv.flake.nix` and the `.devenv` folder should not be commited. When the user is running init we could directly adjust the gitignore and add those. I guess it's better DX

shyim:
Fixes #129


thenonameguy:
LGTM

PR #163, opened 2022-12-01
improve postgres

shyim:
Config was updated only once

Added two new options

```nix
postgres.listen_addresses = "127.0.0.1";
postgres.port = 5000;
```

By default, postgres still listens socket only with default port. With that options you can enable tcp again.

Also configured the default env vars for postgres to `psql` by default works https://www.postgresql.org/docs/current/libpq-envars.html

### DIFF
--- a/src/modules/postgres.nix
+++ b/src/modules/postgres.nix
@@ -9,13 +9,17 @@ let
   setupScript = pkgs.writeShellScriptBin "setup-postgres" ''
     set -euo pipefail
     export PATH=${cfg.package}/bin:${pkgs.coreutils}/bin
+
+    # Setup config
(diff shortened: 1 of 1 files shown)


src/modules/postgres.nix
```diff
@@ -62,8 +80,10 @@ in
     ];
 
     env.PGDATA = config.env.DEVENV_STATE + "/postgres";
+    env.PGHOST = config.env.PGDATA;
+    env.PGPORT = cfg.port;
 
-    scripts."psql-devenv".exec = "${cfg.package}/bin/psql -h $PGDATA $@";
+    scripts."psql-devenv".exec = "${cfg.package}/bin/psql $@";
```
domenkozar, line 86:
```suggestion
```

PR #175, opened 2022-12-02
feat: add bin folder to .devenv

### LINKED ISSUE #167: Generate a bin folder inside .devenv
All Jetbrains IDEs needs an absolute path where the executables are living.

It would be nice if we would symlink php to `.devenv/bin/php`. So you can configure your IDE file path to that directory.
Of course when the package gets updated we update it there too.

A little bit like the `/run/current-system/sw`

shyim:
- Reduces the $PATH by only adding one packages with a bin folder
- Symlinks the bin to `.devenv/bin` so I can pass this as interpreter path to PhpStorm

fixes #167

### DIFF
--- a/src/modules/top-level.nix
+++ b/src/modules/top-level.nix
@@ -48,7 +48,7 @@ in
     ./update-check.nix
   ] ++ map (name: ./. + "/languages/${name}") (builtins.attrNames (builtins.readDir ./languages));
 
-  config = let packageEnv = pkgs.buildEnv { name = "devenv"; paths = config.packages; }; in {
+  config = let packageEnv = pkgs.buildEnv { name = "devenv"; paths = config.packages; ignoreCollisions = true; }; in {
 
     # TODO: figure out how to get relative path without impure mode
     env.DEVENV_ROOT = builtins.getEnv "PWD";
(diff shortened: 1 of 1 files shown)


src/modules/top-level.nix
```diff
@@ -48,7 +48,7 @@ in
     ./update-check.nix
   ] ++ map (name: ./. + "/languages/${name}") (builtins.attrNames (builtins.readDir ./languages));
 
-  config = {
+  config = let packageEnv = pkgs.buildEnv { name = "devenv"; paths = config.packages; }; in {
```
domenkozar, line 51: We need to pass here `ignoreCollisions = true`

thenonameguy:
My reverted mkNakedShell PR fixes this as well. Let me do that instead over the weekend.

shyim:
@thenonameguy will you create also a `bin` folder? Ot should I just revert the shell changes :)

thenonameguy:
@shyim It will create the `/bin` folder as well as well as adding it to the `PATH`.
I have made the modifications, just writing tests for testing the native extension building (correct pkg-config/ld env vars).

shyim:
I will make my own module

PR #190, opened 2022-12-05
mysql: add mysql password option

shyim:
I need for the https://www.adminer.org a password to access the database.

### DIFF
--- a/src/modules/mysql.nix
+++ b/src/modules/mysql.nix
@@ -47,6 +47,18 @@ let
       fi
     '') cfg.initialDatabases}
 
+    ${optionalString (cfg.rootPassword != null) ''
+      # Set root password
+
+      if [[ ! -f "$MYSQL_HOME/.root_password_set" ]]; then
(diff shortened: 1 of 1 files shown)


src/modules/mysql.nix
```diff
@@ -47,6 +47,18 @@ let
       fi
     '') cfg.initialDatabases}
 
+    ${optionalString (cfg.rootPassword != null) ''
+      # Set root password
+
+      if [[ ! -f "$MYSQL_HOME/.root_password_set" ]]; then
+        ${cfg.package}/bin/mysql ${mysqlOptions} -u root -N -e "ALTER USER 'root'@'localhost' IDENTIFIED BY '${cfg.rootPassword}';"
+
+        touch "$MYSQL_HOME/.root_password_set"
```
domenkozar, line 56: This should include the password hash so if it changes, the password is changes

shyim: You need the old password to set the new one

shyim: Maybe we allow adding additional users using options and let the `root` be root and only accessible using socket?

shyim:
I will rework this to allow adding users

Issue #133, opened 2022-11-27
Support for nix-direnv (or copy its mechanism)


andrewhamon:
Hey there, loving the project so far!

Right now my only pain point (as compared to my previous setup using vanilla nix shell + [nix-direnv](https://github.com/nix-community/nix-direnv)) is the speed of opening new shells. Even though all the build artifacts are cached, it still takes several seconds to evaluate all the nix expressions and whatnot for a simple devenv that pulls in node and yarn.

I'm pretty sure nix-direnv solves this by caching the values of all environment variables, bypassing any nix evaluation at all. It is much faster to enter a cached shell with nix-direnv - nearly instant.

domenkozar:
Agreed, I'd really like that to be a feature of Nix one day.

Refs #35

domenkozar:
See https://github.com/cachix/devenv/pull/387

PR #191, opened 2022-12-05
shell: use naked shell concept with env var prepends for leaner shell with autocompletion

### LINKED ISSUE #79: Direnv is extremely verbose
Disclaimer: this boils down to a direnv and `print-dev-env` issue, but it still relevant to the UX of devenv.

Running `direnv allow` after running `devenv init`, on top of the normal output of devenv, we get something like:

> direnv: export +AR +AS +CC +CONFIG_SHELL +CXX +DETERMINISTIC_BUILD +DEVENV_DOTFILE +DEVENV_ROOT +DEVENV_STATE +GREET +HOST_PATH +IN_NIX_SHELL +LD +NIX_BINTOOLS +NIX_BINTOOLS_WRAPPER_TARGET_HOST_x86_64_unknown_linux_gnu +NIX_BUILD_CORES +NIX_BUILD_TOP +NIX_CC +NIX_CC_WRAPPER_TARGET_HOST_x86_64_unknown_linux_gnu +NIX_CFLAGS_COMPILE +NIX_ENFORCE_NO_NATIVE +NIX_HARDENING_ENABLE +NIX_INDENT_MAKE +NIX_LDFLAGS +NIX_STORE +NM +OBJCOPY +OBJDUMP +PYTHONHASHSEED +PYTHONNOUSERSITE +PYTHONPATH +RANLIB +READELF +SIZE +SOURCE_DATE_EPOCH +STRINGS +STRIP +TEMP +TEMPDIR +TMP +TMPDIR +_PYTHON_HOST_PLATFORM +_PYTHON_SYSCONFIGDATA_NAME +buildInputs +buildPhase +builder +cmakeFlags +configureFlags +depsBuildBuild +depsBuildBuildPropagated +depsBuildTarget +depsBuildTargetPropagated +depsHostHost +depsHostHostPropagated +depsTargetTarget +depsTargetTargetPropagated +doCheck +doInstallCheck +dontAddDisableDepTrack +mesonFlags +name +nativeBuildInputs +out +outputs +patches +phases +propagatedBuildInputs +propagatedNativeBuildInputs +shell +shellHook +stdenv +strictDeps +system ~PATH ~XDG_DATA_DIRS

This is very noisy, not very useful, and gives a bad impression.

---

In the end, this is because `nix-shell` (and `nix develop` and `nix print-dev-env`) is designed to "provide the build environment of a derivation", not making shell environments that don't build a derivation.

`devshell` fixes that by reimplementing `mkShell` in a way that removes everything that we don't want (doesn't use `mkDerivation`, and does additional cleanup): https://github.com/numtide/devshell/blob/master/nix/mkNakedShell.nix.

thenonameguy:
fixes #79
closes #175


src/modules/mkNakedShell.nix
```diff
@@ -0,0 +1,136 @@
...
+
+      # For `nix develop`. We get /noshell on Linux and /sbin/nologin on macOS.
+      if [[ "$SHELL" == "/noshell" || "$SHELL" == "/sbin/nologin" ]]; then
+        export SHELL=${bashPath}
+      fi
+
+      # https://github.com/numtide/devshell/issues/158
+      PATH=''${PATH#/path-not-set:}
+
+      export DEVENV_PROFILE=${profile}
```
thenonameguy, line 92: This should probably be documented, question is where.
This allows easy access to the built environment for end users to refer to in their scripts.

src/modules/mkNakedShell.nix
```diff
@@ -0,0 +1,136 @@
...
+
+      # prepend common compilation lookup paths
+      export PKG_CONFIG_PATH="$DEVENV_PROFILE/lib/pkgconfig:$PKG_CONFIG_PATH"
+      export LD_LIBRARY_PATH="$DEVENV_PROFILE/lib:$LD_LIBRARY_PATH"
+      export LIBRARY_PATH="$DEVENV_PROFILE/lib:$LIBRARY_PATH"
+      export C_INCLUDE_PATH="$DEVENV_PROFILE/include:$C_INCLUDE_PATH"
+
+      # these provide shell completions / default config options
+      export XDG_DATA_DIRS="$DEVENV_PROFILE/share:$XDG_DATA_DIRS"
+      export XDG_CONFIG_DIRS="$DEVENV_PROFILE/etc/xdg:$XDG_CONFIG_DIRS"
```
thenonameguy, line 105: Chucking these in here felt kind of crude.
The argument is that this way the `startupEnv`/`shellHook` can interact with these values.

examples/process-compose/devenv.nix
```diff
@@ -7,16 +7,28 @@
 
   postgres.enable = true;
 
+  languages.ruby.enable = true;
+
+  packages = [ pkgs.imagemagick6.dev ];
```
thenonameguy, line 12: This is a bit sub-optimal from a usage standpoint.
Multiple output packages are in general a pain to discover.
Maybe extend `devenv search` to include package outputs?

thenonameguy:
@shyim can you verify if this solves https://github.com/cachix/devenv/issues/167 for you?
`$DEVENV_PROFILE/bin` should be the equivalent of your symlink in your PR.

shyim:
devenv flake lock seems out of sync? 
image

shyim:
@thenonameguy That does not solve my problem. I need a fixed PATH that never changes. So I can say in the IDE this is my Binary and does not have to touch it again.

image

The given path is nix store and changes 😢

thenonameguy:
> @thenonameguy That does not solve my problem. I need a fixed PATH that never changes. So I can say in the IDE this is my Binary and does not have to touch it again.
> 
> image
> 
> The given path is nix store and changes cry

I see. :disappointed: 
Let's see what @domenkozar says. I think creating this static symlink might be too specific to your environment (or JetBrains product users in general). 
The question is whether to put this logic:
- direnv, generally
- direnv, as a jetbrains module 
- in your project devenv.nix as  custom `enterShell`

I would favor option 2 atm, as it allows for greater extensibility for those products in the future.

shyim:
I have no problem to do that in my own project, but you have this problem any GUI software as the PATH is only altered in the shell and not in the OS

thenonameguy:
> 

For these kind of use-cases we usually use: 
- https://plugins.jetbrains.com/plugin/19275-better-direnv
- https://melpa.org/#/direnv

Or any other `direnv` editor plugin. This auto-detects changes just like your shell does.

domenkozar:
Thanks for all this hard work, this looks like a good step forward!

How would we address for example PYTHONPATH using this change? Or any language setup-hook that would work currently?
That's something I'm worried quite a bit since we'd have to do a lot of work to replicate all those setup-hooks.

thenonameguy:
Thanks!

I would add those to language-specific modules.
I already wanted to add these `PKG_CONFIG`/`LD_LIBRARY_PATH` patches to the C language one. Since if you need these, you will most likely need those tools in your env anyway.
I decided not to do it as they are low-level enough that they are used in many contexts (like the Ruby example).

>That's something I'm worried quite a bit since we'd have to do a lot of work to replicate all those setup-hooks.

I understand your worry, but do not share it. 

`devshell` has quite the market share in this space without any explicit support for these env extensions.
Users get by fine (although we can assume more mature Nix users there).
Patching them on a case by case basis is acceptable IMO, especially if people upstream them in devenv language modules.

I would want to reiterate that `nixpkgs` package definitions do not have a nix-less env extension concept in mind.
All the scripts/env vars are geared towards a succesful `nix-build`. This distinction is important, ie. Ubuntu users are running devenv and are using not the same nix-provided toolchain as in the given nixpkgs rev.

A good example is the `gem install rmagick` line in the `process-compose` example. 
Since the ruby language module does not provide `pkg-config` from nix, the installer pulls down a `pkg-config` version for itself, the version being up to it's discretion. This `pkg-config` does not support the `setup-hook` `PKGCONFIG_PATH_FOR_TARGET` env at all, causing a failure.

The above showcases that the  status quo has it's own faults and needs custom code anyway.

Lorri also does this naked shell approach and seem to achieve some better out-of-the-box experience by relying on a few nix fns:
https://github.com/nix-community/lorri/blob/a20fafa4e79ac19409634d0c69bd2e7b1208706a/release.nix#L305-L323 

This should cover a few more of the top10 languages. I would be interested to do this as a follow-up PR.

thenonameguy:
@domenkozar what would you consider necessary to get this merged?

I was thinking along the lines of making more elaborate examples for the top 10 most loved languages: 
https://survey.stackoverflow.co/2022/#technology-most-loved-dreaded-and-wanted
(Assuming the people who still use Fortran won't use devenv, Rust users are more the target audience)

Process being:
1. Find a non-trivial sample application for the language:
   https://codebase.show/projects/realworld?category=backend is a good candidate
   example: https://github.com/snamiki1212/realworld-v1-rust-actix-web-diesel/
   
2. Fork these repos and replace docker-compose with devenv + process-compose:
   https://github.com/snamiki1212/realworld-v1-rust-actix-web-diesel/blob/main/compose.yaml
   
3. Try to keep project `devenv.nix` as small as possible, preferably diff is in our favor. 
      if devenv modules are expressive enough (no env var overrides needed :) )

This would serve multiple purposes:
1. Make sure the language specific tooling work in a real use-case in a limited scope context
2. The current testing stack just initializes devenv, has no real-life testing equivalent. GHA could be integrated for major releases to check for breakages.
3. A good intution builder for language experts who haven't touched nix yet.

How does that plan sound? We could ask the language module maintainers to do this (I happily volunteer for Clojure).

domenkozar:
Thanks for all this awesome work - I want to do a bit of testing - hoping to get a chance on Monday!

domenkozar:
> @domenkozar what would you consider necessary to get this merged?
> 
> I was thinking along the lines of making more elaborate examples for the top 10 most loved languages: https://survey.stackoverflow.co/2022/#technology-most-loved-dreaded-and-wanted (Assuming the people who still use Fortran won't use devenv, Rust users are more the target audience)
> 
> Process being:
> 
> 1. Find a non-trivial sample application for the language:
>    https://codebase.show/projects/realworld?category=backend is a good candidate
>    example: https://github.com/snamiki1212/realworld-v1-rust-actix-web-diesel/
> 2. Fork these repos and replace docker-compose with devenv + process-compose:
>    https://github.com/snamiki1212/realworld-v1-rust-actix-web-diesel/blob/main/compose.yaml
> 3. Try to keep project `devenv.nix` as small as possible, preferably diff is in our favor.
>    if devenv modules are expressive enough (no env var overrides needed :) )
> 
> This would serve multiple purposes:
> 
> 1. Make sure the language specific tooling work in a real use-case in a limited scope context
> 2. The current testing stack just initializes devenv, has no real-life testing equivalent. GHA could be integrated for major releases to check for breakages.
> 3. A good intution builder for language experts who haven't touched nix yet (modify landing page with language select?).
> 4. A test for this PR, such that within a `devenv shell` the language tooling works.
> 
> How does that plan sound? We could ask the language module maintainers to do this (I happily volunteer for Clojure).
> 
> This was my original goal for: [NixOS/nix.dev#367](https://github.com/NixOS/nix.dev/issues/367) (just using `numtide/devshell` and a lot more Flake code).

This is excellent! Would you do it as part of this PR?

thenonameguy:
> > @domenkozar what would you consider necessary to get this merged?
> > I was thinking along the lines of making more elaborate examples for the top 10 most loved languages: https://survey.stackoverflow.co/2022/#technology-most-loved-dreaded-and-wanted (Assuming the people who still use Fortran won't use devenv, Rust users are more the target audience)
> > Process being:
> > 
> > 1. Find a non-trivial sample application for the language:
> >    https://codebase.show/projects/realworld?category=backend is a good candidate
> >    example: https://github.com/snamiki1212/realworld-v1-rust-actix-web-diesel/
> > 2. Fork these repos and replace docker-compose with devenv + process-compose:
> >    https://github.com/snamiki1212/realworld-v1-rust-actix-web-diesel/blob/main/compose.yaml
> > 3. Try to keep project `devenv.nix` as small as possible, preferably diff is in our favor.
> >    if devenv modules are expressive enough (no env var overrides needed :) )
> > 
> > This would serve multiple purposes:
> > 
> > 1. Make sure the language specific tooling work in a real use-case in a limited scope context
> > 2. The current testing stack just initializes devenv, has no real-life testing equivalent. GHA could be integrated for major releases to check for breakages.
> > 3. A good intution builder for language experts who haven't touched nix yet (modify landing page with language select?).
> > 4. A test for this PR, such that within a `devenv shell` the language tooling works.
> > 
> > How does that plan sound? We could ask the language module maintainers to do this (I happily volunteer for Clojure).
> > This was my original goal for: [NixOS/nix.dev#367](https://github.com/NixOS/nix.dev/issues/367) (just using `numtide/devshell` and a lot more Flake code).
> 
> This is excellent! Would you do it as part of this PR?

Only if you deem it so that it is required for merging. 
Clojure works with this branch, I'm using it for an internal company project already. 
Doing it would not showcase the edge cases that come up from abandoning the `setup-hook`s. (point 4)
That would be more of Python/Ruby/dynamic lang with $LANG's package manager compiling C code problem, as shown above.

I'll clarify that this is a breaking change if it ever gets merged. 
The tradeoff for breakage is that language users will have to extend their specific language module necessary env var settings, which makes them explicit in the devenv context, instead of pulling it sort-of-accidentally for purposes of Nix compilation. 

If you deem it necessary to showcase that we are not breaking devenv users too badly, I can do a realworld project transform to devenv using a $LANG selected by you.

domenkozar:
Could you handle Ruby and Rust? Those two are most popular.

@shyim can you check that PHP works with this PR?

thenonameguy:
> Could you handle Ruby and Rust? Those two are most popular.
> 
> @shyim can you check that PHP works with this PR?

Here is the Rust example, had to extend the postgres module to be on par with the Postgres Docker image:
https://github.com/snamiki1212/realworld-v1-rust-actix-web-diesel/pull/15

docs/reference/options.md
```diff
@@ -3525,7 +3525,7 @@ attribute set of attribute set of (INI atom (null, bool, int, float or string) o
 
 
 ## services.postgres.createDatabase
-Create a database named like current user on startup.
+Create a database named like current user on startup. Only applies when initialDatabases is an empty list.
```
shyim, line 3528: Maybe makes more sense to remove it and the user migrates to the new option :)

thenonameguy:
I made the Ruby example without flakes. This one uses by default a transient sqlite database, so the devenv.nix is really small. Nokogiri was installed succesfully (with the updated dependencies) :tada: 
https://github.com/blrB/hanami-realworld-example-app/pull/1

flake.nix
```diff
@@ -45,23 +45,27 @@
...
+      templates =
+        let
+          simple = {
+            path = ./templates/simple;
+            description = "A direnv supported Nix flake with devenv integration.";
+          };
+        in
+        {
+          inherit simple;
+          default = simple;
```
thenonameguy, line 67: Fixes https://devenv.sh/guides/using-with-flakes/
```
nix flake init --template github:cachix/devenv
```
Not working at all.

flake.nix
```diff
@@ -74,7 +78,10 @@
               };
               modules = [
                 (self.modules + /top-level.nix)
-                { devenv.warnOnNewVersion = false; }
+                {
+                  packages = [ self.packages.${pkgs.system}.devenv ];
```
thenonameguy, line 82: I think it makes sense that in a Flake context the devenv input Flake defines the version used, not `devenv.yaml`'s `devenv` input. This way developers don't need to globally install devenv either, it remains project-scoped.

examples/process-compose/devenv.nix
domenkozar: We should just call `getOutput` on all the outputs for packages and include them all. I think that's a better default for non-experienced Nixers, while you can still pass `pkg.dev` if you want to reduce the closure size.

domenkozar (changes requested):

src/modules/mkNakedShell.nix
domenkozar: https://devenv.sh/files-and-variables/

templates/simple/.envrc
```diff
@@ -1 +1,4 @@
-use flake --impure
+# https://github.com/nix-community/nix-direnv
+nix_direnv_watch_file devenv.nix
```
domenkozar, line 2: What happens if the user doesn't have nix-direnv installed?

domenkozar:
@thenonameguy added a few nitpicks, other than that we're good to go here! I'll merge it after devenv 0.5 will be released, hopefully tomorrow.

examples/process-compose/devenv.nix
domenkozar:
```
nix-repl> lib.flatten (builtins.map (package: builtins.map lib.getOutput package.outputs) [git vim])
[ «lambda @ /home/domen/dev/nixpkgs/lib/attrsets.nix:825:23» «lambda @ /home/domen/dev/nixpkgs/lib/attrsets.nix:825:23» «lambda @ /home/domen/dev/nixpkgs/lib/attrsets.nix:825:23» «lambda @ /home/domen/dev/nixpkgs/lib/attrsets.nix:825:23» ]
```

docs/reference/options.md
thenonameguy: I did not want to break the users, also there is no way of specifying ATM that `$USER` with a program fallback should be used with the `initialDatabases` syntax. It is determined at Nix eval time.
In my personal projects I almost always have a static database name locally, so it can be hardcoded for maximum ease of use. This way we don't have to pass the `$USER` everywhere.

thenonameguy:
@domenkozar thanks for the review and the great pointers!

I addressed all of them and updated both `realworld` devenv projects as a smoke test:
https://github.com/blrB/hanami-realworld-example-app/pull/1
https://github.com/snamiki1212/realworld-v1-rust-actix-web-diesel/pull/15

Both work well, see for yourself.
Please tell if you have any further changes in mind. :rocket:

domenkozar:
Nice work :rocket: 

I see that in the examples you're using nix flakes interface instead of devenv.nix. Is that a personal preference or is there something that plain devenv doesn't support?

docs/reference/options.md
domenkozar: If it makes sense to deprecate a flag you can do it like:

```
imports = [
    (lib.mkRenamedOptionModule [ "postgres" "enable" ] [ "services" "postgres" "enable" ])
  ];
```

examples/process-compose/devenv.nix
```diff
@@ -7,16 +7,28 @@
 
   services.postgres.enable = true;
 
+  languages.ruby.enable = true;
+
+  packages = [ pkgs.imagemagick6 ];
+
+  scripts.compile-rmagick.exec = "gem install --install-dir /tmp rmagick";
```
domenkozar, line 14: It would be better to move this into https://github.com/cachix/devenv/blob/main/examples/ruby/devenv.nix

templates/simple/.envrc
```diff
@@ -1 +1,12 @@
-use flake --impure
```
domenkozar, line 1: I'd put this into a separate PR, because I'd like to understand the difference and currently nix-direnv installation instructions are too confusing for mainstream adoption.

domenkozar (changes requested):
Almost there :)

thenonameguy:
> Nice work rocket
> 
> I see that in the examples you're using nix flakes interface instead of devenv.nix. Is that a personal preference or is there something that plain devenv doesn't support?

It's a mix of both. There is no way at the moment to easily package the project with Nix for advanced Nix users. For this you still have to use Flakes. I did not do the build packaging for that project though.
I expected this feedback, that's why I included a Flake-less example as well for the Ruby project.

templates/simple/.envrc
thenonameguy: Maybe as a middle-ground we could just source it automatically?
```
if ! has nix_direnv_version || ! nix_direnv_version 2.2.0; then
  source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/2.2.0/direnvrc" "sha256-5EwyKnkJNQeXrRkYbwwRBcXbibosCJqyIUuz9Xq+LRc="
fi
```

The speed benefits and superior caching of nix-direnv are IMO worth it, without it you get issues like:
https://github.com/cachix/devenv/issues/133

examples/process-compose/devenv.nix
thenonameguy: I don't want to deprecate the `createDatabase` flag just yet, to keep the amount of breaking changes minimal.
Could the `compile-rmagick` line stay here? It tests that the `scripts` are in the `PATH` of `process-compose`. Hitting 2 birds with 1 stone.

domenkozar: :+1:

templates/simple/.envrc
domenkozar: Nice, that works!

thenonameguy: Looks like this in practice:
```
$ cd realworld-v1-rust-actix-web-diesel 
direnv: loading ~/projects/realworld-v1-rust-actix-web-diesel/.envrc
nix-direnv: error current version v2.1.0 is older than the desired version v2.2.0
direnv: loading https://raw.githubusercontent.com/nix-community/nix-direnv/2.2.0/direnvrc (sha256-5EwyKnkJNQeXrRkYbwwRBcXbibosCJqyIUuz9Xq+LRc=)
direnv: nix-direnv: using cached dev shell
pre-commit-hooks.nix: hooks up to date
direnv: export +APIURL +C_INCLUDE_PATH +DATABASE_URL +DEVENV_DOTFILE +DEVENV_PROFILE +DEVENV_ROOT +DEVENV_STATE +FRONTEND_ORIGIN +IN_NIX_SHELL +LD_LIBRARY_PATH +LIBRARY_PATH +PC_HTTP_PORT +PC_TUI_ENABLED +PGDATA +PGHOST +PGPORT +PKG_CONFIG_PATH +name ~PATH ~XDG_CONFIG_DIRS ~XDG_DATA_DIRS
```

PR #192, opened 2022-12-06
Propagate the nixpkgs dependency to reduce the amount of nixpkgs instances.

R-VdP:


domenkozar:
I pushed a fix into `relaxed-flakes`, so if you `devenv update` and `nix flake update`, CI should pass again.

domenkozar:
It's weird this fails and https://github.com/NixOS/nix/pull/7367 succeeds. Needs investigation :)

R-VdP:
But that PR also removes the boehm patch, so I guess we'd have to do the same in the `relaxed-flakes` branch to get it to work with 22.11+.

domenkozar:
Pushed a removal of the boehm patch

Issue #154, opened 2022-11-30
Support unfree and overlays


domenkozar:
Wire up overlays and allowUnfree

https://nixos.wiki/wiki/Flakes

kubukoz:
My team is currently looking at switching to Devenv for a lower entry barrier (compared to raw flakes), overlays seem to be the remaining blocker atm. Do you have an idea on how this could work?

In particular, the overlays I want to add right now come from one of the inputs.

domenkozar:
I'll work on this next then, thanks :)

domenkozar:
@kubukoz @smunix @pikdum @shyim @jochenmanz @arjan-s @mstine See https://github.com/cachix/devenv/pull/383

PR #193, opened 2022-12-06
flake: avoid creating another instance of nixpkgs

zimbatm:
Related to https://zimbatm.com/notes/1000-instances-of-nixpkgs
and https://github.com/numtide/nixpkgs-unfree/issues/12

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -40,7 +40,7 @@
     {
       packages = forAllSystems (system:
         let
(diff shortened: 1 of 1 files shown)


domenkozar:
If one wanted to pass `allowUnfree` as in #154, would that still work?

zimbatm:
I think the best for that is to replace the nixpkgs flake input with `github:numtide/nixpkgs-unfree`.

PR #200, opened 2022-12-12
move all services in their prefix

zimbatm:
This makes it easier to find out which modules are to declare new
processes. It also follows the convention set by NixOS that services
should be prefixed with "services.".


domenkozar:
While it's early days, let's still use renaming hints so devenv users get a nice error with the help text to migrate over.

Also seems that the tests fail because not everything was renamed

zimbatm:
Cool, I wanted to check if you would accept the PR before doing the extra work.

zimbatm:
@domenkozar do you think it's enough to alias the enable option? I'm hoping that the user gets the hint and adjusts the other options on their own.

domenkozar:
Sounds good!

zimbatm:
Ok. Docs fixed, options aliased, examples fixed. Should be good to go.

PR #202, opened 2022-12-12
ruby: update GEM_HOME, GEM_PATH and PATH

bobvanderlinden:
Currently Ruby works well with bundler in devenv. However, `bundle exec` is always needed to set the environment correctly.

Some tools require calling the Ruby-gem executabeles directly without `bundle exec`. For instance, these call `rails` instead of `bundle exec rails`. This will fail, because `rails` is not part of `PATH`.

This PR attempts to set `GEM_HOME` to the right directory of the bundler state directory, which holds the gems. It also prepends the gems directory to `GEM_PATH` and prepends the gem-bin directory to `PATH`.

I was able to do:

```console
$ devenv shell
$ bundle
$ rails new .
$ rails server
```


domenkozar:
Needs resolved config :)

domenkozar:
It would be great to also add nixpkgs-ruby to the ruby example :)

PR #213, opened 2022-12-16
Add the correct RUST_SRC_PATH

farcaller:
Required for the rust-analyzer to work.

It's a bit of a questionable approach, but I feel like hacking around fenix's env generation would be even more tricky. At least it's explicit this way.


domenkozar:
I'd set this by default in https://github.com/cachix/devenv/blob/main/src/modules/languages/rust.nix#L44

farcaller:
@domenkozar done.

PR #219, opened 2022-12-19
Fix the default template name in the docs

farcaller:
It does seem there's no default.


thenonameguy:
My PR addresses this problem by fixing the implementation vs the docs:
https://github.com/cachix/devenv/pull/191/files#r1051900899

IMO the former is preferable.

farcaller:
Fair point!

PR #222, opened 2022-12-20
add starship integration

alejandrosame:
This PR adds a module to integrate the cross-shell prompt `Starship`.

The example inside the `examples` folder provides basic documentation about the options available in this module. The comments can easily be copied into a .md file to add it to the documentation website once the integration is reviewed and merged.

### DIFF
--- a/src/modules/integrations/starship.nix
+++ b/src/modules/integrations/starship.nix
@@ -1,0 +1,45 @@
+{ pkgs, config, lib, ... }:
+
+{
+  options.starship = {
+    enable = lib.mkEnableOption "Enable the Starship command prompt.";
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.starship;
+      defaultText = "pkgs.starship";
+      description = "The Starship package to use.";
+    };
+
+    config.enable = lib.mkEnableOption "Enable Starship config override.";
+
+    config.path = lib.mkOption {
+      type = lib.types.path;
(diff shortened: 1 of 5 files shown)


src/modules/integrations/starship.nix
```diff
@@ -0,0 +1,45 @@
...
+    enable = lib.mkEnableOption "Enable the Starship command prompt.";
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.starship;
+      defaultText = "pkgs.starship";
+      description = "The Starship package to use.";
+    };
+
+    config.enable = lib.mkEnableOption "Enable Starship config override.";
```
shyim, line 14: Maybe manage the starship config also here :)

alejandrosame: If you mean adding there individual entries to write the file for the user, I purposefully left this to be handled by Starship itself because:

- I don't want to remake work that was already done by [home-manager](https://github.com/nix-community/home-manager/blob/7ae7250df8f38c4efc8cd6669a36272ab7a575ed/modules/programs/starship.nix). I need to research more about how to reuse modules in the whole Nix ecosystem.

- This PR is meant to give devenv users a quick way to have a `"virtualenv"` aware PS1, and they can customize it further by providing the Starship TOML with their project. Honestly, originally I just wanted to add `starship.enable` but, since Starship's logic for the prompt is too naive at times, it is useless/frustating to provide it without a way to change the prompt options.

PR #226, opened 2022-12-21
mysql: fix missing coreutils

### LINKED ISSUE #223: processes.mysql-configure.exec needs gsleep on mac
**Describe the bug**
devenv up shuts down immediately when ensureUsers uses sleep instead of gsleep on mac. 

**To Reproduce**
services.mysql.enable = true;
  services.mysql.initialDatabases = [{ name = "test"; }];
  services.mysql.ensureUsers = [
    {
      name = "test";
      password = "test";
      ensurePermissions = { "test.*" = "ALL PRIVILEGES"; };
    }
  ];

**Version**
 devenv version
devenv: 0.4

shyim:
`sleep infinity` does not work on macOS. So I adjusted the path to use always `coreutils`. fixes #223

### DIFF
--- a/src/modules/services/mysql.nix
+++ b/src/modules/services/mysql.nix
@@ -20,6 +20,7 @@ let
     exec ${cfg.package}/bin/mysqld ${mysqldOptions}
   '';
   configureScript = pkgs.writeShellScriptBin "configure-mysql" ''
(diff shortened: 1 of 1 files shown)


shyim:
Fixed also ensure users on MariaDB platform.

PR #228, opened 2022-12-21
rust: feature parity with fenix

szlend:
The rust language support in devenv has two "backends" (nixpkgs and fenix). It's surprising to me that setting the `languages.rust.version` changes the backend. I think ideally this should be more obvious to the user, or there should just be a single backend for simplicity's sake. I have to wonder whether maintaining both is worth the effort.

Regardless, I ported the missing functionality into the nixpkgs backend, so now rust-analyzer and pre-commit hooks work there as well. I also added `rustfmt`, `clippy` and `rust-analyzer` to `PATH` so they can be picked up by editors (i.e. vscode). All packages are now overridable (e.g. i can use the nightly rustfmt with the stable rust toolchain).

You can override a package like so:
```nix
{ languages.rust.packages.rustfmt = inputs.fenix.packages.${pkgs.system}.latest.rustfmt; }
```

### DIFF
--- a/src/modules/languages/rust.nix
+++ b/src/modules/languages/rust.nix
@@ -19,8 +19,16 @@ in
     enable = lib.mkEnableOption "Enable tools for Rust development.";
 
     packages = lib.mkOption {
-      type = lib.types.submodule ({ ... }: {
-        options = genAttrs tools (name: lib.mkOption {
+      type = lib.types.submodule ({
+        options = {
+          rust-src = lib.mkOption {
+            type = lib.types.either lib.types.package lib.types.str;
+            default = pkgs.rustPlatform.rustLibSrc;
+            defaultText = "pkgs.rustPlatform.rustLibSrc";
+            description = "rust-src package";
+          };
+        }
+        // genAttrs tools (name: lib.mkOption {
           type = lib.types.package;
           default = pkgs.${name};
           defaultText = "pkgs.${name}";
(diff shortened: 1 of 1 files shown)


src/modules/languages/rust.nix
```diff
@@ -31,10 +32,13 @@ in
...
-        cfg.packages.rustc
-        cfg.packages.cargo
-      ];
+      packages = attrValues (getAttrs [ "rustc" "cargo" "rustfmt" "clippy" "rust-analyzer" ] cfg.packages);
+      pre-commit.tools.cargo = lib.mkForce cfg.packages.cargo;
+      pre-commit.tools.rustfmt = lib.mkForce cfg.packages.rustfmt;
+      pre-commit.tools.clippy = lib.mkForce cfg.packages.clippy;
+    })
+    (lib.mkIf (cfg.version == null) {
+      env.RUST_SRC_PATH = pkgs.rustPlatform.rustLibSrc;
```
szlend, line 41: I wonder if this should be configurable. If you can specify a custom `rustc` package, then you would likely want to set this as well.

src/modules/languages/rust.nix
```diff
@@ -31,23 +34,21 @@ in
...
         rustPackages = fenix.packages.${pkgs.system}.${cfg.version} or (throw "languages.rust.version is set to ${cfg.version}, but should be one of: stable, beta or latest.");
       in
       {
-        languages.rust.packages = rustPackages;
-
-        pre-commit.tools.cargo = lib.mkForce rustPackages.cargo;
-        pre-commit.tools.rustfmt = lib.mkForce rustPackages.rustfmt;
-        pre-commit.tools.clippy = lib.mkForce rustPackages.clippy;
-
+        languages.rust.packages = genAttrs tools (package: lib.mkDefault rustPackages.${package});
```
szlend, line 51: With `mkDefault` you can now also override fenix packages.

src/modules/languages/rust.nix
```diff
@@ -17,7 +20,7 @@ in
 
     packages = lib.mkOption {
       type = lib.types.attrsOf lib.types.package;
-      default = { inherit (pkgs) rustc cargo; };
+      default = getAttrs tools pkgs;
```
szlend, line 23: This needs to be a submodule, otherwise it's hard to override a single package.

szlend: Fixed.

src/modules/languages/rust.nix
szlend: Done

PR #237, opened 2022-12-23
feat(nix-direnv): integrate nix-direnv with bare devenv usage [WIP]

thenonameguy:


examples/simple/.envrc
```diff
@@ -1,4 +1,8 @@
-watch_file devenv.nix 
-watch_file devenv.yaml 
-watch_file devenv.lock
-eval "$(devenv print-dev-env)"
\ No newline at end of file
+if ! has nix_direnv_version || ! nix_direnv_version 2.2.0; then
+  source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/2.2.0/direnvrc" "sha256-5EwyKnkJNQeXrRkYbwwRBcXbibosCJqyIUuz9Xq+LRc="
+fi
+
```
domenkozar, line 4: Should we here add something like `$(which devenv)/bin` to $PATH?

thenonameguy: At this point, `devenv` is not on the PATH yet, if you assume the user hasn't installed it on their machine. So that would actually fail.

domenkozar:
I'd really like to first tackle #240 and then see if we can ignore the whole Nix business.

Otherwise this would be a good compromise :)

domenkozar:
Hey @thenonameguy, thanks for this work. It took me a while, but I've made a devenv specific direnv integration, based on nix-direnv.

It differs that it can track imports of devenv.yaml and simplifies a lot of code since it's devenv aware.

See https://github.com/cachix/devenv/pull/387

domenkozar:
@thenonameguy @fschmtt @janvotava I'm going to close this as #387 will land soon. Thanks again for moving this forward!

PR #268, opened 2023-01-02
add workaround for tmpdir errors on Darwin, fixes #255

### LINKED ISSUE #255: TMPDIR is set to null
**Describe the bug**

When you switch your process implementation a different one, MySQL and PHP are broken.

The issue is that `TMPDIR` is set to string `null`. This seems not to happen with the default implementation, as it sources variables from a file.

shyim:
Darwin always sets a `TMPDIR`. As the mkNakedShell removes the shell the `TMPDIR` gets `null` and breaks a lot of applications like VSCode with Direnv.

Before:

image

After:
image

### DIFF
--- a/src/modules/mkNakedShell.nix
+++ b/src/modules/mkNakedShell.nix
@@ -26,6 +26,7 @@
 , coreutils
 , system
 , writeTextFile
+, pkgs
 , lib
 }:
 let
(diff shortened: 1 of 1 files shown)


shyim:
I guess some pipelines needs to be skipped when the PR is not in cachix org :)

domenkozar:
Could you rebase on top of main to see if that fixes CI?

shyim:
seems to work

PR #278, opened 2023-01-06
Add Unison language

ereslibre:


src/modules/languages/unison.nix
```diff
@@ -0,0 +1,23 @@
+{ pkgs, config, lib, ... }:
+
+let
+  cfg = config.languages.unison;
+in
+{
+  options.languages.unison = {
+    enable = lib.mkEnableOption "Enable tools for Unison development.";
+
+    package = lib.mkOptionn {
```
domenkozar, line 10:
```suggestion
    package = lib.mkOption {
```

ereslibre: Ooops, I have no idea how that happened!

domenkozar:
It's now complaining that `unison-code-manager` is marked as unfree, do you know if that's correct?

domenkozar:
See https://github.com/cachix/devenv/actions/runs/5377840454/jobs/9756883028#step:8:442

ereslibre:
> so it seems like we need to add back arm/linux support.

I don't think it ever had it. The binaries they provide with their automation for Linux is x86_64 only.

From: https://www.unison-lang.org/learn/quickstart/#installation-options, "The current alpha release is for Mac OS X, 64-bit Linux, and Windows users!"

PR #282, opened 2023-01-08
Enable default users for mysql commands via mysqlOptions

jochenmanz:
Enables users to set default user and passwords for mysql binaries that are automatically used.

```
  services.mysql.settings = {
    mysql = {
      user = "test";
      password = "test";
    };
    mysqldump = {
      user = "test";
      password = "test";
    };
  };
```
  
Without the changes the configure scripts breaks and the service cant be startet.

### DIFF
--- a/src/modules/services/mysql.nix
+++ b/src/modules/services/mysql.nix
@@ -23,7 +23,7 @@ let
     PATH="${lib.makeBinPath [ cfg.package pkgs.coreutils ]}:$PATH"
     set -euo pipefail
 
-    while ! ${cfg.package}/bin/mysqladmin ping ${lib.optionalString (!isMariaDB) "-u root --password=''"} --silent; do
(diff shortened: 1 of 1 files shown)


domenkozar:
cc @shyim

src/modules/services/mysql.nix
```diff
@@ -23,7 +23,7 @@ let
     PATH="${lib.makeBinPath [ cfg.package pkgs.coreutils ]}:$PATH"
     set -euo pipefail
 
-    while ! ${cfg.package}/bin/mysqladmin ping ${optionalString (!isMariaDB) "-u root"} --silent; do
+    while ! ${cfg.package}/bin/mysqladmin ping ${optionalString (!isMariaDB) "-u root --password=''"} --silent; do
```
shyim, line 26: Did you try it on MariaDB? I guess when you do the similar config there, this should break. 
Maybe we need the same for that too

PR #283, opened 2023-01-09
rust: link frameworks on darwin

### LINKED ISSUE #267: Rust language can't link with Security framework on darwin
**Describe the bug**
When compiling some rust code which requires openssl or other Security-related dependencies compilation fails with 
```
  = note: ld: framework not found Security
          clang-11: error: linker command failed with exit code 1 (use -v to see invocation)
```

**To Reproduce**
https://gist.github.com/onsails/d2d33a63d88a5e7d656048bd5fa5a6b9.

**Version**
 
0.5

onsails:
fix #267

### DIFF
--- a/examples/rust/devenv.nix
+++ b/examples/rust/devenv.nix
@@ -1,6 +1,10 @@
-{ pkgs, ... }:
+{ pkgs, lib, ... }:
 
 {
+  packages = lib.optionals pkgs.stdenv.isDarwin (with pkgs.darwin.apple_sdk; [
+    frameworks.Security
+  ]);
+
   languages.rust = {
     enable = true;
     # https://devenv.sh/reference/options/#languagesrustversion
--- a/src/modules/languages/rust.nix
+++ b/src/modules/languages/rust.nix
@@ -60,6 +60,10 @@ in
       pre-commit.tools.rustfmt = lib.mkForce cfg.packages.rustfmt;
       pre-commit.tools.clippy = lib.mkForce cfg.packages.clippy;
     })
+    (lib.mkIf pkgs.stdenv.isDarwin {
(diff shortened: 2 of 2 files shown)


examples/rust/devenv.nix
```diff
@@ -1,6 +1,10 @@
-{ pkgs, ... }:
+{ pkgs, lib, ... }:
 
 {
+  packages = [ ] ++ lib.optionals pkgs.stdenv.isDarwin (with pkgs.darwin.apple_sdk; [
```
domenkozar, line 4:
```suggestion
  packages = lib.optionals pkgs.stdenv.isDarwin (with pkgs.darwin.apple_sdk; [
```

src/modules/languages/rust.nix
```diff
@@ -60,6 +60,10 @@ in
       pre-commit.tools.rustfmt = lib.mkForce cfg.packages.rustfmt;
       pre-commit.tools.clippy = lib.mkForce cfg.packages.clippy;
     })
+    (lib.mkIf pkgs.stdenv.isDarwin {
+      env.RUSTFLAGS = [ "-L framework=${config.env.DEVENV_PROFILE}/Library/Frameworks" ];
+      env.RUSTDOCFLAGS = [ "-L framework=${config.env.DEVENV_PROFILE}/Library/Frameworks" ];
```
smunix, line 65: My haskell project broke after an upgrade of devenv (using flake). I tracked the failure down to an infinite recursion introduced by this line on Darwin.
@onsails can you have another look at this fix?

smunix (commented):
It breaks devenv on darwin

src/modules/languages/rust.nix
domenkozar: Could you provide a minimal example to reproduce?

domenkozar: Fixed in master, thanks!

PR #284, opened 2023-01-10
fix: avoid infinite recursion

roman:
Remove the `DEVENV_PROFILE` entry from the processes `envList` set to avoid infinite recursions.

This happens when a devenv module uses the `config` parameter inside a scripts entry _and_ there are custom `processes` definitions. This is very common on projects that use devenv via the [nix flakes porcelain](https://github.com/cachix/devenv/blob/main/docs/guides/using-with-flakes.md).

## Reproducing infinite recursion

```nix
# devenv.nix
{ config, ... }: {
  processes.boom.exec = "echo boom";
  scripts.devenv-up.exec = ''
    ${config.procfileScript}
  '';
}
```

### DIFF
--- a/src/devenv-devShell.nix
+++ b/src/devenv-devShell.nix
@@ -32,6 +32,8 @@ pkgs.writeScriptBin "devenv" ''
       ;;
     *)
       echo "https://devenv.sh (version ${version}): Fast, Declarative, Reproducible, and Composable Developer Environments"
+      echo 
+      echo "This is a flake integration wrapper that comes with a subset of functionality from the flakeless devenv CLI."
       echo
       echo "Usage: devenv command"
       echo
--- a/src/devenv-devShell.nix
+++ b/src/devenv-devShell.nix
@@ -1,0 +1,45 @@
+{ config, pkgs }:
+let
+  lib = pkgs.lib;
+  version = lib.fileContents ./modules/latest-version;
+in
+pkgs.writeScriptBin "devenv" ''
+  #!/usr/bin/env bash
+
+  # we want subshells to fail the program
+  set -e
+
+  NIX_FLAGS="--show-trace --extra-experimental-features nix-command --extra-experimental-features flakes"
+
+  command=$1
+  if [[ ! -z $command ]]; then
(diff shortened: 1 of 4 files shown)


thenonameguy (approved):
Makes sense, thanks!

domenkozar:
We should introduce and conditionalize this on something like `devenv.flakesIntegration` since depending on the profile is useful within processes otherwise.

roman:
@domenkozar, that's a great idea, I added a new commit that adds a `config.devenv.flakesIntegration` option, not sure if this is the way to do it, so I'll wait for your guidance. 

In another note, I'm wondering if [including the `devenv` binary](https://github.com/cachix/devenv/blob/77297e2f3fefabe18a3077c223af0988a9c455bf/flake.nix#L88) when using `devenv.lib.mkShell` makes sense, this binary always assumes that `devenv` is being used without `devenv.lib.mkShell`.

Would it make sense to create a package similar to `src/devenv.nix` that assumes that we are running the `devenv` infra from a `flake.nix` file? I'm primarily interested in the `devenv up` command.

roman:
I went ahead and added this new binary (`src/devenv-devShell.nix`) that replicates _some_ of the `devenv` sub-commands. 

Mind you. This last commit is just a proposal; if you don't like it, I can remove it, or if you want me to create a new PR for it, we can do that as well.

src/devenv-devShell.nix
```diff
@@ -0,0 +1,45 @@
...
+        exit 1
+      else
+        $procfilescript
+      fi
+      ;;
+    version)
+      echo "devenv: ${version}"
+      ;;
+    *)
+      echo "https://devenv.sh (version ${version}): Fast, Declarative, Reproducible, and Composable Developer Environments"
```
domenkozar, line 34:
```suggestion
      echo "https://devenv.sh (version ${version}): Fast, Declarative, Reproducible, and Composable Developer Environments"
      echo 
      echo "This is a flake integration wrapper that comes with a subset of functionality from the flakeless devenv CLI."
```

PR #285, opened 2023-01-10
languages/r: add option for R package

adfaure:
Hello,

I added a config option to specify the R package. It is similar at how it is done for python.
I did that to be able to add additional dependencies to my R installation using  `rWrapper` such as:
```nix
  languages.r = {
    enable = true;
    package = (pkgs.rWrapper.override{packages = with pkgs.rPackages; [ viridis tidyverse ];});
  };
```

### DIFF
--- a/src/modules/languages/r.nix
+++ b/src/modules/languages/r.nix
@@ -6,11 +6,17 @@ in
 {
   options.languages.r = {
     enable = lib.mkEnableOption "Enable tools for R development.";
+    package = lib.mkOption {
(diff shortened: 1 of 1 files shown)


luerhard:
This should be somewhere in the documentation :) Took me quite a while to find out how to add r-packages to devenv!

domenkozar:
We could add another option for that:

```nix
 languages.r = {
    enable = true;
    withPackages = r: [ r.viridis r.tidyverse ];
  };
```

PR #295, opened 2023-01-15
caddy: fix unused dataDir option

emilytrau:
The `services.caddy.dataDir` option was not being applied causing caddy to pollute state in the default locations - `~/.local/share/caddy` and `~/.config/caddy`


domenkozar:
cc @shyim heads up

shyim (approved):

PR #296, opened 2023-01-15
languages/python: add virtual environment option

bobvanderlinden:
Currently Python always needs manually creating and activating a virtual environment. This can be done using `virtualenv`, but also `python -m venv venv`. It would be nice if this was done upon entering the devenv shell.

I previously also had poetry on top of this, but I thought it was better to split it up. First venv, maybe later poetry.

This also adds an example named `python-venv` to show how it would work.

### DIFF
--- a/examples/python-venv/devenv.nix
+++ b/examples/python-venv/devenv.nix
@@ -1,0 +1,6 @@
+{ pkgs, ... }:
+
+{
+  languages.python.enable = true;
+  languages.python.venv.enable = true;
+}
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -25,7 +25,10 @@ in
     env.PYTHONPATH = "${config.env.DEVENV_PROFILE}/${cfg.package.sitePackages}";
 
     enterShell = lib.mkIf cfg.venv.enable ''
-      python -m venv ${config.env.DEVENV_STATE}/venv
+      if [ ! -d ${config.env.DEVENV_STATE}/venv ]
+      then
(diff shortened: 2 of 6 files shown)


examples/python-venv/devenv.nix
```diff
@@ -0,0 +1,6 @@
+{ pkgs, ... }:
+
+{
+  languages.python.enable = true;
+  languages.python.venv.enable = true;
```
domenkozar, line 5: I really like a bit more descriptive names like `createVirtualenv` since the expectation of what will happen is clearer.

bobvanderlinden: Should it have two options then? Once for automatically creating a virtualenv and one for activating?

src/modules/languages/python.nix
```diff
@@ -21,5 +23,10 @@ in
     ];
 
     env.PYTHONPATH = "${config.env.DEVENV_PROFILE}/${cfg.package.sitePackages}";
+
+    enterShell = lib.mkIf cfg.venv.enable ''
+      python -m venv ${config.env.DEVENV_STATE}/venv
```
domenkozar, line 28: This takes about 4-5s on my machine, should we skip it if it exists?

bobvanderlinden: I tried the following:

```
time python -m venv ${config.env.DEVENV_STATE}/venv
```

```
real	0m1.293s
user	0m1.230s
sys	0m0.063s
```

It wasn't that noticeable on my laptop, but it is indeed unnecessary. I've added a check so that it only creates venv if the directory doesn't exist. :+1:

PR #297, opened 2023-01-15
languages/python: add poetry support

bobvanderlinden:
Depends on #296 

In #296, venv support was added. This PR adds poetry support on top. It automatically runs `poetry install` upon entering the devenv shell if there is a pyproject.toml and a poetry.lock.

It gives instructions when pyproject.toml or poetry.lock are missing.

Lastly, running `poetry install` on the example took 0.7s on my machine each time the shell was entered. I added 'caching' that first compares the checksum of poetry.lock before running `poetry install`.

It is possible to use poetry without venv, but by default when poetry is enabled, venv is enabled as well.

The example shows its usage by installing numpy using poetry.


bobvanderlinden:
@domenkozar I think this PR is ready for review. Could you give your opinion?

bobvanderlinden:
Thanks!

PR #298, opened 2023-01-15
rust: link frameworks on darwin only if rust is enabled

onsails:
fix [this](https://github.com/cachix/devenv/pull/283/files#r1070561727)


onsails:
already fixed on main

PR #301, opened 2023-01-16
feat: make composer optional

shyim:
Allows disabling composer and adding own using packages

### DIFF
--- a/src/modules/languages/php.nix
+++ b/src/modules/languages/php.nix
@@ -144,6 +144,12 @@ in
   options.languages.php = {
     enable = lib.mkEnableOption "tools for PHP development";
 
+    enableComposer = lib.mkOption {
+      type = lib.types.bool;
+      description = lib.mdDoc "Enable composer";
+      default = true;
+    };
+
     package = lib.mkOption {
       type = lib.types.package;
       default = pkgs.php;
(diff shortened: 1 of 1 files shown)


domenkozar:
Same as #315, this should be called `languages.php.packages.composer` that can be set to `null`.

kkirsche:
It sounds like that would be this piece of configuration from Rust:
https://github.com/cachix/devenv/blob/main/src/modules/languages/rust.nix#L31-L36 if I understood @domenkozar's comment in #315 and the languages file structure properly

src/modules/languages/php.nix
```diff
@@ -162,6 +162,22 @@ in
       '';
     };
 
+    packages = lib.mkOption {
+      type = lib.types.submodule ({
+        options = {
+          composer = lib.mkOption {
+            type = lib.types.nullOr lib.types.str;
+            default = "composer";
```
domenkozar, line 170: the type should be `types.package` and the default `cfg.package.packages.composer`.

shyim: I have no idea how it should work with package 🤔 . I need to point to the `cfg.package.packages.XXXXX`

domenkozar: This should work:

```nix
config = {
...
  languages.php.composer = lib.mkDefault cfg.pacakge.packages.composer;
...
}
```

shyim:
@domenkozar refactored it like you said and works :)

shyim:
macos fails randomly installing nix https://github.com/cachix/devenv/actions/runs/4102872885/jobs/7078523073 🙈

PR #306, opened 2023-01-18
services/minio: add buckets option

shyim:
```nix
services.minio.enable = true;
services.minio.buckets = [ "some-test" ];
```

Allows me to create buckets on start.

image

### DIFF
--- a/src/modules/services/minio.nix
+++ b/src/modules/services/minio.nix
@@ -12,6 +12,10 @@ let
       mkdir -p "$MINIO_CONFIG_DIR"
     fi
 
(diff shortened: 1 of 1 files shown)


src/modules/services/minio.nix
```diff
@@ -72,6 +76,14 @@ in
       type = types.package;
       description = lib.mdDoc "Minio package to use.";
     };
+
+    ensureBuckets = lib.mkOption {
```
domenkozar, line 80: We could call it just `buckets` as all devenv settings are about ensuring the state :)

shyim: done :)

PR #312, opened 2023-01-20
add Swift language support

domenkozar:
Thanks to @stephank, we can support Swift as soon as it lands in nixpkgs-unstable!

I'm planning to test on https://github.com/PerfectlySoft/PerfectTemplate.git once Hydra catches up.

### DIFF
--- a/src/modules/languages/swift.nix
+++ b/src/modules/languages/swift.nix
@@ -1,0 +1,28 @@
+{ pkgs, config, lib, ... }:
+
+let
+  cfg = config.languages.swift;
+in
+{
+  options.languages.swift = {
+    enable = lib.mkEnableOption "tools for Swift development";
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.swift;
+      defaultText = "pkgs.swift";
+      description = ''
+        The Swift package to use.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    packages = with pkgs; [
+      cfg.package
+      pkgs.clang
+    ];
+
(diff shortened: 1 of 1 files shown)


domenkozar:
@stephank, do you have an idea what's a good starter repo to get this with? The one I link it doesn't compile for me at all.

stephank:
What's the error you're running into? Some things from my testing:

- Using the Xcode tools, it builds fine.
- I've been doing mostly release builds (`swift build -c release`), because that's what we package in nixpkgs. But apparently, debug builds (the default) fail because they try to use clang `-index-store-path`, which is an Apple extension. Passing `--disable-index-store` fixes that, but I wonder if we should make that the default for Swift in nixpkgs.
- The compile error I'm running into:
  ```
  PerfectTemplate/.build/checkouts/PerfectLib/Sources/PerfectLib/Log.swift:73:28: error: cannot convert value of type 'CVaListPointer' to expected argument type '__darwin_va_list?' (aka 'Optional<UnsafeMutablePointer<Int8>>')
                          vsyslog(priority, "%s", $0)
  ```
  Not something I've seen before, so don't have an immediate answer.

domenkozar:
Enabling `--disable-index-store` by default makes sense to me, given that compilation fails otherwise.

I'm going to wait until staging is merged to master and retest it then.

stephank:
https://github.com/NixOS/nixpkgs/pull/213936 fixes the index store issues, but not yet the compile issue. There's something going wrong in the type mapping when importing C headers, it seems.

domenkozar:
Seems like we're quite close now, waiting for a channel update.

Do you have a good example of a swift project to test with?

PR #317, opened 2023-01-21
add self to flake.nix

smunix:
Allow access to local packages if `devenv init` is run in a flake repository

### DIFF
--- a/src/flake.nix
+++ b/src/flake.nix
@@ -9,7 +9,7 @@
          then (builtins.fromJSON (builtins.readFile ./.devenv/devenv.json)).inputs
          else {});
 
-    outputs = { nixpkgs, ... }@inputs:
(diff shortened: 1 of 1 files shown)


domenkozar:
I'm not sure I understand how this would be used. Did you try it?

smunix:
> I'm not sure I understand how this would be used. Did you try it?

Yes, I did (see details in [this discord thread](https://discord.com/channels/1036369714731036712/1036369758024650792/1066357307115581511))

Basically, in `devenv.nix` I can now write this:

image

`devenv info` now works, outside of `flake.nix`:
image

domenkozar:
You should be able to access `inputs.self` without this change.

smunix:
> You should be able to access `inputs.self` without this change.

This is awesome! I tried without this change and it works indeed. I'll close this PR now, it's not needed at all

PR #318, opened 2023-01-21
correct mariadb database and user handling

jochenmanz:
correct ensured users and database creation that i might have broken before 🙈 

config used for validation:

```
services.mysql.enable = true;
  services.mysql.package = pkgs.mariadb;
  services.mysql.initialDatabases = lib.mkDefault [{ name = "test_database"; }];
  services.mysql.ensureUsers = lib.mkDefault [
    {
      name = "test_user";
      password = "test_password";
      ensurePermissions = { "*.*" = "ALL PRIVILEGES"; };
    }
  ];
services.mysql.settings = {
    mysql = {
      user = "test";
      password = "test";
    };
    mysqldump = {
      user = "test";
      password = "test";
    };
    mysqladmin = {
      user = "test";
      password = "test";
    };
  };
```

### DIFF
--- a/src/modules/services/mysql.nix
+++ b/src/modules/services/mysql.nix
@@ -14,7 +14,11 @@ let
 
     if [[ ! -d "$MYSQL_HOME" ]]; then
       mkdir -p "$MYSQL_HOME"
-      ${if isMariaDB then "${cfg.package}/bin/mysql_install_db" else "${cfg.package}/bin/mysqld"} ${mysqldOptions} ${optionalString (!isMariaDB) "--initialize-insecure"}
+      ${
+        if isMariaDB
+          then "${cfg.package}/bin/mysql_install_db ${mysqldOptions} --auth-root-authentication-method=normal"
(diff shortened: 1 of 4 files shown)


src/modules/services/mysql.nix
```diff
@@ -23,16 +27,16 @@ let
     PATH="${lib.makeBinPath [ cfg.package pkgs.coreutils ]}:$PATH"
     set -euo pipefail
 
-    while ! ${cfg.package}/bin/mysqladmin ping ${optionalString (!isMariaDB) "-u root --password=''"} --silent; do
+    while ! MYSQL_PWD="" ${cfg.package}/bin/mysqladmin ping  -u root --silent; do
```
jochenmanz: could we refactor this as a global (as in current shell) function so everyone can reuse this?

domenkozar:
cc @shyim

shyim (approved):
tested works. the example config user was just wrong.

src/modules/services/mysql.nix
shyim: I guess that won't work so easy. Maybe exposing it somehow with nix?

PR #319, opened 2023-01-23
processes: quote env vars

domenkozar:
Otherwise

```nix
{
  env = {
  ENVVAR = "Hello world";
};
processes.example.exec = "echo $ENVVAR";
}
```

results into

```
14:56:13 system    | example.1 started (pid=1518566)
14:56:13 example.1 | Helloworld
14:56:13 system    | example.1 stopped (rc=0)
```

### DIFF
--- a/src/modules/processes.nix
+++ b/src/modules/processes.nix
@@ -35,7 +35,7 @@ let
   implementation-options = config.process.${implementation};
   envList =
     lib.mapAttrsToList
-      (name: value: "${name}=${toString value}")
+      (name: value: "${name}=${builtins.toJSON value}")
       (if config.devenv.flakesIntegration then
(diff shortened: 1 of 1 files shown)


src/modules/processes.nix
```diff
@@ -35,7 +35,7 @@ let
   implementation-options = config.process.${implementation};
   envList =
     lib.mapAttrsToList
-      (name: value: "${name}=${toString value}")
+      (name: value: ''${name}="${toString value}"'')
```
bobvanderlinden, line 38: Might be even safer to use `toJSON`? Not sure what other side-effects `toJSON` will bring that aren't compatible with bash, but it at least makes so that `"` inside the value are escaped.

```suggestion
      (name: value: "${name}=${toJSON value}")
```

PR #320, opened 2023-01-23
User can specify SQL to run during initialization

### LINKED ISSUE #313: Add ability to run some commands during initialization of Postgres
I wanted to run this by you before writing up a PR. I would like to be able to do something like `CREATE USER postgres SUPERUSER` the first time Postgres initializes. Is this something that you would like supported within devenv? If so, I can throw together a PR.

ankhers:
This change will allow users to specify SQL commands to run when the database is first initialized. An example use of this would to create database user(s). Though, you can run any arbitrary SQL.

resolves #313

### DIFF
--- a/src/modules/services/postgres.nix
+++ b/src/modules/services/postgres.nix
@@ -42,6 +42,14 @@ let
       lib.optionalString cfg.createDatabase ''
         echo "CREATE DATABASE ''${USER:-$(id -nu)};" | postgres --single -E postgres '';
 
+  runInitialScript =
+    if cfg.initialScript != null then
+      ''
+        echo "${cfg.initialScript}" | postgres --single -E postgres
+      ''
+    else
+      "";
+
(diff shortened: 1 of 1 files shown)


ankhers:
> maybe would be cool to have something like ensureUsers like in MySQL module

That was what I was going to do initially, but there are currently 24
different options[0] that can go into a create user statement in
postgres (I don't use mysql, not sure the differences). So I thought
instead of trying to create a whole bunch of options that tries to
encapsulate everything you would want to do, it would just be better to
let people create users the way that would most benefit them.

This also has the benefit of not just creating users, but you can run
any arbitrary SQL during database initilization. This could also be
creating groups or roles or anything that someone would want to do.

0 - https://www.postgresql.org/docs/15/sql-createuser.html

domenkozar:
In any case it's useful to have this hook!

PR #322, opened 2023-01-24
Allow the user to use a different version of OCaml

ankhers:
This will allow the user to specify which version of OCaml and associated tools that they would like to use for their project.

This will also include a few extra common tools within the OCaml ecosystem.

### DIFF
--- a/src/modules/languages/ocaml.nix
+++ b/src/modules/languages/ocaml.nix
@@ -6,13 +6,25 @@ in
 {
   options.languages.ocaml = {
     enable = lib.mkEnableOption "tools for OCaml development";
+
+    packages = lib.mkOption
+      {
+        type = lib.types.attrs;
+        description = "The package set of OCaml to use";
+        default = pkgs.ocaml-ng.ocamlPackages;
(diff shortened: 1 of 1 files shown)


domenkozar:
I reverted it since it fails on darwin, probably we need a different version of ocamlPackages?

ankhers:
Let me pull out a mac and see if I can reproduce this.

domenkozar:
https://github.com/cachix/devenv/actions/runs/3997813963/jobs/6860703321#step:6:1889

ankhers:
I'm not sure what to say about this. I just tested on the only mac I have access to and it seems to work. Keep in mind this machine is apparently on an old-ish mac version (11.x). I am updating to the most recent version (13.x) and will test again once that is complete.

ankhers:
I can't seem to replicate this even after updating to the most recent version of macos. Is there any chance this was a flaky test in some way? Does anyone else have a mac that they can test this on?

domenkozar:
@ankhers could you open a new PR?

ankhers:
#379

PR #328, opened 2023-01-25
Files and variables language edit

sixhobbits:


domenkozar:
I've pushed manually since there we conflicts. Thanks!

PR #329, opened 2023-01-25
Options.md language edit

sixhobbits:
Line 4420: "attribute set of attribute set of (INI atom (null, bool, int, float or string) or a list of them for duplicate keys)" - not sure if "attribute set of attribute set of" is correct or a repetition.


sandydoo:
@sixhobbits, this is great! The bad news is that this file is auto-generated. You'd have re-apply your changes to the modules in `src/modules`. Each option has a description, like so:
https://github.com/cachix/devenv/blob/755670d41ba697224a074c6ec97bb36cdaa668c6/src/modules/services/elasticsearch.nix#L88-L93

Some of the documentation is fetched from other repos. For example, pre-commit is taken from here: https://github.com/cachix/pre-commit-hooks.nix

domenkozar:
I've ported most of these manually in the main branch.

PR #338, opened 2023-01-30
Fix direnv with strict_env enabled

stephank:
I use nix-darwin + home-manager, and the home-manager zsh + direnv modules to configure my shell. I had to make these changes to get the devenv environment to load using direnv. I believe that's down to me having `strict_env` enabled globally in my home-manager config:

```nix
  programs.direnv = {
    enable = true;
    config.global.strict_env = true;
  };
```

Per direnv docs, this may become the default in the future: https://direnv.net/man/direnv.toml.1.html#codestrictenvcode

### DIFF
--- a/src/modules/mkNakedShell.nix
+++ b/src/modules/mkNakedShell.nix
@@ -96,10 +96,10 @@ let
       export PATH="$DEVENV_PROFILE/bin:$PATH"
 
       # prepend common compilation lookup paths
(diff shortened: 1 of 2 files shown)


domenkozar:
Could you rebase, since CI was failing?

stephank:
Done!

domenkozar:
We should probably test this on CI, any ideas?

PR #341, opened 2023-01-31
First class phps support

shyim:
Depends on #301 

Allows to use phps by fossar directly

```
languages.php.enable = true;
languages.php.version = "8.1";
languages.php.ini = ''
    memory_limit = 2G
'';
languages.php.extensions = [ "redis" "amqp" ];
```

When you use a PHP version which is not in nixpkgs, you get:

```
       error: To use languages.php.version, you need to add the following to your devenv.yaml:

       inputs:
         phps:
           url: github:fossar/nix-phps
           inputs:
             nixpkgs:
               follows: nixpkgs
```

after adding phps to inputs it's working then

### DIFF
--- a/src/modules/languages/php.nix
+++ b/src/modules/languages/php.nix
@@ -1,9 +1,30 @@
-{ pkgs, config, lib, ... }:
+{ pkgs, config, lib, inputs, ... }:
 
 with lib;
 
 let
+  inherit (lib.attrsets) attrValues genAttrs;
+
   cfg = config.languages.php;
+
+  setup = ''
+    inputs:
+      phps:
+        url: github:fossar/nix-phps
+        inputs:
+          nixpkgs:
+            follows: nixpkgs
+  '';
+
+  phps = inputs.phps or (throw "To use languages.php.version, you need to add the following to your devenv.yaml:\n\n${setup}");
+
+  configurePackage = package:
+    package.buildEnv {
+      extensions = { all, enabled }: with all; enabled ++ attrValues (getAttrs cfg.extensions package.extensions);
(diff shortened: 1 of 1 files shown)


domenkozar:
Nice! Would be good to update the example(s) with this new configurations

tcarrio:
Agreed. I use the nix-phps repo too so this is definitely a welcome addition for OOB PHP support 👏

src/modules/languages/php.nix
```diff
@@ -248,19 +291,25 @@ in
     };
   };
 
-  config = lib.mkIf cfg.enable {
-    packages = with pkgs; [
-      cfg.package
-    ] ++ lib.optional (cfg.packages.composer != null) cfg.packages.composer;
+  config =
+    let
+      customPhpPackage = (if ((builtins.hasAttr "php${version}" pkgs) && (builtins.tryEval (toString pkgs."php${version}")).success) then pkgs."php${version}" else phps.packages.${pkgs.system}."php${version}");
```
domenkozar, line 296: - the outer parenthesis isn't needed
- it would be good to split up the conditionals in multiple lines for readability
- if `phps.packages.${system}.php${version} isn't there, we should `throw` an error to the list of supported versions.

shyim: image
done :)

shyim:
updated also all examples :)

shyim:
I have to update the examples in a second PR when it's merged. As the lock needs to be updated 😅

shyim:
I cannot reproduce the CI failure with swift local. I have no glue why 😓. But I guess it's also not related to this xD

PR #343, opened 2023-02-01
language support: GNU Awk

0pointerexception:
WIP PR (still requires testing) to add GNU Awk language support including shared library extension provided by `gawkextlib`.

### DIFF
--- a/src/modules/languages/gawk.nix
+++ b/src/modules/languages/gawk.nix
@@ -1,0 +1,17 @@
+{ pkgs, config, lib, ... }:
+
+let
+  cfg = config.languages.gawk;
+in
+{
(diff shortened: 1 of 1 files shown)


domenkozar:
How can we test this simply?

0pointerexception:
> How can we test this simply?

A simple `awk --version` in a spawned devenv should be enough to test; the PR/awk language support should work fine as I use the same packages (gawk and gawkextlib) in a flake.nix with nix develop in some of my awk projects. The reason I wrote that it requires testing is, that I haven't migrated those projects to use devenv yet/had no possibility to test wether I can succesfuly use the awk language support I wrote for this PR in devenv this week (due to FOSDEM).

PR #347, opened 2023-02-02
Test direnv, and bash strict mode

stephank:
Conveniently, `direnv exec ./examples/simple true` already fails, because the lockfile points to an older devenv that doesn't have #338. So I was able to build a failing test first, then fix it.

Though, whether my workflow concoction works, we'll just have to see. 🤷‍♂️

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -64,3 +64,28 @@ jobs:
           nix profile remove '.*'
           nix profile install .
      - run: devenv shell devenv-test-example ${{ matrix.example }}
+  direnv:
+    name: direnv (${{ join(matrix.os) }})
+    strategy:
+      matrix:
+        os: [[ubuntu-latest], [macos-latest], [self-hosted, macOS]]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


stephank:
I'm not sure I can fix the failure, because the example lockfile references the github repo.

domenkozar:
@stephank which repo is the failure coming from?

stephank:
I'm testing direnv with `examples/simple`, but the `direnv.lock` points to `github.com/cachix/devenv`, so I can't apply the Linux fixes in this PR, because `main` won't have them

I guess I need to do those fixes in a separate PR, land them, then rebase this one if I want to fix it. But, I don't really have a Linux machine handy to test, so was relying on CI of this PR to fix issues.

domenkozar:
Let's merge and see if it passes

stephank:
I think we still need to do one last update of the direnv.lock in that example for it to pick up e1a8b96badd53b7a0c9a99257fc4eafcc787a39d, so I expect it to fail.

stephank:
It just dawned on me that this means we won't actually be testing (changes to) `main`, we'll only ever be testing whatever is pinned in that `devenv.lock`. That sounds... a little useless. At the same time, isn't that a problem for all the examples we're testing?

PR #358, opened 2023-02-05
Move example path overrides to CI

stephank:
Following on https://github.com/cachix/devenv/pull/347#issuecomment-1414286036, this makes the `simple` example again  test against the checkout, instead of what's in the lockfile. This benefits the direnv test, which can now pick up on any breakages.

I thought this also affected other examples, but noticed they all have a `devenv.yaml` path override. At the same time, there's [a todo](https://github.com/cachix/devenv/blob/871faa3fee0f4550074212b3db180e62fa699ab4/src/devenv.nix#L132-L133) to expand `devenv init` to other examples, so these would also be affected down the line.

This PR removes all those path overrides, and instead adds them in CI right before running the test. ~~This modification makes the lockfiles a little useless, so I've removed them. That also saves a few kb in releases, and helps us catch any breakages via nixpkgs or other inputs.~~

I also noticed [the comment in `examples/simple/devenv.yaml`](https://github.com/cachix/devenv/blob/871faa3fee0f4550074212b3db180e62fa699ab4/examples/simple/devenv.yaml#L4) was leaking in `devenv init`. This can be seen in the 0.5.1 release. With this change, we can get rid of it.

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -33,15 +33,15 @@ jobs:
         devenv shell devenv-run-tests
   generate-examples:
     runs-on: ubuntu-latest
-    outputs:                                                                                                            
-      examples: ${{ steps.set-examples.outputs.examples }}                                                                    
-    steps:                                                                                                              
-      - name: Checkout base repo                                                                                        
-        uses: actions/checkout@v3                                                                                      
-      - id: set-examples                                                                                                  
-        run: |                                                                                                          
-          json=$(tree -J -L 1 examples | jq -c '[.[0].contents[] | .name]')
-          echo "examples=$json" >> $GITHUB_OUTPUT                     
+    outputs:
+      examples: ${{ steps.set-examples.outputs.examples }}
+    steps:
+    - name: Checkout base repo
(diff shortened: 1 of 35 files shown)


stephank:
Hmm, removing the lockfiles apparently means we hit GitHub API rate limits.

I'm always mixed on if it's better to test against latest (nixpkgs et al) or if CI stability is preferred. Think I'm coming back on this, and stability is better.

I've kept the change moving path overrides to the CI workflow, and updated the lockfile `devenv` entries.

(https://github.com/cachix/install-nix-action/pull/157 probably helps with rate-limits too, but I think we should be good now.)

stephank:
Huh, CI fails at the `nix profile install .` step. That seems like a mishap?

domenkozar:
Seems like my conflict resolution broke something. Could you revert that and rebase?

stephank:
Definitely was my fault, simple typo. 😅 I discovered CI works decently well on the `main` branch of my own fork, so was able to test more there. Simply appending the input to `devenv.yaml` also doesn't work for `simple-remote`, because it has an imports section, so I changed it to to insert after `inputs:` via awk, and that works.

PR #375, opened 2023-02-07
cli: use column from util-linux package

bobvanderlinden:
Currently whenever a user doesn't have the `column` command, `devenv search` will fail.

For instance, I tried using devenv in `docker run -it ubuntu`. It resulted in the following output:

```
$ devenv search firefox
...
/home/me/.nix-profile/bin/devenv: line 100: column: command not found
```

This change should resolve that issue.

### DIFF
--- a/src/devenv.nix
+++ b/src/devenv.nix
@@ -104,14 +104,14 @@ pkgs.writeScriptBin "devenv" ''
       if [ "$results" = "{}" ]; then
         echo "No packages found for '$name'."
       else
-        ${pkgs.jq}/bin/jq -r '[to_entries[] | {name: ("pkgs." + (.key | split(".") | del(.[0, 1]) | join("."))) } * (.value | { version, description})] | (.[0] |keys_unsorted | @tsv) , (["----", "-------", "-----------"] | @tsv), (.[]  |map(.) |@tsv)' <<< "$results" | column -ts $'\t'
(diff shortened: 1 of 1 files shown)


src/devenv.nix
```diff
@@ -104,14 +104,14 @@ pkgs.writeScriptBin "devenv" ''
       if [ "$results" = "{}" ]; then
         echo "No packages found for '$name'."
       else
-        ${pkgs.jq}/bin/jq -r '[to_entries[] | {name: ("pkgs." + (.key | split(".") | del(.[0, 1]) | join("."))) } * (.value | { version, description})] | (.[0] |keys_unsorted | @tsv) , (["----", "-------", "-----------"] | @tsv), (.[]  |map(.) |@tsv)' <<< "$results" | column -ts $'\t'
+        ${pkgs.jq}/bin/jq -r '[to_entries[] | {name: ("pkgs." + (.key | split(".") | del(.[0, 1]) | join("."))) } * (.value | { version, description})] | (.[0] |keys_unsorted | @tsv) , (["----", "-------", "-----------"] | @tsv), (.[]  |map(.) |@tsv)' <<< "$results" | ${pkgs.util-linux}/bin/column -ts $'\t'
```
sandydoo, line 107: There's also `unixtools.column` which explicitly supports macOS. 

Amusingly, both versions of `column` eventually resolve to the same netbsd-flavoured derivation, with `util-linux` taking two symlinks to do so. 🤷

PR #376, opened 2023-02-07
add TeX Live support

BurNiinTRee:
This contains a new module for Tex Live under `languages.texlive`.
The module itself is taken almost verbatim from Home-Manager, hence the comment and license header.

### DIFF
--- a/examples/texlive/devenv.yaml
+++ b/examples/texlive/devenv.yaml
@@ -1,0 +1,3 @@
+inputs:
+  nixpkgs:
+    url: github:NixOS/nixpkgs/nixpkgs-unstable
--- a/src/modules/languages/texlive.nix
+++ b/src/modules/languages/texlive.nix
@@ -1,0 +1,26 @@
+{ pkgs, config, lib, ... }:
+let
+  cfg = config.languages.texlive;
+
+  base = cfg.base;
+  packages = lib.genAttrs cfg.packages (name: base.${name} or (throw "No such texlive package ${name}"));
+  package = base.combine packages;
+in
+{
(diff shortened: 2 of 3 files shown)


domenkozar:
Thanks for the contribution! A few notes:

- Just today I've merged a PR that removes lock files from examples, so that we can catch errors early on. Could you remove it here too?
- I'd rename `packageSet` to `base`, `extraPackages` would become `packages` and should be a list of strings. The current `package` isn't really needed so I'd remove it.
- Given that we've completely rewritten how texlive is integrated, the license disclaimer no longer applies.

BurNiinTRee:
done

src/modules/languages/texlive.nix
```diff
@@ -0,0 +1,26 @@
+{ pkgs, config, lib, ... }:
+let
+  cfg = config.languages.texlive;
+
+  base = cfg.base;
+  packages = lib.genAttrs cfg.packages (name: base.${name});
```
domenkozar, line 6:
```suggestion
  packages = lib.genAttrs cfg.packages (name: base.${name} or (throw "No such texlive package ${name}"));
```

examples/texlive/devenv.yaml
```diff
@@ -0,0 +1,5 @@
+inputs:
+  nixpkgs:
+    url: github:NixOS/nixpkgs/nixpkgs-unstable
+  devenv:
+    url: path:../../src/modules
```
domenkozar, line 5: `devenv` input is also gone from other examples as of today, sorry!

PR #377, opened 2023-02-08
php: cleanup with the new options

domenkozar:
cc @shyim


shyim (approved):
looks good :). But better than before

Issue #185, opened 2022-12-05
Validate devenv.yaml using strictyaml


domenkozar:
https://github.com/crdoconnor/strictyaml

domenkozar:
Ideally, but I haven't found a good tool that would produce reasonable error messages against yaml.

domenkozar:
We use json schema now :)

PR #383, opened 2023-02-09
allowUnfree and overlays

### LINKED ISSUE #154: Support unfree and overlays

domenkozar:
Fixes #154 #185 

You can now do `devenv.yaml`:

```yaml
allowUnfree: true
inputs:
  nixpkgs:
  rust-overlay:
    url: github:oxalica/rust-overlay
    overlays:
      - default
```

### DIFF
--- a/examples/overlays/devenv.yaml
+++ b/examples/overlays/devenv.yaml
@@ -1,0 +1,9 @@
+inputs:
+  nixpkgs:
+    url: github:NixOS/nixpkgs/nixpkgs-unstable
+  rust-overlay:
+    url: github:oxalica/rust-overlay
+allowUnfree: true
+overlays:
+  rust-overlay:
+   - default
--- a/examples/overlays/devenv.yaml
+++ b/examples/overlays/devenv.yaml
@@ -1,9 +1,8 @@
+allowUnfree: true
 inputs:
   nixpkgs:
     url: github:NixOS/nixpkgs/nixpkgs-unstable
   rust-overlay:
     url: github:oxalica/rust-overlay
-allowUnfree: true
-overlays:
-  rust-overlay:
-   - default
+    overlays:
+      - default
--- a/src/devenv-yaml.nix
+++ b/src/devenv-yaml.nix
@@ -1,0 +1,43 @@
+{ pkgs }:
+
+pkgs.writers.writePython3Bin "devenv-yaml" { libraries = with pkgs.python3Packages; [ strictyaml path ]; } ''
+  from strictyaml import Map, MapPattern, Str, Seq
+  from strictyaml import load, Bool, Any, Optional, YAMLError
+  import json
+  import sys
+  import os
+  from path import Path
+
+  inputsSchema = MapPattern(Str(), Map({
+      "url": Str(),
+      Optional("flake", default=None): Bool(),
+      Optional("inputs", default=None): Any(),
+      Optional("overlays", default=None): Seq(Str())
+  }))
+
+  schema = Map({
+      Optional("inputs", default=None): inputsSchema,
+      Optional("allowUnfree", default=False): Bool(),
+      Optional("imports", default=None): Seq(Str())
+  })
+
+  filename = Path("devenv.yaml").bytes().decode('utf8')
+  try:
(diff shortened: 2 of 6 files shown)


examples/overlays/devenv.yaml
```diff
@@ -0,0 +1,9 @@
+inputs:
+  nixpkgs:
+    url: github:NixOS/nixpkgs/nixpkgs-unstable
+  rust-overlay:
+    url: github:oxalica/rust-overlay
+allowUnfree: true
+overlays:
+  rust-overlay:
+   - default
```
bobvanderlinden, line 9: I don't want to uphold changes like this, but has `overlays` as part of the individual `inputs` also been considered? It seems less work and less error-prone to mention `rust-overlay` once.

```suggestion
inputs:
  nixpkgs:
    url: github:NixOS/nixpkgs/nixpkgs-unstable
  rust-overlay:
    url: github:oxalica/rust-overlay
    overlays:
      - default
allowUnfree: true
```

domenkozar: I thought of that, but what if Nix flakes suddenly also get this parameter? I'm afraid it could break devenv.

Maybe it's worth the risk and hope that it's done with the same API :)

domenkozar: Maybe

```
inputs:
  rust-overlay:
    devenv:
       overlays:
         - default
```

bobvanderlinden: Or just not presume `inputs` is going to be passed through to Nix flakes. We have no idea what flakes are going to do before stabilization :sweat_smile:

If we explicitly support `inputs.<input>.url` and `inputs.<input>.follows`, then it is fine if thinks of more features, devenv first needs to be updated before those features can be used in devenv. It gives a bit more control over the `devenv.yaml` structure.

domenkozar: Implemented and also #185

shyim:
Will this need that the user update devenv on their machine or is it enough that youre modules update

domenkozar:
You'll need to update devenv :)

kubukoz:
Works great for me!

domenkozar:
I was thinking about the UX. Now that we use strictyaml for validation we can do this and nudge people to upgrade.

I'm also thinking to add a field like `minimalVersion: XXX` to `devenv.yaml` that would kind of tell users straight away which version the project needs.

domenkozar:
Moved the scope for the UX into https://github.com/cachix/devenv/issues/388

kubukoz (approved):

kubukoz:
:shipit:

src/devenv-yaml.nix
```diff
@@ -0,0 +1,46 @@
...
+
+  devenv_state = sys.argv[1]
+
+  with open(os.path.join(devenv_state, "flake.json"), 'w') as f:
+      f.write(json.dumps(inputs))
+
+  with open(os.path.join(devenv_state, "devenv.json"), 'w') as f:
+      f.write(json.dumps(devenv))
+
+  with open(os.path.join(devenv_state, "imports.txt"), 'w') as f:
```
hugosenari, line 44: **🎯 suggestion:** Keep the same var name of bash to reduce cognitive load

```suggestion
  devenv_dir = sys.argv[1]

  with open(os.path.join(devenv_dir, "flake.json"), 'w') as f:
      f.write(json.dumps(inputs))

  with open(os.path.join(devenv_dir, "devenv.json"), 'w') as f:
      f.write(json.dumps(devenv))

  with open(os.path.join(devenv_dir, "imports.txt"), 'w') as f:
```

src/devenv-yaml.nix
```diff
@@ -0,0 +1,46 @@
...
+      f.write(" ".join(devenv.get('imports', [])))
```
hugosenari, line 45: As requires (?) for [387](https://github.com/cachix/devenv/pull/387#pullrequestreview-1294050577)
```suggestion
      f.write("\n".join(devenv.get('imports', [])))
```

PR #387, opened 2023-02-10
Caching direnv integration

### LINKED ISSUE #382: Direnv Flake Caching
**Describe the bug**
For a project at work, I currently use a flake-defined devShell environment, which reuses a cached evaluation when opening the environment. 

I'm evaluating devenv as a possible replacement that could bring the benefits of nix to the rest of my dev team with a reduced learning curve. However, devenv appears to lose the flake cache behavior, and instead rebuilds the shell environment even when the watch_file entries have not changed. I can avoid this by using devenv with the flake integration directly and `use flake` in .envrc, but this adds some complexity for users who aren't familiar with nix.

Is there a known strategy to use the default configuration from `devenv init` while still benefiting from flake environment caching?

**To Reproduce**
```
devenv init
DIR=$(pwd)
cd ../
cd $DIR
# shell rebuilds even if devenv.nix didn't change
```

**Version**
 
`0.5.1`

domenkozar:
- Fixes #382 #133 #237 

It differs to `nix-direnv` by being devenv specific:

- it tracks `imports` section of `devenv.yaml`
- it removes non-flake code
- it removes GC roots that are handled by devenv

Needs testing.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,7 +1,6 @@
-watch_file devenv.nix
-watch_file devenv.yaml
-watch_file devenv.lock
-watch_file result/bin/devenv
-watch_dir src
-nix-build
-eval "$(./result/bin/devenv print-dev-env)"
+source_url "https://raw.githubusercontent.com/cachix/devenv/af67fa52f0d0588b185069354a4c94aba0e32476/direnvrc" "sha256-sLuqWQl6hJ8eO1xaDwUETRnobEV+LL/9rhje9calUz8="
+
+# use our own last built devenv/nix in CLI
+PATH_add result/bin
+
+use devenv
--- a/direnvrc
+++ b/direnvrc
@@ -114,14 +114,11 @@ use_devenv() {
         log_status "$(devenv version) failed to parse devenv.yaml, make sure to use version 0.6 or newer and fix the errors above."
         exit 1
       fi
-      if has python; then
-        files=$(python -c "import json;print(' '.join(json.load(open('.devenv/devenv.json')).get('imports', [])))")
-        for file in $files; do
-          files_to_watch+=("$file")
+
+      if [[ -f "$flake_dir/.devenv/imports.txt" ]]; then
+        for file in $(cat "$flake_dir/.devenv/imports.txt"); do
+            files_to_watch+=("$file")
         done
-      # TODO: support jq
-      else
-        log_status "install python to watch all devenv.yaml imports"
       fi
     fi
   fi
--- a/direnvrc
+++ b/direnvrc
@@ -1,5 +1,5 @@
 # shellcheck shell=bash
-# adopted from https://github.com/nix-community/nix-direnv/blob/master/direnvrc
+# adapted from https://github.com/nix-community/nix-direnv/blob/master/direnvrc
 
 REQUIRED_DIRENV_VERSION="2.21.3"
 
--- a/direnvrc
+++ b/direnvrc
@@ -1,0 +1,157 @@
+# shellcheck shell=bash
+# adopted from https://github.com/nix-community/nix-direnv/blob/master/direnvrc
+
+REQUIRED_DIRENV_VERSION="2.21.3"
+
+_nix_direnv_preflight () {
+  if [[ -z "$direnv" ]]; then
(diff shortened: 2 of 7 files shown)


.envrc
```diff
@@ -1,7 +1,8 @@
-watch_file devenv.nix
-watch_file devenv.yaml
-watch_file devenv.lock
-watch_file result/bin/devenv
-watch_dir src
-nix-build
-eval "$(./result/bin/devenv print-dev-env)"
+if ! has nix_direnv_watch_file; then
```
shyim, line 1: Due to this if condition my direnv does not download it. You check for nix_direnv and not devenv stuff

thenonameguy: Yes, this should be different, or at least add a version check similar to my previous PR.
```
if ! has nix_direnv_version || ! nix_direnv_version 2.2.0-devenv; then
  ...
fi

File: examples/simple/.envrc [hunk 2]
```diff
@@ -1,4 +1,5 @@
-watch_file devenv.nix 
-watch_file devenv.yaml 
-watch_file devenv.lock
-eval "$(devenv print-dev-env)"
\ No newline at end of file
+if ! has nix_direnv_watch_file; then
```
thenonameguy, line 1: Same as [above.](https://github.com/cachix/devenv/pull/387/files#r1102398905)

.envrc
domenkozar: I've pushed a version that just always applies the latest direnv integration, don't see a reason not to do this :)

domenkozar:
I consider this ready :rocket:

examples/simple/.envrc
```diff
@@ -1,4 +1 @@
-watch_file devenv.nix 
-watch_file devenv.yaml 
-watch_file devenv.lock
-eval "$(devenv print-dev-env)"
\ No newline at end of file
+source_url "https://raw.githubusercontent.com/cachix/devenv/af67fa52f0d0588b185069354a4c94aba0e32476/direnvrc" "sha256-sLuqWQl6hJ8eO1xaDwUETRnobEV+LL/9rhje9calUz8="
```
hugosenari, line 1: We don't need `use devenv` like in other files?

direnvrc
```diff
@@ -0,0 +1,157 @@
+# shellcheck shell=bash
+# adopted from https://github.com/nix-community/nix-direnv/blob/master/direnvrc
```
hugosenari, line 2: Typo ?

direnvrc
```diff
@@ -0,0 +1,157 @@
...
+
+  if [[ -d "$flake_dir" ]]; then
+    files_to_watch+=("$flake_dir/devenv.nix" "$flake_dir/devenv.lock" "$flake_dir/devenv.yaml" "$flake_dir/devenv.local.nix")
+    if [[ -f "$flake_dir/devenv.yaml" ]]; then
+      if ! devenv assemble; then
+        log_status "$(devenv version) failed to parse devenv.yaml, make sure to use version 0.6 or newer and fix the errors above."
+        exit 1
+      fi
+      if has python; then
+        files=$(python -c "import json;print(' '.join(json.load(open('.devenv/devenv.json')).get('imports', [])))")
```
hugosenari, line 118:
```suggestion
        files=$(python -c "import json;print(' '.join(json.load(open('$flake_dir/.devenv/devenv.json')).get('imports', [])))")
```

hugosenari: Why not 'assemble' .devenv/imports.txt ?

direnvrc
domenkozar: Why do you think it's a typo?

direnvrc
domenkozar: Ohh good one! Done.

direnvrc
hugosenari:
```suggestion
# adapted from https://github.com/nix-community/nix-direnv/blob/master/direnvrc
```

direnvrc
hugosenari: But now `devenv assemble` should create imports.txt :-D

domenkozar: It's part of https://github.com/cachix/devenv/pull/383

hugosenari (approved):
Depends on #383

PR #391, opened 2023-02-11
Init Cassandra service

ankhers:
This is not yet fully working.

When using the default Cassandra package you will get warnings about python being an insecure package. There is also an issue where I seem unable to connect to the database using `cqlsh`.

When using the `cassandra_4` package, I am able to connect using the `cqlsh` tool.

I'm sure the code itself can be cleaned up as well.

### DIFF
--- a/src/modules/services/cassandra.nix
+++ b/src/modules/services/cassandra.nix
@@ -56,8 +56,8 @@ in
     package = lib.mkOption {
       type = types.package;
       description = "Which version of Cassandra to use";
-      default = pkgs.cassandra;
-      defaultText = lib.literalExpression "pkgs.cassandra";
+      default = pkgs.cassandra_4;
+      defaultText = lib.literalExpression "pkgs.cassandra_4";
       example = lib.literalExpression "pkgs.cassandra_4;";
     };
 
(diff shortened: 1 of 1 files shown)


src/modules/services/cassandra.nix
```diff
@@ -0,0 +1,103 @@
...
+      description = ''
+        Extra options to be merged into `cassandra.yaml` as nix attribute set.
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    packages = [ cfg.package startScript ];
+
+    env.JVM_OPTS = (builtins.getEnv "JVM_OPTS") +
```
shyim, line 96: I guess JVM_OPTS is a standard environment variable for Java apps. It would make sense to move this into `startScript` to it does not collidate with other java apps.

Then the `JVM_OPTS` could be also an option :)

domenkozar:
Looks good to me :)

domenkozar:
CI failure was fixed in `main` branch

ankhers:
Just rebased from the `main` branch. 

One thing to note is that I am still unable to seem to connect using the default `cassandra` package. But I think I have found out why. Avoid merging this for now and I will see if I can figure it out.

domenkozar:
Looking at https://github.com/NixOS/nixpkgs/pull/213140, we are probably best off using Cassandra 4 that supports Python 3.

PR #397, opened 2023-02-13
Overmind root

domenkozar:


src/modules/languages/raku.nix
```diff
@@ -5,7 +5,7 @@ let
 in
 {
   options.languages.raku = {
-    enable = lib.mkEnableOption "Enable tools for Raku development.";
+    enable = lib.mkEnableOption "tools for Raku development";
```
bobvanderlinden: Unrelated change, but a good one nonetheless 👍

bobvanderlinden (approved):
Documentation confirms this is how--root works:

> Specify a working directory of application. Default: directory containing the Procfile

PR #401, opened 2023-02-13
ruby: add nixpkgs-ruby integration

### LINKED ISSUE #220: Integrate nixpkgs-ruby as a first-class support
In similar style than Rust, let's support https://github.com/domenkozar/devenv-ruby/blob/main/devenv.nix out of the box.

https://github.com/DevelopMan/nix-rails-test-app/blob/master/devenv.nix

bobvanderlinden:
Currently the ruby package can be manually changed by the user, but it is often inconvenient to change it to a specific ruby version.
[nixpkgs-ruby](https://github.com/bobvanderlinden/nixpkgs-ruby) holds package definitions (and caches) all ruby versions.
Just like [`phps` integration in devenv](https://github.com/cachix/devenv/pull/341), this makes use of nixpkgs-ruby whenever the version option is set.

In addition, since the ruby community is based mostly around [rbenv](https://github.com/rbenv/rbenv), support for setting the version using a `versionFile` (`.ruby-version`) is
also added. This makes use of nixpkgs-ruby's `packageFromRubyVersionFile`
helper function.

When when 2 or more of the version, versionFile or package properties are set, a conflict will show up, because multiple sources are trying to set the package property.

Note that I explicitly have not chosen to use `follows: { nixpkgs: nixpkgs }` for `nixpkgs-ruby`. This is because this will result in cache-misses. It's a bit awkward, since it is choosing between 2 evils. When _not_ following nixpkgs, I think this _could_ cause trouble when for instance glibc is updated in nixpkgs, while it isn't updated in the version of nixpkgs-ruby. On the other hand, when we're following nixpkgs we won't be able to make use of caches of nixpkgs-ruby.

Fixes #220


domenkozar:
This looks great! I wonder if we should set `cfg.versionFile` to default to `${self}/.ruby-version` and that should be used as a preference over `cfg.version`.

domenkozar:
No reason to block this PR though :)

bobvanderlinden:
> This looks great! I wonder if we should set `cfg.versionFile` to default to `${self}/.ruby-version` and that should be used as a preference over `cfg.version`.

I'm not sure. Defaulting this would make it less clear what is going on. On the other hand, if the file exists and defaulting the version automagically _is_ nice from a UX perspective. `devenv init` inside your Ruby project and you're all set up...

Apart from that, I'd first like to have `assertions` in denenv so that I can give a nicer error message when both `version` and `versionFile` are set.

PR #405, opened 2023-02-13
add ability to define assertions

bobvanderlinden:
Currently it is not possible to conveniently define erroneous state in modules. This happens when for instance options are incompatible with each-other and a clear error with instructions should be shown.

This adopts the `assertions` options that is used in NixOS modules as well.

I was thinking about how to properly test this, but the examples weren't fitting to test (on purpose) evaluation errors. I have manually done the following:

With `devenv.nix`:
```nix
{ pkgs, ... }:

{
  assertions = [{ assertion = false; message = "Oh noes"; }];
}
```

It resulted in the following output:

```
       error: Failed assertions:
       - Oh noes
```

Note that it did _also_ output the whole stack trace, which isn't great for UX atm. That said, I thought it's better than not having error message.

The usecase for this feature is to show some instructions when `language.ruby.version` as well as `languages.ruby.versionFile` are both defined. Currently it shows an error like:

```
       … while calling 'mergeUniqueOption'

       error: The option `languages.ruby.package' is defined multiple times while it's expected to be unique.
```

### DIFF
--- a/src/modules/top-level.nix
+++ b/src/modules/top-level.nix
@@ -16,6 +16,17 @@ let
     paths = lib.flatten (builtins.map drvOrPackageToPaths config.packages);
     ignoreCollisions = true;
   };
+
+  failedAssertions = builtins.map (x: x.message) (builtins.filter (x: !x.assertion) config.assertions);
+
+  performAssertions =
+    if failedAssertions != [ ]
+    then
+      throw ''
+        Failed assertions:
+        ${lib.concatStringsSep "\n" (builtins.map (x: "- ${x}") failedAssertions)}
+      ''
+    else lib.id;
 in
 {
   options = {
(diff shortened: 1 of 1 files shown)


bobvanderlinden:
> Having the full stack trace can be useful when you don't expect the assertion to happen.

Certainly. For this usecase though it is an expected 'error', but the stack trace points to the place where the assertion error was thrown and not where the violating pieces of code where the assertion potentially refers to.
No matter though, could be something to iterate on later on, but for now it's already useful to be able to define assertions, regardless of potential future UX improvements. :+1:

> I'm not sure where the infinite recursion is coming from, maybe you can't use `lib.id` on config?

Found out I didn't push the latest changes. It's around `mkNakedShell` instead of `config` now.

PR #406, opened 2023-02-13
add mkcert as integration

jochenmanz:
Ive added mkcert as a basic integration and refaktored the hostctl integration, so both "feel" the same when using them. Feedback is appreciated :)

### DIFF
--- a/examples/mkcert/devenv.nix
+++ b/examples/mkcert/devenv.nix
@@ -1,11 +1,11 @@
 { pkgs, config, ... }:
 
 {
-  mkcert.domains = [
-    "example.com"
+  certificates = [
+    "*.2example.com"
   ];
 
-  hostctl.hosts = {
+  hosts = {
     "example.com" = "127.0.0.1";
   };
 
--- a/examples/mkcert/devenv.nix
+++ b/examples/mkcert/devenv.nix
@@ -1,0 +1,9 @@
+{ pkgs, config, ... }:
+
+{
+  mkcert.domains = [
+    "foo.de"
+  ];
+
+  services.caddy.enable = true;
+}
--- a/examples/mkcert/devenv.nix
+++ b/examples/mkcert/devenv.nix
@@ -2,7 +2,7 @@
 
 {
   certificates = [
-    "*.2example.com"
+    "example.com"
   ];
 
   hosts = {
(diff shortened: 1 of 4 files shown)


domenkozar:
I'd like to have the options not tied to the implementation. This allows to switch to a better/different implement in the future.

That's why we decided to name `hosts` and not `hostctl`. 

I'm thinking how to name this option, `hostCertificates`?

examples/mkcert/devenv.nix
```diff
@@ -0,0 +1,20 @@
+{ pkgs, config, ... }:
+
+{
+  mkcert.domains = [
+    "example.com"
```
shyim, line 5: This interface is like easy, but not flexible. What if I want to have one certificate with `*.example.com` and `example.com`. 

Also the referencing directly into the state feels not like a good ux 🙈. Maybe a helper to assemble the path 🤔

jochenmanz: Do you have an example on how to add such a helper for the path? 🤔

jochenmanz:
> I'd like to have the options not tied to the implementation. This allows to switch to a better/different implement in the future.
> 
> That's why we decided to name `hosts` and not `hostctl`.
> 
> I'm thinking how to name this option, `hostCertificates`?

Ive renamed the option to certificates, that can be easily switched later on.

domenkozar:
How does it work when you generate `*.example.com`? Does it create a file named `*.example.com`?

src/modules/integrations/mkcert.nix
```diff
@@ -0,0 +1,42 @@
...
+in
+{
+  options = {
+    certificates = lib.mkOption {
+      type = lib.types.listOf lib.types.str;
+      default = [ ];
+      description = "List of domains to generate certificates for.";
+      example = [
+        "example.com"
+        "127.0.0.1"
```
domenkozar, line 15:
```suggestion
        "*.example.com"
```

jochenmanz:
> How does it work when you generate `*.example.com`? Does it create a file named `*.example.com`?

Wildcards are generated as: _wildcard.example.com-key.pem and _wildcard.example.com.pem

schliesser:
> > How does it work when you generate `*.example.com`? Does it create a file named `*.example.com`?
> 
> Wildcards are generated as: _wildcard.example.com-key.pem and _wildcard.example.com.pem

You can define the filename for the certificate manually if you want:

`mkcert -cert-file certfile.crt -key-file keyfile.key "example.com"`

PR #416, opened 2023-02-18
fix: avoid eval error by discouraging overzealus module signatures

blaggacao:


blaggacao:
Stuff like this:
```console
error: attribute 'pre-commit-hooks' missing

       at /nix/store/h3vlwnvhbkjlck4ps694kmz20jwvgscc-source/lib/modules.nix:483:28:

          482|         builtins.addErrorContext (context name)
          483|           (args.${name} or config._module.args.${name})
             |                            ^
          484|       ) (lib.functionArgs f);
/nix/store/1c147j530hi4p2v5dlpg5x9dm309pivi-source/direnv_lib.s
```

The following consumption pattern is fixed by this change:
```nix
{
  inputs.devenv.url = "github:cachix/devenv?dir=src/modules";
}
```

blaggacao:
boah, I guess we need an integration point for `services`....

```
error: The option `pre-commit' does not exist. Definition values:
       - In `/nix/store/xyp27kgag9yp8ldix8rsg38wpr0xv1ff-source/src/modules/languages/rust.nix':
           {
             _type = "if";
             condition = false;
             content = {
               tools = {
           ...
```

blaggacao:
Uncomfortable, but works:
```nix
{
  eval = module: (lib.evalModules {
    modules = [
        (devenv.modules + /top-level.nix)
        {
          disabledModules = []
            ++ (listEntries (devenv.modules + /integrations))
            ++ (listEntries (devenv.modules + /languages))
          ;
        }
        module
    ];
    specialArgs = {
      pkgs = nixpkgs;
      inherit inputs;
    };
  }).config;
```

Maybe this should become an issue?

src/modules/integrations/pre-commit.nix
```diff
@@ -1,5 +1,14 @@
-{ pkgs, self, lib, pre-commit-hooks, config, ... }:
+{ pkgs, self, lib, inputs, config, ... }:
+let
 
+  pre-commit-hooks = inputs.pre-commit-hooks
+    or (throw ''
+    To use integrations.pre-commit, you need to add the following to your flake inputs:
+
+      inputs.pre-commit-hooks.url = "github:cachix/pre-commit-hooks.nix";
+  '');
```
bobvanderlinden, line 9: Might be good to refer to devenv.yaml, instead of flakes, as that is the default way people are using devenv. The same is done in the Ruby module:

https://github.com/cachix/devenv/blob/ecd528f2dd648091469577a5a30512f1bc7e974b/src/modules/languages/ruby.nix#L6-L12

bobvanderlinden (commented):
I like that this makes pre-commit-hooks less special :+1: Makes sense to pass is as an input in specialArgs.

domenkozar:
The issue with this change is that it will break for all users that won't use `devenv` from master, as soon as they upgrade.

We need to provide backwards compatible way to transition (it should just work for devenv 0.5.1 or lower).

blaggacao:
Please take over!

domenkozar:
For anyone that picks this up:

- instructions show should devenv.yaml snippet (there could be a function that chooses based on `config.devenv.flakesIntegration` which format to pick)
- make sure it doesn't break backwards compatibility

domenkozar:
Closing as it's completely out of date, if someone wants to pick it up, please do.

PR #417, opened 2023-02-18
fix: process-compose invokation (use values from config as defaults)

blaggacao:


blaggacao:
Avoid:
```console
❯ devenv up
Error: invalid argument "--tui=" for "-p, --port" flag: strconv.ParseInt: parsing "--tui=": invalid syntax
```

PR #419, opened 2023-02-19
python: rebuild venv when devenv-profile is updated 

### LINKED ISSUE #336: venv becomes stale after some time (nix store gc?)
**Describe the bug**
venv breaks down after not touching it for a couple days. I have `nix.gc.automatic = true` enabled.

```
$ pip install polars
exec: Failed to execute process '/Users/farcaller/src/example/.devenv/state/venv/bin/pip': The file specified the interpreter '/Users/farcaller/src/example/.devenv/state/venv/bin/python', which is not an executable command.

$ ls -l /Users/farcaller/src/example/.devenv/state/venv/bin/pip
.rwxr-xr-x@ 267 farcaller 25 Jan 13:22 /Users/farcaller/src/example/.devenv/state/venv/bin/pip

$ ls -l /Users/farcaller/src/example/.devenv/state/venv/bin/python
lrwxr-xr-x@ 69 farcaller 25 Jan 13:22 /Users/farcaller/src/example/.devenv/state/venv/bin/python -> /nix/store/dchvlb6c2qn78b4623xsfrm7kqa2487g-devenv-profile/bin/python

$ stat /nix/store/dchvlb6c2qn78b4623xsfrm7kqa2487g-devenv-profile/bin/python
stat: /nix/store/dchvlb6c2qn78b4623xsfrm7kqa2487g-devenv-profile/bin/python: stat: No such file or directory
```

**To Reproduce**
https://gist.github.com/farcaller/57670b22ecb6e91f82a41a95a0637f1d

**Version**
 
devenv: 0.5

bobvanderlinden:
Resolves #336

This PR includes 3 changes

22adea33daf57df14debc6cf6bc55b9db1ddfad5 is to resolve #336. It recreates venv each time devenv-profile is updated.

While working on this change I ran into syntax errors in my enterShell script. This resulted in errors like:

```
$ devenv shell
Building shell ...
Entering shell ...

bash: eval: line 231: syntax error near unexpected token `devenv'
bash: eval: line 231: `export PS1="\[\e[0;34m\](devenv)\[\e[0m\] ${PS1-}"'
```

It was hard to pin down where the problem actually lays, so I decided to move the venv and poetry scripts from enterShell to separate .sh files. This made the error messages clearer and the enterShell code more readable. See 057b1c019ae27a4cfd5eef37b88a7b20b8ba8bef

Lastly I noticed that there could be problems with running poetry when venv is rebuild. I resolved these in 9590d815a059dfc1dc1a483ddcb21352e472e854.

See the commit messages of the individual commits for further explanation.

### DIFF
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -43,12 +43,11 @@ let
 
       if [ "$ACTUAL_POETRY_CHECKSUM" != "$EXPECTED_POETRY_CHECKSUM" ]
       then
-        poetry install --no-interaction --quiet
+        ${cfg.poetry.package}/bin/poetry install --no-interaction --quiet
         echo "$ACTUAL_POETRY_CHECKSUM" > "$POETRY_CHECKSUM_FILE"
       fi
     fi
   '';
-
 in
 {
   options.languages.python = {
@@ -67,8 +66,14 @@ in
       enable = lib.mkEnableOption "poetry";
       package = lib.mkOption {
         type = lib.types.package;
-        default = cfg.package.pkgs.poetry-core;
-        defaultText = lib.literalExpression "config.languages.python.package.pkgs.poetry";
+        default = pkgs.poetry.override {
+          python3 = cfg.package;
+        };
+        defaultText = lib.literalExpression ''
+          pkgs.poetry.override {
+            python3 = config.languages.python.package;
+          }
+        '';
         description = "The Poetry package to use.";
       };
     };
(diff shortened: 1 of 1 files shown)


bobvanderlinden:
@farcaller Is this a viable solution for your use-case? Rebuilding venv should avoid the problems you ran into. It will take a bit longer to activate devenv upon making changes to devenv.nix. In addition, it will remove all ad-hoc python packages that were installed inside devenv using for instance `pip install`.

farcaller:
Sounds like a reasonable trade-off.

domenkozar:
We should probably expose a hook to run after `venv` has been created, so that you can do `pip install -r requirements.txt`.

PR #420, opened 2023-02-19
feat: add couchdb as a new service

MSBarbieri:


src/modules/services/couchdb.nix
```diff
@@ -0,0 +1,151 @@
...
+      touch ${baseDir}/.erlang.cookie
+      chmod 600 ${baseDir}/.erlang.cookie
+      dd if=/dev/random bs=16 count=1 | base64 > ${baseDir}/.erlang.cookie
+    fi
+
+    exec ${cfg.package}/bin/couchdb
+  '';
+in
+{
+  imports = [
```
domenkozar, line 45: This import is not needed because this option is new, so it doesn't need renaming.

src/modules/services/couchdb.nix
```diff
@@ -0,0 +1,151 @@
...
+    (lib.mkRenamedOptionModule [ "couchdb" "enable" ] [
+      "services"
+      "couchdb"
+      "enable"
+    ])
+  ];
+
+  options.services.couchdb = {
+    enable = lib.mkEnableOption "Add CouchDB process script.";
```
domenkozar, line 54:
```suggestion
    enable = lib.mkEnableOption "CouchDB process";
```

domenkozar (changes requested):

src/modules/services/couchdb.nix
```diff
@@ -0,0 +1,143 @@
...
+      [admins]
+      ${cfg.adminUser} = ${cfg.adminPass}
+
+      [chttpd]
+      port = ${toString cfg.port}
+      bind_address = ${cfg.bindAddress}
+      [log]
+      file = ${cfg.logFile}
+    ''
+  );
```
shyim, line 27: What about we adopt this here https://github.com/NixOS/rfcs/blob/master/rfcs/0042-config-option.md

Basically we do:

```
settingsFormat = pkgs.formats.ini {};
configFile = settingsFormat.generate "couchdb.ini" cfg.settings;
```

...

```
settings = lib.mkOption {
      type = lib.types.submodule {
        freeformType = settingsFormat.type;

        options.couchdb.database_dir = lib.mkOption {
          type = lib.types.str;
          default = baseDir;
          description = lib.mdDoc ''
            Base Database dir
          '';
        };

        options.couchdb.uri_file = lib.mkOption {
          type = types.path;
          default = "${baseDir}/couchdb.uri";
          description = lib.mdDoc ''
            This file contains the full URI that can be used to access this
            instance of CouchDB. It is used to help discover the port CouchDB is
            running on (if it was set to 0 (e.g. automatically assigned any free
            one). This file should be writable and readable for the user that
            runs the CouchDB service (couchdb by default).
          '';
        };

        options.couchdb.view_index_dir = lib.mkOption {
          type = types.path;
          default = baseDir;
          description = lib.mdDoc ''
            Specifies location of CouchDB view index files. This location should
            be writable and readable for the user that runs the CouchDB service
            (couchdb by default).
          '';
        };

        options.couchdb.single_node = lib.mkOption {
          type = types.bool;
          default = true;
          description = lib.mdDoc ''
            Single node instance
          '';
        };
      };

      default = {};

      description = lib.mdDoc ''
        CouchDB configuration.
      '';
    };
```

then the user is able to configure anything inside the ini file 🤔 . also the options are specified in same position as the real config

src/modules/services/couchdb.nix
```diff
@@ -0,0 +1,89 @@
...
+            port = 5984;
+            logFile = "${baseDir}/couchdb.log";
+          };
+        }
+      '';
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    packages = [ cfg.package startScript ];
```
shyim, line 85: You should not add the startScript as package :thinking: or is there a specific reason. The process works also without adding it

src/modules/services/couchdb.nix
```diff
@@ -0,0 +1,156 @@
...
+        options.chttpd.logFile = lib.mkOption {
+          type = lib.types.path;
+          default = "${baseDir}/couchdb.log";
+          description = lib.mdDoc ''
+            Specifies the location of file for logging output.
+          '';
+        };
+      };
+    };
+    description = ''
```
domenkozar, line 108: This should be inside the `settings` option probably and not use xml but markdown for formatting. Then we're done :)

shyim (approved):

PR #418, opened 2023-02-19
fix process compose 2

blaggacao:
- fix: process-compose invokation (use values from config as defaults)
- fix: env format for process-compose

PR #424, opened 2023-02-22
fix(process-compose): do not escape strings for process-compose env vars

thenonameguy:
fixes breakage caused by https://github.com/cachix/devenv/commit/8bc83de411910ec5ed9759f8478c49639326e5f0

Bug caused `"/` folders to be created in `$DEVENV_ROOT`.


thenonameguy:
closes #418

thenonameguy:
Verified in private project that this fixes the regression, tested via the Postgres service.

PR #429, opened 2023-02-23
Using `[` instead of `[[` with (maybe) empty bash vars can cause errors.

mexisme:
e.g.
When I tried to run `devenv build`, this caused this error:
```
/nix/store/[...]-devenv-profile/bin/devenv: line 118: [: -eq: unary operator expected
```

... even though the code suggested it should default to running as `devenv build shell`

### DIFF
--- a/src/devenv.nix
+++ b/src/devenv.nix
@@ -122,7 +122,7 @@ pkgs.writeScriptBin "devenv" ''
 
       paths=$($CUSTOM_NIX/bin/nix $NIX_FLAGS build --impure --no-link --print-out-paths "''${builds[@]}")
  
-      if [ $exec -eq 1 ]; then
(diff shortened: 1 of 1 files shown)


mexisme:
The error seems to have come back?
Did this PR's commit get dropped/over-written in the upstream `cachix/devenv`#`build-container`, @domenkozar ?
- https://github.com/cachix/devenv/blob/build-container/src/devenv.nix#L125

domenkozar:
Seems like I fixed the wrong one before force-pushing. Should be fixed now!

mexisme:
> Seems like I fixed the wrong one before force-pushing. Should be fixed now!

No worries!
I wasn't sure if you had discovered a better way to fix it, anyway, TBH.

PR #448, opened 2023-03-02
Fix indentation in the devenv script

madjar:
Part of the script was less indented than the root, which makes most of the final script indented by 2 spaces.

For some reason, this makes `fish` run ignore the shebang and run it with `sh`, which doesn't work.


madjar:
Note: after running `nixpkgs-fmt`, the change is not about indenting the lines that were not indented enough (https://github.com/cachix/devenv/blob/main/src/devenv.nix#L107-L114, for the record), but de-indenting the rest. I'm very sorry about the diff.

Issue #453, opened 2023-03-02
Issue with postgres for process-compose and devenv 0.6


manuelhc621:
**Describe the bug**
After upgrade to 0.6. tried the porcess-compose implementation, but the postgres service failed with the below error:
```
Failed to parse /nix/store/wasd2h7ir2bjp1vqga2w2lgrpcqc6h1z-process-compose.yaml - yaml: unmarshal errors:
  line 16: cannot unmarshal !!map into string
  line 31: cannot unmarshal !!map into int
```

**To Reproduce**

Use this [deven.nix file](https://gist.github.com/manuelhc621/63ea4b3faa314dfbfaef1b2baf52bdf3)

** Notes **
The generated yaml looks like this (at least the offending lines):
```
 12   │ processes:
  13   │   postgres:
  14   │     availability:
  15   │       restart:
  16   │         _type: override
  17   │         content: on_failure
  18   │         priority: 1000
  19   │     command: /nix/store/a1x2cdkaq80hcvfypsn1n02x1vxz6hr9-start-postgres/bin/start-postgres
  20   │     readiness_probe:
  21   │       exec:
  22   │         command: /nix/store/bb052x7g4ybm0yzsdh57583hhc005dgg-postgresql-15.2/bin/pg_isready
  23   │           -h $PGDATA -d template1
  24   │       failure_threshold: 5
  25   │       initial_delay_seconds: 2
  26   │       period_seconds: 10
  27   │       success_threshold: 1
  28   │       timeout_seconds: 4
  29   │     shutdown:
  30   │       signal:
  31   │         _type: override
  32   │         content: 2
  33   │         priority: 1000
```

If I manually edit the file to look like the below:
```
  12   │ processes:
  13   │   postgres:
  14   │     availability:
  15   │       restart: on_failure  # <----- the change happens here
  16   │     command: /nix/store/a1x2cdkaq80hcvfypsn1n02x1vxz6hr9-start-postgres/bin/start-postgres
  17   │     readiness_probe:
  18   │       exec:
  19   │         command: /nix/store/bb052x7g4ybm0yzsdh57583hhc005dgg-postgresql-15.2/bin/pg_isready
  20   │           -h $PGDATA -d template1
  21   │       failure_threshold: 5
  22   │       initial_delay_seconds: 2
  23   │       period_seconds: 10
  24   │       success_threshold: 1
  25   │       timeout_seconds: 4
  26   │     shutdown:
  27   │       signal: 2  # <----- the change happens here
[...]

PR #454, opened 2023-03-02
fix(postgres): remove override options, as currently they are serialized

thenonameguy:
fixes https://github.com/cachix/devenv/issues/453

I tried to fix this properly by modifying the offending line:
https://github.com/cachix/devenv/blob/main/src/modules/processes.nix#L142

But regardless, the YAML file was written with the non-merged values (despite the NixOS module system stating that `config` should be always the final resolved value).
Given the severity of the bug, this is an intermediate quick fix. We should investigate where the override serialization problem exists.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ result
 /include
 /lib
 pyvenv.cfg
+/.direnv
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -6,3 +6,4 @@ result
 /include
 /lib
 pyvenv.cfg
+/.direnv
```
thenonameguy, line 9: Somewhat unrelated change, but I was getting a dirty worktree without this while iterating.

Issue #456, opened 2023-03-03
Starting shell with an old devenv.lock is missing the command description


M-arcus:
**Describe the bug**
Starting shell with an old devenv.lock is missing the command description to update

**To Reproduce**
Any devenv.lock from 0.5.1

```bash
$ devenv shell
Building shell ...
Entering shell ...

✨ devenv 0.6 is newer than devenv input in devenv.lock. Run  to sync.
```

**Version**
```nash
$ devenv version
devenv: 0.6
```

PR #457, opened 2023-03-03
Fix echo output of update command when using an old devenv.lock

M-arcus:
Resolves https://github.com/cachix/devenv/issues/456


domenkozar:
Thanks for fixing this :heart:

PR #461, opened 2023-03-06
lang/nix: add nix-tree

edrex:
I'm not sure if this is in scope for devenv.

nix-tree is a graphical tool for inspecting closures I reach for a lot when I'm trying to optimize dependency trees. It is not a nix language tool per se. Maybe behind a config flag, like `languages.nix.storeTools.enable`?

Feel free to close if it seems out of scope.

### DIFF
--- a/src/modules/languages/nix.nix
+++ b/src/modules/languages/nix.nix
@@ -15,6 +15,7 @@ in
       vulnix
       deadnix
       nil
+      nix-tree
     ];
   };
 }


bobvanderlinden:
For what it's worth, I found out about nix-tree because of this pr. Thanks! It's a great tool 👌

I don't know which packages are a good fit to automatically add and which ones aren't, but for exposure it is great.

edrex:
looking at the other languages, only compilers, runtimes, build tools, and language servers are included, so I think this is out of scope. Another repo can provide a bundle for extra nix tooling, as a devenv or/and a flake-parts module. :beers:

edrex:
TODO: make a "devenv-extras" repo which provides additional language tooling bundles.

PR #475, opened 2023-03-09
javascript: add corepack option

nicknovitski:
```nix
# Before: 
{
  languages.javascript.enable = true;
  languages.javascript.package = pkgs.nodejs-16_x;
  packages = [ (pkgs.yarn.override { nodejs = pkgs.nodejs-16_x; }) ];
}

# Or alternatively:
{
  languages.javascript.enable = true;
  languages.javascript.package = pkgs.nodejs-16_x;
  scripts.yarn.exec = ''exec corepack yarn "$@"'';
}

# But with this change:
{
  languages.javascript.enable = true;
  languages.javascript.package = pkgs.nodejs-16_x;
  languages.javascript.corepack.enable = true;
}
```

### DIFF
--- a/src/modules/languages/javascript.nix
+++ b/src/modules/languages/javascript.nix
@@ -13,11 +13,18 @@ in
       defaultText = lib.literalExpression "pkgs.nodejs";
       description = "The Node package to use.";
     };
+
+    corepack = {
+      enable = lib.mkEnableOption "shims for package managers besides npm";
+    };
   };
 
   config = lib.mkIf cfg.enable {
     packages = [
       cfg.package
-    ];
+    ] ++ lib.optional cfg.corepack.enable (pkgs.runCommand "corepack-enable" { } ''
+      mkdir -p $out/bin
+      ${cfg.package}/bin/corepack enable --install-directory $out/bin
+    '');
   };
 }


domenkozar:
Seems like you need to run `pre-commit run -a`

nicknovitski:
```
~~> Setting up the default profile
  installing 'nix-2.13.3'
  error: unexpected EOF reading a line
```

Since this happened on macos, I assume it's this: NixOS/nix#3605.

shyim (approved):

tennox:
I'm curious if corepack brings any advantage over nix-native installing of package-managers, and if the problem described couldn't be fixed in a more nix-y way (and not installing package managers later via this shim) by adding package-manager selection & overlay handling to devenv's `languages/javascript.nix`.

Also, adding an [overlay with the nodejs version selected](https://discourse.nixos.org/t/how-to-use-pnpm-with-recent-nodejs/21867/4?u=tennox) might be a good idea so if users install e.g. yarn/pnpm, it will use the right version. :thinking:

shyim:
@tennox I guess it always depends on the team. If it's required to use Nix for the setup definitively Yes. But if an project have multiple setups, I guess this solution is okay. Also it's optional :)

PR #490, opened 2023-03-15
refactor(dotnet): add package option and needed environment variables

ratsclub:
This PR introduces:

- `package` option
- [DOTNET_ROOT](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables#dotnet_root-dotnet_rootx86) environment variable
- `icu` package on the `LD_LIBRARY_PATH` for [globalization](https://learn.microsoft.com/en-us/dotnet/core/extensions/globalization-icu)

### DIFF
--- a/src/modules/languages/dotnet.nix
+++ b/src/modules/languages/dotnet.nix
@@ -6,11 +6,21 @@ in
 {
   options.languages.dotnet = {
     enable = lib.mkEnableOption "tools for .NET development";
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.dotnetCorePackages.sdk_7_0;
+      defaultText = lib.literalExpression "pkgs.dotnet-sdk";
+      description = "The .NET SDK package to use.";
+    };
   };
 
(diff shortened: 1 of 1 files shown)


src/modules/languages/dotnet.nix
```diff
@@ -6,11 +6,21 @@ in
 {
   options.languages.dotnet = {
     enable = lib.mkEnableOption "tools for .NET development";
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.dotnetCorePackages.sdk_7_0;
```
ratsclub, line 12: According to the [manual](https://nixos.org/manual/nixpkgs/unstable/#dotnetcorepackages.sdk-vs-dotnetcorepackages.runtime-vs-dotnetcorepackages.aspnetcore):
> `dotnetCorePackages.sdk` contains both a runtime and the full sdk of a given version

Should be a good default package, however, we don't have a generic `dotnetCorePackages.sdk` package pointing to the latest release. This is why I set 7.0 as the default one.

PR #492, opened 2023-03-16
[wip] improve c support

Mic92:
I would by default only provide the default toolchain from nixpkgs itself and also make
sure that CC/LD/STRIP/NM environment variables are set so that buildsystems pickup the devenv compiler rather than defaulting to impure /usr/bin/gcc etc.
Furthermore I would only provide one module for both c/c++ as these cannot be separated and configurations would conflict with each other.

Also interesting would be to have some sane abstraction for cross compiling as this is a complex topic where the nix documentation is not super helpful.
Let me know what you think about the direction.

### DIFF
--- a/src/modules/languages/c.nix
+++ b/src/modules/languages/c.nix
@@ -6,16 +6,29 @@ in
 {
   options.languages.c = {
     enable = lib.mkEnableOption "tools for C development";
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.stdenv.cc;
+      example = "pkgs.clangStdenv.cc";
+      defaultText = "pkgs.stdenv.cc";
+      description = "C/C++ toolchain to use. Defaults to nixpkgs default C compiler (i.e. gcc on Linux and clang on macOS)";
+    };
+    languageServer = lib.mkOption {
+      type = lib.types.nullOr lib.types.package;
+      default = pkgs.clang-tools;
+      example = pkgs.ccls;
+      defaultText = "pkgs.clang-tools";
+      description = "Language server to use. Defaults to clangd from the clang-tools package.";
+    };
   };
 
   config = lib.mkIf cfg.enable {
     packages = with pkgs; [
-      stdenv
       gnumake
-      clang
-      ccls
-      gcc
       pkg-config
+
+      cfg.package
+      cfg.languageServer
     ];
   };
 }


src/modules/languages/c.nix
```diff
@@ -6,16 +6,29 @@ in
...
+      example = "pkgs.clangStdenv.cc";
+      defaultText = "pkgs.stdenv.cc";
+      description = "C/C++ toolchain to use. Defaults to nixpkgs default C compiler (i.e. gcc on Linux and clang on macOS)";
+    };
+    languageServer = lib.mkOption {
+      type = lib.types.nullOr lib.types.package;
+      default = pkgs.clang-tools;
+      example = pkgs.ccls;
+      defaultText = "pkgs.clang-tools";
+      description = "Language server to use. Defaults to clangd from the clang-tools package.";
```
Mic92, line 21: Clangd is better maintained and also maintained by the original author of ccls. It scales very well also to large projects like the linux kernel.

src/modules/languages/c.nix
```diff
@@ -6,16 +6,29 @@ in
...
+    };
   };
 
   config = lib.mkIf cfg.enable {
     packages = with pkgs; [
-      stdenv
```
Mic92, line 13: I don't see what this package adds. It only contains nix-support hooks which I do not see being used here.

src/modules/languages/c.nix
```diff
@@ -6,16 +6,29 @@ in
...
       gnumake
-      clang
-      ccls
-      gcc
       pkg-config
```
Mic92, line 28: PKG_CONFIG_PATH should point to devenv for this to be useful.
Something similar should exists for `$CMAKE_INCLUDE_PATH` for cmake.

domenkozar: It does, but it's part of the profile

domenkozar: That would be great :)

src/modules/languages/c.nix
domenkozar: I don't remember why I added it, should have added a comment. Let's see if it breaks something.

domenkozar:
Hey @Mic92, do you still want to do this now that naked shell has been removed?

Mic92:
We can close this. I see value in overriding the compiler with overrideCC, but currently do not have plans to add this.

PR #493, opened 2023-03-16
use miliseconds when creating shell in case we launch two

domenkozar:


bobvanderlinden (approved):
Nice, this is an improvement. Avoiding conflicts altogether would require a bigger change.

PR #496, opened 2023-03-17
Fix two errors in Rosetta pattern

hynek:
Especially the missing semicolon leads to an interminable error output.


domenkozar:
Thanks :heart:

PR #498, opened 2023-03-17
feat(shell): add option to turn off naked shell implementation, disable for C compiler using languages

thenonameguy:
fixes https://github.com/cachix/devenv/issues/497

Issue #497, opened 2023-03-17
Set subset of setup-hook provided env vars automatically, like NIX_CFLAGS_COMPILE


thenonameguy:
**Describe the bug**
Thanks to #191, we are not using the Nix build system's setup-hook mechanism anymore.
This gives us faster devenv evaluation, but comes with the tradeoff that certain compilation-specific implementation details are not carried over.

One of these is the C-compiler flags that are passed down through env vars to the Nix-provided GCC/Clang stdenv compilers, that are non-standard. This leads to issues like #267 and #258.

It seems we need to find a middle-ground between naked shell and the plain `nix-shell` to have a better out-of-the-box experience for compiling things within a `devenv shell`.

See description of how the normal compilation workflow is extended:
https://nixos.wiki/wiki/C

Some research was made here:
- https://github.com/cachix/devenv/issues/267#issuecomment-1473510291
- https://github.com/cachix/devenv/issues/267#issuecomment-1473515745

thenonameguy:
As a middle-ground I'll create an option to switch between using naked shell vs `nix-shell`-like behavior and turn naked shell off  for the language modules that commonly invoke a C compiler (C-like langs, Rust, Ruby). Should fix most of the woes related to the issue.

domenkozar:
Could we just source some of the needed setup-hooks instead all of them?

thenonameguy:
> Could we just source some of the needed setup-hooks instead all of them?

Deciding that would be:
1. pkg setup-hook dependent, usually they contain a single setup hook that does multiple things
2. while also depending on the given nixpkgs commit, as the setup-hooks evolve over time

Therefore deciding to which setup hooks to source is too cumbersome. We could run them regardless and keep a subset of the env vars `NIX_.*`, etc. to be a bit more safe.

domenkozar:
This is fixed in #745

thenonameguy:
Closing as the 1.0 major version fixed this for the most part.

Issue #486, opened 2023-03-13
Cannot import `pkgs.glibc` in flake with devenv


aaronmondal:
**Describe the bug**
While porting the [Nix flake in rules_ll](https://github.com/eomii/rules_ll/blob/main/flake.nix) to devenv, I found that `pkgs.glibc` in the `packages` section will lead to `nix develop` hanging indefinitely on startup.

Seems depending on glibc somehow causes binaries to use an incompatible version. In my case I think this can be worked around by adding paths to glibc headers and library in scripts manually and passing that as environment variables, but the current behavior doesn't seem like it's WAI.

**To Reproduce**

```nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-22.11";
    devenv.url = "github:cachix/devenv";
  };

  outputs = { self, nixpkgs, devenv, ... } @ inputs:
    let
      systems = [ "x86_64-linux" "i686-linux" "x86_64-darwin" "aarch64-linux" "aarch64-darwin" ];
      forAllSystems = f: builtins.listToAttrs (map (name: { inherit name; value = f name; }) systems);
    in
    {
      devShells = forAllSystems
        (system:
          let
            pkgs = import nixpkgs {
              inherit system;
            };
          in
          {
            default = devenv.lib.mkShell {
              inherit inputs pkgs;
              modules = [
                {
                  # https://devenv.sh/reference/options/
                  packages = [
                    # Base dependencies.
                    pkgs.hello
                    pkgs.glibc
                  ];

                  enterShell = ''
                  hello
                  '';
                }
              ];
            };
          });
    };
}
```

<details>
<summary>flake.lock</summary>

```json
{
  "nodes": {
    "devenv": {
      "inputs": {
        "flake-compat": "flake-compat",
        "nix": "nix",
        "nixpkgs": "nixpkgs",
        "pre-commit-hooks": "pre-commit-hooks"
      },
      "locked": {
        "lastModified": 1678717095,
        "narHash": "sha256-fRUvPhdYNd/B15oKEdqKSavmOyDjXHpGJK+dlSwZWQQ=",
        "owner": "cachix",
        "repo": "devenv",
        "rev": "b331906925f3ba23d67e461b14ed042c743259c4",
        "type": "github"
      },
      "original": {
        "owner": "cachix",
        "repo": "devenv",
        "type": "github"
      }
    },
    "flake-compat": {
      "flake": false,
      "locked": {
        "lastModified": 1673956053,
[...]

PR #507, opened 2023-03-24
feat(shell): revert from naked shell to pkgs.mkShell + unsetting irrelevant env vars

thenonameguy:
closes https://github.com/cachix/devenv/pull/498
fixes https://github.com/cachix/devenv/issues/497


aaronmondal:
Also fixes https://github.com/cachix/devenv/issues/486.

aaronmondal:
Ok I tested this against the flake in [rules_ll](https://github.com/eomii/rules_ll) and things seem to work fine. This change of course leads to more environment variables being set than before, but at least we know that it won't break existing more "exotic" C++ toolchains 😊

domenkozar:
It seems that Nim is broken in nixpkgs, not here: https://hydra.nixos.org/build/213226629

thenonameguy:
> It seems that Nim is broken in nixpkgs, not here: https://hydra.nixos.org/build/213226629
> 
> Possible fix [NixOS/nixpkgs#222285](https://github.com/NixOS/nixpkgs/pull/222285)

Thanks, I removed irrelevant change.

sandangel:
Also my neovim could not find `tsserver` installed by `languages.typescript.enable = true;`. It was working when I use the standard `mkShell`.

hurricanehrndz:
Although this patch addresses a lot of issues I am still seeing some issues on nix-darwin, specifically with CoreFoundation.

sandangel:
Hi, sorry may I ask is there any update on this PR?

domenkozar:
I mainly would like people to test it thoroughly before it's merged.

sandangel:
I think it will be easier for people to test if we merge after all the CI check passed. If there is any issue, we can create a PR to fix. WDYT?

aaronmondal:
I was actually wondering: What are the benefits of unsetting all those environment variables? I didn't really get the "keep it lean" comment.

sandangel:
Hi, how long do we expect people to test this out before merging?

bobvanderlinden:
I've updated this branch in my fork: https://github.com/bobvanderlinden/devenv/tree/feat/stripped-shell

There will be some regressions when merging this. Question is whether that is a good thing (honestly do not know).

With `mkNamedShell`, `LD_LIBRARY_PATH` was set based on `packages`. This is not the case anymore. This is partly a good thing, as I think setting potentially 'injecting' many libraries using `LD_LIBRARY_PATH` will cause problems like https://github.com/cachix/devenv/issues/555.

This is why the `python-poetry` example needed a change:

https://github.com/bobvanderlinden/devenv/commit/b669f22a471e06fd356e04b5b63cbb2da5fe5e2e#diff-3ea46e5d7bf8933097511b155be31bf25b2c6e01081c6d9a6dfe61624717082aR4

Question is whether setting `LD_LIBRARY_PATH` inside `mkShell` should now be added again.

Maybe as an alternative we should introduce a different option. Something like `runtimeLibraries = [ pkgs.zlib ]` instead of `packages = [ pkgs.zlib ]`, which is meant to solely set `LD_LIBRARY_PATH` and `DYLD_LIBRARY_PATH`. People can still use it, but need to do so more explicitly.

The real solution is for people to not rely on pre-built binaries, but I think that's the world we live in right now. `devenv` is a good middleground for that.

aaronmondal:
`LD_LIBRARY_PATH` should [not be set at all if possible](http://xahlee.info/UnixResource_dir/_/ldpath.html). I think the linked python-poetry change is desirable.

> Maybe as an alternative we should introduce a different option. Something like runtimeLibraries = [ pkgs.zlib ] instead of packages = [ pkgs.zlib ], which is meant to solely set LD_LIBRARY_PATH and DYLD_LIBRARY_PATH. People can still use it, but need to do so more explicitly.

IMO it's risky to let users unknowningly manipulate these variables, but if it's explicit it could be a viable option. Though I wonder whether such an option is actually necessary. Personally, I think I'd prefer explicitly setting `LD_LIBRARY_PATH="${somepackage}/lib/"` manually or do something similar to what you did in python-poetry.

bobvanderlinden:
> is it possible now to add openssl dev deps?

Yes, the packages are added as `nativeBuildInputs`.

See https://github.com/NixOS/nixpkgs/blob/3ed18a352739221ebaba0d11c38a2158faa29887/pkgs/build-support/mkshell/default.nix#L37

> Personally, I think I'd prefer explicitly setting `LD_LIBRARY_PATH="${somepackage}/lib/"` manually or do something similar to what you did in python-poetry.

There are people already relying on just adding runtime libraries as packages. Having a separate option for runtimeLibraries will make it a bit clearer how to migrate. It also allows to document what it does and why it is a bad idea to use it. The specific use-cases where it's the only option is also good to mention in the docs.

I agree it's a bad idea in general, but shipping binaries is the underlying problem. That's sometimes quite hard to get around when using ruby/python/node package managers.

asymmetric:
This seems like the right thing to do. Setting `LD_LIBRARY_PATH` comes with troubling consequences, which are additionally quite hard to debug unless one knows already what's going on.

domenkozar:
I've done testing today on examples on top of #745 and python-poetry example fails with

```
+ poetry run python -c 'import numpy'
  File "/tmp/python-poetry_nj6_mkml/.venv/lib/python3.10/site-packages/numpy/core/__init__.py", line 23, in <module>
    from . import multiarray
  File "/tmp/python-poetry_nj6_mkml/.venv/lib/python3.10/site-packages/numpy/core/multiarray.py", line 10, in <module>
    from . import overrides
  File "/tmp/python-poetry_nj6_mkml/.venv/lib/python3.10/site-packages/numpy/core/overrides.py", line 6, in <module>
    from numpy.core._multiarray_umath import (
ImportError: libz.so.1: cannot open shared object file: No such file or directory
```

We'll need to fix this one, any ideas? I tried adding `pkgconfig` to the deps and explicilty setting `zlib.dev` as a package.

thenonameguy:
Great, thanks for bringing this forward. Closing this PR.

domenkozar:
Thank you for making this happen :heart:

cameronraysmith:
I imagine

<details><summary>this instance of zlib.h not found</summary>
<p>

  creating build/temp.linux-x86_64-cpython-310/pybedtools/include
  gcc -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -I/nix/store/ixvka49i2mjnxzk8l8anykc1idilws1r-libxcrypt-4.4.36/include -fPIC -Ipybedtools/include/ -I/tmp/tmpd_84vc8x/.venv/include -I/nix/store/50kabpj0s79040a42b7jj2dxn85wmbfd-python3-3.10.13/include/python3.10 -c pybedtools/cbedtools.cpp -o build/temp.linux-x86_64-cpython-310/pybedtools/cbedtools.o
  In file included from pybedtools/include/bedFile.h:16,
                   from pybedtools/cbedtools.cpp:802:
  pybedtools/include/gzstream.h:35:10: fatal error: zlib.h: No such file or directory
     35 | #include <zlib.h>
        |          ^~~~~~~~
  compilation terminated.
  error: command '/nix/store/6jnx2r93796cnqr5dznrgc1bwgyl488b-devenv-profile/bin/gcc' failed with exit code 1
  

  at /nix/store/1hhzpyy8rvjavf894qy0r98hq8ah59qb-python3.11-poetry-1.7.1/lib/python3.11/site-packages/poetry/installation/chef.py:164 in _prepare
      160│ 
      161│                 error = ChefBuildError("\n\n".join(message_parts))
      162│ 
      163│             if error is not None:
    → 164│                 raise error from None
      165│ 
      166│             return path
      167│ 
      168│     def _prepare_sdist(self, archive: Path, destination: Path | None = None) -> Path:

Note: This error originates from the build backend, and is likely not a problem with poetry but with pybedtools (0.9.1) not supporting PEP 517 builds. You can verify this by running 'pip wheel --no-cache-dir --use-pep517 "pybedtools (==0.9.1)"'.

</p>
</details> 

is another of https://github.com/cachix/devenv/pull/507#issuecomment-1660105656,

in this case, from

<details><summary>devenv devshell in flake.nix + pyproject.toml.</summary>
<p>

```nix
    devShells =
      forEachSystem
      (system: let
      in {
          inherit inputs pkgs;
          modules = [
            {
              packages = with pkgs; [
                atuin
                bat
                bedtools
                gcc
                gh
                git
                gnumake
                htslib
                lazygit
                poethepoet
                ripgrep
                starship
                tree
                zlib
                zlib.dev
                zsh
              ];

              dotenv = {
                enable = true;
                filename = ".env";
                # disableHint = true;
              };

              pre-commit.hooks = {
                alejandra.enable = true;
                ruff.enable = true;
                pyright.enable = true;
              };

              difftastic.enable = true;

              languages.python = {
                enable = true;
                package = pkgs.python310;
                poetry = {
                  enable = true;
                  activate.enable = true;
                  install = {
                    enable = true;
                    installRootPackage = true;
                    groups = [
                      "lint"
                      "test"
                      "docs"
                      "workflows"
                    ];
                  };
                };
              };
            }
          ];
        };
      });
```

with

```toml
[tool.poetry.dependencies]
pybedtools = "0.9.1"
```

</p>
</details> 

If this isn't a mistake, is there a recommended workaround?

PR #510, opened 2023-03-31
introduce mkDevenv

zimbatm:
Separate the concerns of mkNakedShell into two:

* mkDevenv creates the profile, a buildEnv that contains all the
  packages, plus an /etc/profile that contains all the environments,
  plus a /bin/devenv script that can be used to enter said environment.
* mkNakedShell is just the wrapper that should be used as an adapter for
  `nix develop` and `nix-shell`. `inputDerivation` is kept points to the
  profile.

That makes the reasoning a bit cleaner, and also opens the road of
removing mkNakedShell entirely once the entrypoint is stable enough.

> TODO: This isn't working yet. There is an issue because env vars are being
>       used at Nix eval time right now and DEVENV_PROFILE cannot be calulated
>       upfront anymore. Another issue is that DEVENV_ROOT depends on the repo
>       location, making the eval result vary between computers.
>
>       I wanted to share this experiment to get some feedback on the general
>       idea, but in order to finish it, I propose to move all the env vars to
>       be used at runtime instead.


domenkozar:
See #507 as we're having trouble with too many things breaking because we don't support setup-hooks.

Do you have an idea if we could support some setup-hooks using nakedshell approach?

zimbatm:
Using stdenv will make it harder to avoid invoking `nix-shell` or `nix develop` (in contrast with `nix run .#devenv`), and our reload ideas. It also comes with its own set of problems as it tends to assume that the user is inside of a build sandbox, and the logic varies between versions of nixpkgs.

The best is to build an interface specific to devenv, and understand the underlying reason of why those packages are breaking. But it also brings more logic and work to devenv, so I understand if you want to avoid making that tradeoff.

thenonameguy:
I agree with @zimbatm on the purity aspect, that's why I did the toggle-able implementation here:
https://github.com/cachix/devenv/pull/498

This way we keep the naked shell / `nix run` route, for folks who don't compile C, while allowing higher-level language users to benefit from the speed of not using `stdenv`. The concern in the PR from Domen is very valid, as every bug report might use either one of these approaches with their own solutions. :thinking:

PR #517, opened 2023-04-03
fix: hanging direnv

shyim:
image

direnv hangs without that and I cannot accept it


shyim:
yes it is fish

sandydoo:
Possibly related direnv issue: https://github.com/direnv/direnv/issues/967

shyim:
seems to work

PR #522, opened 2023-04-06
python: rework how venv and poetry are handled

bobvanderlinden:
Currently devenv manages its own virtualenv in DEVENV_STATE/venv. In most circumstances this works fine. However, since poetry in some cases wants to manage its own virtualenv it breaks down.

To reproduce the bug:

* Initialize devenv with venv and poetry.
* Initialize pyproject.toml
* Change the python version in pyproject.toml to one that doesn't match the python version in devenv.
* Run poetry install _or_ rerun devenv

Poetry will (by default) create a new virtualenv in $HOME and use that, instead of the DEVENV_STATE/venv one from devenv.

[It is not possible to instruct poetry to specifically use the virtualenv that devenv created](https://github.com/python-poetry/poetry/issues/1579). However, it is possible to instruct poetry to use `PROJECT_ROOT/.venv` and it is able to create it when it's gone.

I opted to also use `PROJECT_ROOT/.venv` when `languages.python.venv.enable` is set instead of `DEVENV_STATE/venv`. I found that it is a convention to use `PROJECT_ROOT/.venv` and tools (like poetry) are presuming this is the case.

Lastly, the change also includes a fix for an issue that happens when poetry install fails. When it failed previously the checksum of the lock file was still stored. This results in devenv not running poetry install anymore the next time devenv runs. Now devenv tells the user that installation fails and to run `devenv install` manually.

### DIFF
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -104,14 +104,8 @@ in
 
       package = lib.mkOption {
         type = lib.types.package;
-        default = pkgs.poetry.override {
-          python3 = cfg.package;
-        };
-        defaultText = lib.literalExpression ''
-          pkgs.poetry.override {
-            python3 = config.languages.python.package;
-          }
-        '';
+        default = pkgs.poetry;
+        defaultText = lib.literalExpression "pkgs.poetry";
         description = "The Poetry package to use.";
       };
     };
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -22,18 +22,22 @@ let
   '';
 
   initPoetryScript = pkgs.writeShellScript "init-poetry.sh" ''
-    if [ ! -f pyproject.toml ]
-    then
-      echo "No pyproject.toml found. Run 'poetry init' to create one." >&2
-    elif [ ! -f poetry.lock ]
-    then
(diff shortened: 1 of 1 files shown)


bobvanderlinden:
> One nice property that we have currently is that all state is stored inside DEVENV_STATE, could we somehow symlink `.env` there to comply with it?

Good suggestion :+1: That would indeed be nice.

I worry that that it doesn't work nicely with poetry. Since poetry will recreate `.venv` whenever it find that the python version inside .venv is not up-to-date with the current python version. It can happen that poetry will create the directory, which results in a bit of a shaky state (devenv looks at DEVENV_STATE/venv, poetry looks at .venv).

I'll check whether poetry actually recreates the .venv or whether it will automatically follow the symlink.

bobvanderlinden:
I must admit that I was a bit skeptic about the .venv symlink, thinking that poetry wouldn't handle it properly, but I was wrong. I have moved the virtualenv to DEVENV_STATE/venv again and made sure PROJECT_ROOT/.venv always links to DEVENV_STATE/venv. This even works when DEVENV_STATE/venv is gone and poetry wants to recreate the venv: it'll do so in DEVENV_STATE/venv without overwriting the PROJECT_ROOT/.venv symlink :+1:

Thanks for the suggestion, this makes things quite a bit better :+1:

PR #537, opened 2023-04-20
mysql: default to pkgs.mariadb

domenkozar:


domenkozar:
@stephank seems like swift on nixpkgs-master is broken?

```
57 Illegal instruction     (core dumped) TERM=dumb swift-build -c release
```

stephank:
I'm not sure about that crash, but it currently doesn't even build on Linux because of the glibc upgrade: https://github.com/NixOS/nixpkgs/pull/188492#issuecomment-1494543503

I'm now first working on the Swift 5.8 upgrade: https://github.com/NixOS/nixpkgs/pull/225197

domenkozar:
I'm going to merge this as mysql is broken currently.

PR #541, opened 2023-04-20
dotnet: disable telemetry and welcome message

ratsclub:
Reference: https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-environment-variables


shyim (approved):

ratsclub:
I agree with this, I also don't think we should document the tooling itself on this module's option.

PR #548, opened 2023-04-21
devenv: use explicit package paths for executables

bobvanderlinden:
As a follow-up from #545 I thought there might be more instances where executable were used from the global system. This turned out to be true.

I'm not sure whether any of the currently used executables result in incompatiblities, but I thought to avoid such potential problems in the future.

I also took the liberty to align the nix and devenv-yaml calls so they look the same as calls to other packages.

I checked for usage of other external commands by using `PATH=""` in the top of the devenv-script and look for command not found errors.

### DIFF
--- a/src/devenv.nix
+++ b/src/devenv.nix
@@ -3,6 +3,13 @@ let
   examples = ../examples;
   lib = pkgs.lib;
   version = lib.fileContents ./modules/latest-version;
+  inherit (pkgs)
+    coreutils
+    findutils
+    gnugrep
+    jq
+    docopts
+    util-linuxMinimal;
 in
 pkgs.writeScriptBin "devenv" ''
   #!/usr/bin/env bash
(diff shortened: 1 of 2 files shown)


src/devenv.nix
```diff
@@ -3,6 +3,14 @@ let
...
+    coreutils
+    findutils
+    gnugrep
+    jq
+    docopts
+    util-linuxMinimal;
+  devenv-yaml = import ./devenv-yaml.nix { inherit pkgs; };
 in
 pkgs.writeScriptBin "devenv" ''
   #!/usr/bin/env bash
```
shyim, line 16: do we want to reference here also our bash 🤔

bobvanderlinden: Oof, good point :+1: I've changed to it use `pkgs.bash`!

bobvanderlinden:
Hmm CI is failing because it tries to run:

`devenv ci`

with the new changes, which results in:

```
/nix/store/00i4b6rg45xinc8jhf4cfdamnpisn02l-devenv-profile/bin/devenv: line 228: /nix/store/isia7synl0mdbcvfwav5ywnrwmwbwvxv-source/bin/nix: No such file or directory
```

There seems to be a mixup of `src/devenv.nix` with an older `flake.nix`. This is because `src/devenv.nix` is imported from other places than just `/flake.nix`.

I have now made sure the original call signature is used for `src/devenv.nix`, so it stays compatible. I did the resolve nix to `nix.packages.${pkgs.stdenv.system}.nix` inside `src/devenv.nix` instead of inside `/flake.nix`.

bobvanderlinden:
The `python-poetry` test should've been resolved by https://github.com/cachix/devenv/pull/552

PR #552, opened 2023-04-24
python: avoid using existing virtual env or poetry binary

bobvanderlinden:
Currently `poetry` is called in one instance with just `poetry`. There were cases where the poetry binary from `devenv`'s root project leaked into the devenv of the python-poetry test. This often is not a problem, but in this case it was a poetry version that was built against a different glibc, resulting in an error.

I've tried to harden the poetry module a bit more by explicitly calling all executables to avoid leakage of executables from a parent environment.

While debugging also found a minor issue when manually executing poetry after removing `.venv`, poetry will create a new `.venv` directory. In this case devenv should remove it and create a new symlink. This is what `ln -sf` was meant to do, however it instead created a symlink inside the `.venv` directory. I tried to avoid this by using `--no-target-directory`.

Lastly, this resets the `VIRTUAL_ENV` so that poetry won't try to use `python` from a `VIRTUAL_ENV` that was already active before devenv shell started.

### DIFF
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -17,7 +17,7 @@ let
       ${lib.optionalString cfg.poetry.enable ''
         [ -f "${config.env.DEVENV_STATE}/poetry.lock.checksum" ] && rm ${config.env.DEVENV_STATE}/poetry.lock.checksum
       ''}
-      python -m venv ${venvPath}
+      ${cfg.package.interpreter} -m venv ${venvPath}
       ln -sf ${config.env.DEVENV_PROFILE} ${venvPath}/devenv-profile
     fi
     source ${venvPath}/bin/activate
@@ -34,7 +34,7 @@ let
       if [ ! -d ${venvPath} ] \
         || [ ! "$(readlink ${venvPath}/bin/python)" -ef "${cfg.package.interpreter}" ]
       then
-        poetry env use --no-interaction ${cfg.package.interpreter}
+        ${cfg.poetry.package}/bin/poetry env use --no-interaction ${cfg.package.interpreter}
       fi
     }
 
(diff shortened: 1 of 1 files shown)


domenkozar:
Seems to still have the same failure

bobvanderlinden:
> Seems to still have the same failure

I had a hard time understanding what was going wrong. I _think_ it's a combination:

It seems that all examples (that I run using `devenv-test-example`) are run against devenv modules from github:cachix/devenv. To circumvent this during development I used:

```yaml
inputs:
  nixpkgs:
  devenv:
    url: /home/bob.vanderlinden/projects/devenv/src/modules
```

Next it seemed as it poetry would default to using a different Python version than what was in `PATH`. It used the one from the devenv root's venv. Unsetting `VIRTUAL_ENV` seemed to fix the problem locally.

I had a hard time getting stable results from the tests, as there are numerous factors that affect the test:

* Am I in a devenv of `~/projects/devenv` or not? (if not, it already worked)
* Is there a lock file still in examples/python-poetry from a previous debug session?
* Have I still got a devenv entry in examples/python-poetry/devenv.yaml?

Might need to look into the tests for a solution.

examples/python-poetry/devenv.yaml
```diff
@@ -1,3 +1,5 @@
 inputs:
   nixpkgs:
     url: github:NixOS/nixpkgs/nixpkgs-unstable
+  devenv:
```
domenkozar, line 4: I don't think this is needed, because it's added by the testing script to point to the local branch (it also fails CI as it now has double devenv)

bobvanderlinden:
Could you try rerunning CI?

bobvanderlinden:
Yay, it succeeds now :+1: Just MySQL that is still failing.

bobvanderlinden:
Thanks for your patience

domenkozar:
Thanks for fixing this :)

Issue #550, opened 2023-04-22
postgres does not work because /.devenv does not exist


matthiasbeyer:
**Describe the bug**
When I enable postgres via `services.postgresql.enable = true;`, I cannot `pg_start` the database because `/.devenv` does not exist on my system (and I guess it also should not).

**To reproduce**
Use default devenv setup and enable postgres as described above. Don't have `/.devenv` on your system when you do.

**Version**
 
0.6.2

domenkozar:
How did you run devenv?

matthiasbeyer:
IIRC by adding the devenv in my flake as `devShells.default`...

domenkozar:
You need to invoke `nix develop --impure` for devenv to work.

PR #557, opened 2023-04-26
devenv: assert --impure is used

bobvanderlinden:
Currently it is possible to run devenv without `--impure`. However, strange things may happen, because `$PWD` is not set. This results in `env.DEVENV_ROOT` being `""`, which results in various other variables to contain an invalid path (like `/.devenv`).
This will go unnoticed, but can cause problems.

This PR makes fetchign `env.DEVENV_ROOT` throw an error whenever it is fetched without Nix running in `--impure` mode.

I can imagine this can also cause annoyances for people who are using devenv without _actually_ depending on `DEVENV_ROOT`. So it isn't a clear win in that regard. I thought I'd just suggest the change.

This was in response to https://github.com/cachix/devenv/issues/550

### DIFF
--- a/src/modules/top-level.nix
+++ b/src/modules/top-level.nix
@@ -104,7 +104,18 @@ in
 
   config = {
     # TODO: figure out how to get relative path without impure mode
-    env.DEVENV_ROOT = builtins.getEnv "PWD";
+    env.DEVENV_ROOT =
+      let
+        pwd = builtins.getEnv "PWD";
+      in
+      if pwd == "" then
+        throw ''
+          devenv was not able to determine the current directory.
+          Make sure Nix runs with the `--impure` flag.
+
+          See https://devenv.sh/guides/using-with-flakes/
+        ''
+      else pwd;
     env.DEVENV_DOTFILE = config.env.DEVENV_ROOT + "/.devenv";
     env.DEVENV_STATE = config.env.DEVENV_DOTFILE + "/state";
     env.DEVENV_PROFILE = profile;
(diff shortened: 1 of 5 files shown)


domenkozar:
Some weird CI errors popped up.

bobvanderlinden:
I see. Strange indeed. It seems the `system` attribute isn't passed by `callPackage`, `pkgs.system` doesn't exist?
It _seems_ it is only happening on the self-hosted MacOS. Does it have a different version of Nix?

That said, I can work around it by manually supplying `system` using `callPackage { system = pkgs.stdenv.system; }` or something alike. Would that be a good solution?

domenkozar:
That's a good solution, since `pkgs.system` is deprecated and we're using `pkgs.stdenv.system` where appropriate.

src/modules/top-level.nix
```diff
@@ -104,7 +104,18 @@ in
 
   config = {
     # TODO: figure out how to get relative path without impure mode
-    env.DEVENV_ROOT = builtins.getEnv "PWD";
+    env.DEVENV_ROOT =
+      let
+        pwd = builtins.getEnv "PWD";
+      in
+      if pwd == null then
```
domenkozar, line 111:
```suggestion
      if pwd == "" then
```

bobvanderlinden: Good call. Fixed this. Also found a way to test this properly, which also catches this :+1:

bobvanderlinden:
I made mkNakedShell use `pkgs.stdenv.system` instead of `pkgs.system`, added a test that checks whether nix-develop fails in pure mode, lastly fixed the getEnv resulting in `""` instead of `null`.

PR #561, opened 2023-04-27
fix error when devenv ci in devcontainer

### LINKED ISSUE #560: Devcontainer failing with "seccomp_load invalid argument"
**Describe the bug**
When using the default [.devcontainer.json](https://devenv.sh/integrations/codespaces-devcontainer/) I get the following error  `"seccomp_load invalid argument"`.

A solution to the above problem has been suggested at https://github.com/NixOS/nix/issues/5258#issuecomment-921030592 but not included in the [nix.conf](https://github.com/cachix/devenv/blob/main/devcontainer/nix.conf) yet.

**To reproduce**
I got the same issue with multiple devenv.nix configs. I don't think that's relevant here.

In order to reproduce it, add the following line

```
devcontainer.enable = true;
```

to devenv.nix and then run `devenv shell` so that devenv can generate a default .devcontainer.json

**Version**
 
➜ devenv version
devenv: 0.6.2

gsantoro:
Correctly run devenv.sh in a devcontainer.

Solution provided at https://github.com/NixOS/nix/issues/5258#issuecomment-921030592

Closes #560

I have now tested my changes and it works perfectly on my M1 Mac book.

My only concern is that the docker image is 1.47GB which is much bigger than the official image at 944MB. I am not sure what happened there since the only difference is a single line in the nix.conf.

### DIFF
--- a/devcontainer/nix.conf
+++ b/devcontainer/nix.conf
@@ -3,3 +3,4 @@ max-jobs = auto
 build-users-group = nixbld
 experimental-features = nix-command flakes
 sandbox = false
+filter-syscalls = false


gsantoro:
One more thing. when running commands (eg. `devenv search <package>`) from inside the devcontainer running from a named volume I got an error `GLIB not found`. 

I found [here](https://github.com/NixOS/nix/issues/902#issue-154743974) someone having a similar problem.

I solved the problem by `unset LD_LIBRARY_PATH`. I am sure that it is a better way to fix it but I am not sure.

PR #565, opened 2023-05-01
feat: add HashiCorp Vault service

brandonhamilton:


src/modules/services/vault.nix
```diff
@@ -0,0 +1,121 @@
...
+
+    while true
+    do
+      sleep 1
+    done
+  '';
+in
+{
+  imports = [
+    (lib.mkRenamedOptionModule [ "vault" "enable" ] [ "services" "vault" "enable" ])
```
domenkozar, line 68: This isn't needed since we never had `vault.enable`

domenkozar (changes requested):

src/modules/services/vault.nix
brandonhamilton: Thanks for the feedback, I've removed that in the new commit.

domenkozar:
Thanks a lot for contributing this!

I'd ask you to write a basic test to assert value has started, using a special file named `.test.sh`. See an example https://github.com/cachix/devenv/blob/main/examples/python-poetry/.test.sh

examples/vault/.test.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+set -ex
```
domenkozar, line 2: You'll need to run `devenv up&` and wait for the port to be ready

brandonhamilton: Aha, thanks! I now understand the purpose of this test script. I've pushed an updated version.

brandonhamilton:
Yes, I suspect there is some race condition with the test script, as it passes on both the macOs platform tests.

brandonhamilton:
I've pushed a change that will hopefully give better info if the test fails

PR #566, opened 2023-05-01
Add an option to let DEVENV_ROOT be a relative path

Atry:
This option lets DEVENV_ROOT and other devenv environment variables be relative paths, which works without `--impure` flag. However, some services may not support relative paths.


domenkozar:
What's the motivation to have this setting? Having too many knobs like this makes things harder to test and support.

Atry:
There is a TODO about this
> # TODO: figure out how to get relative path without impure mode

Atry:
I think there are users who don't want to use `--impure`

bobvanderlinden:
There are generated configuration files that depend on `DEVENV_ROOT`. If it is set to `.` those configuration files will fail in strange ways.

Maybe adding an option for `devenv.root`:

```
options.devenv.root = mkOption { type = lib.types.path; };
config.assertions = [
  { assertion = config.devenv.root != ""; message = ''devenv.root must be set. This is set to PWD by default, but only if you're running Nix with --impure. Running with --impure is recommended. Otherwise you need to set it manually. "." is possible, but not recommended as there are integrations that need to presume this value to be absolute.''; }
];
config.devenv.root = mkDefault (buitlins.getEnv "PWD");
config.env.DEVENV_ROOT = config.devenv.root;
```

In flakes you can add a `{ devenv.root = "."; }` module to get around the issue.

I do agree that being able to call devenv in pure mode would have benefits, but using `.` at the moment is probably asking for hidden trouble/bugs.

Atry:
> There are generated configuration files that depend on DEVENV_ROOT. 

Is it possible to make sure all generated configuration files include relative paths?

Atry:
I think most of the usage of `env.DEVENV_ROOT` can be replaced with `inputs.self`

bobvanderlinden:
> Is it possible to make sure all generated configuration files include relative paths?

It's a bit tricky. Some commands may be used in different directories while it should still refer to the `.devenv/` directory.

For instance:

https://github.com/cachix/devenv/blob/4da6478ed21c3ab643bed241660c1e8d62be4371/src/modules/processes.nix#L53

It is possible to use `$DEVENV_ROOT` here at runtime, however, this goes wrong if there are devenv environments within devenv environments. You should still be able to use a script from a parent devenv environment and that still needs to be able to refer to the parent `.devenv` directory.

> In the case of flakes, we could just use `inputs.self`!

:thinking: I wasn't aware of that input, but that makes a lot of sense!

Atry:
> It is possible to use $DEVENV_ROOT here at runtime, however, this goes wrong if there are devenv environments within devenv environments. You should still be able to use a script from a parent devenv environment and that still needs to be able to refer to the parent .devenv directory.

For shell scripts generated under `.devenv/path/file`, I would use `"$(dirname "$0")/../.."` to find the work directory

Atry:
`toString inputs.self` is a Nix store path, which is good to read files added to `git`, but not good for read or write ignored files.

domenkozar:
#583

PR #574, opened 2023-05-04
Avoid using DEVENV_ROOT

Atry:
As mentioned in https://github.com/cachix/devenv/pull/566#issuecomment-1534909975, we want to allow for using devenv without `--impure`. This is an attempt to reduce the usage of absolute DEVENV_ROOT


domenkozar:
We need to solve this in another way

PR #576, opened 2023-05-04
Find nix2container and mk-shell-bin from current inputs

Atry:
This PR would simplify the users' setup for indirect dependencies. Users still can change the `nix2container` input using `devenv.inputs.nix2container.follows`


domenkozar:
It would however increase the number of inputs that come by default, which may not be that problematic.

domenkozar:
We now have shortcut of `devenv inputs add`, so I feel like we should keep the inputs minimal and it's easy to add them.

PR #579, opened 2023-05-05
ruby: add bundler options

bobvanderlinden:
Currently there is no way to disable bundler nor is it possible to alter the bundler package.

In addition, there was a problem with the prioritization of the ruby/bundler packages, as modern ruby packages include a bundler executable.

This was reported here: https://github.com/bobvanderlinden/nixpkgs-ruby/issues/72

I also wanted to avoid regressions, so I added a stricter example/test.

Resolves https://github.com/bobvanderlinden/nixpkgs-ruby/issues/72

### DIFF
--- a/examples/ruby/.test.sh
+++ b/examples/ruby/.test.sh
@@ -1,4 +1,7 @@
 #!/bin/sh
 set -ex
 ruby --version | grep "$(cat .ruby-version)"
+ruby -e "puts RUBY_VERSION" | grep "$(cat .ruby-version)"
 ruby -e "require 'puma'"
+bundle --version
+bundle exec ruby -e "puts RUBY_VERSION" | grep "$(cat .ruby-version)"
--- a/examples/ruby/Gemfile
+++ b/examples/ruby/Gemfile
@@ -2,5 +2,7 @@
 
 source "https://rubygems.org"
 
+ruby "3.2.1"
+
 gem "rails"
 gem "puma"
--- a/examples/ruby/Gemfile.lock
+++ b/examples/ruby/Gemfile.lock
@@ -151,5 +151,8 @@ DEPENDENCIES
   puma
   rails
 
+RUBY VERSION
+   ruby 3.2.1p31
+
 BUNDLED WITH
    2.3.24
--- a/examples/ruby/devenv.yaml
+++ b/examples/ruby/devenv.yaml
@@ -3,3 +3,6 @@ inputs:
     url: github:NixOS/nixpkgs/nixpkgs-unstable
   nixpkgs-ruby:
     url: github:bobvanderlinden/nixpkgs-ruby
+    inputs:
+      nixpkgs:
+        follows: nixpkgs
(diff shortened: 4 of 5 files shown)


bobvanderlinden:
@ryansch could you give this a try?

domenkozar:
Seems like the ruby test is still failing

bobvanderlinden:
Found out the problem. Bundler was internally still using the Ruby version from nixpkgs.
When I `pkgs.bundler.override { ruby = cfg.package; }` I ran into a strange issue where `env` inside `mkDerivation` was not supported. This was because `nixpkgs-ruby` was still based on `nixos-22.11`, while devenv was using `nixpkgs-unstable`. The `env` in `mkDerivation` was introduced inbetween these versions(?).
To circumvent this I've let the nixpkgs-ruby input follow nixpkgs in the Ruby example.

bobvanderlinden:
I ran the following, which succeeds locally. :crossed_fingers: that this works for CI as well:

```
git checkout . && git clean -xdf && pushd examples/ruby && mv devenv.yaml devenv.yaml.orig && awk '
            { print }
            /^inputs:$/ {
              print "  devenv:";
              print "    url: path:../../src/modules";
            }
          ' devenv.yaml.orig > devenv.yaml && popd && devenv-test-example ruby
```

bobvanderlinden:
It seems that the self-hosted MacOS machine xcode doesn't seem to be installed (or up-to-date?). The native dependencies fail to build:

```
`try_do': The compiler failed to generate an executable file. (RuntimeError)
You have to install development tools first.
```

I think the x86_64-darwin build automatically comes with xcode installed, so it relies on those libraries to build the native dependencies.

I'm contemplating whether to include `darwin.apple_sdk.frameworks.*` or whether rely on preinstalled xcode.

I see that the rust has:

https://github.com/cachix/devenv/blob/4da6478ed21c3ab643bed241660c1e8d62be4371/src/modules/languages/rust.nix#L66

Should this be part of `languages.c`?

domenkozar:
We should rely on xcode not being installed and list all the deps we need. Most likely #507 addresses this, I'll look into it next week.

For the YAML schema I was hoping we can render it with strictyaml, how can we keep it in sync with the validation logic. Any ideas?

sandydoo:
@bobvanderlinden, if you'd like to rebase this over master, I can help iron out the remaining macos issues 🙂

bobvanderlinden:
> @bobvanderlinden, if you'd like to rebase this over master, I can help iron out the remaining macos issues 🙂

I had some other PRs that were becoming problematic with the lack of a MacOS system, so I have put in some time to get my old x86_64 mac working again. :+1: I was able to reproduce the problem by not installing xcode.

sandydoo:
> EDIT: Ah! I just saw your addition of libllvm, I think that fixed it here as well. Nice 👍 I rebased the PR.

Yeah, sorry, I really should've mentioned that 😂

bobvanderlinden:
Awesome, thanks!

PR #598, opened 2023-05-16
add devenv.schema.json for devenv.yaml

bobvanderlinden:
Currently devenv.yaml doesn't have autocompletion or a schema. Whenever an editor supports the yaml language server it is possible to add schema information using `# yaml-language-server: $schema=`.

This adds a JSON schema for devenv.yaml. It should be hosted on https://devenv.sh/devenv.schema.json

The schema url is used in the simple example's devenv.yaml.

For devenv.yaml in the root of devenv I added the line, but referring to the devenv.schema.json in the same repo. This should already work regardless of whether the schema is published on devenv.sh already.

### DIFF
--- a/devenv.yaml
+++ b/devenv.yaml
@@ -1,3 +1,4 @@
+# yaml-language-server: $schema=./docs/devenv.schema.json
 inputs:
   nixpkgs:
     url: github:NixOS/nixpkgs/nixpkgs-unstable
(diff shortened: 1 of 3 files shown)


shyim (approved):

bobvanderlinden:
> After merge we should adjust https://github.com/SchemaStore/schemastore/blob/master/src/api/json/catalog.json so it works in most editors out of the box :)

I wasn't aware! Great idea :smile::+1:

domenkozar:
For #1005 we have a way to generate this from the types, but it's blocked on https://github.com/moonrepo/schematic/issues/105

PR #599, opened 2023-05-16
devenv: add build subcommand

bobvanderlinden:
As discussed in https://discord.com/channels/1036369714731036712/1100034303913635860

This adds a `devenv build` command. It allows building derivations inside a devenv config. For instance:

```
$ devenv build languages.python.package
```

I explicitly did not pass through any options to `nix build`, as I wasn't sure of the use-cases and the future compatibility. We might want to explicitly add those once the use-case arises.

My use-case for this is to wrap some of the tests of my project into a Nix build, but I didn't want to migrate the project from using devenv to using devenv within flakes.

### DIFF
--- a/src/devenv.nix
+++ b/src/devenv.nix
@@ -148,6 +148,24 @@ pkgs.writeScriptBin "devenv" ''
         $(${nix}/bin/nix $NIX_FLAGS build --print-out-paths --no-link --impure ".#devenv.containers.\"$container\".dockerRun")
       fi
       ;;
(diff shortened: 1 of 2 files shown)


domenkozar:
This is now part of #745 

```
❯ devenv build languages.python.package
• Using Cachix: devenv 
• Built:
• /nix/store/1r6n7v2wam7gkr18gxccpg7p5ywgw551-python3-3.10.12
```

PR #603, opened 2023-05-18
Implement `aws-vault` integration

sestrella:
The following wrapper makes it more convenient to call `terraform` without the need of specifying the `AWS_PROFILE` environment variable or prefixing commands with `aws-vault exec`.

### DIFF
--- a/examples/aws-vault/devenv.nix
+++ b/examples/aws-vault/devenv.nix
@@ -4,6 +4,7 @@
   aws-vault = {
     enable = true;
     profile = "aws-profile";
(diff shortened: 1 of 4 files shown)


domenkozar:
I think this would be better built if it was a separate module that would override `languages.terraform.package` when certain `option.enable = true;` is set.

sestrella:
@domenkozar thank you for the code review, the code associated with the Terraform wrapper was extracted to a separate standalone module. Please let me know your thoughts

PR #605, opened 2023-05-18
modules: fix evaluation when modules emit warnings

sandydoo:
The `warnings` option is used by functions like `lib.mkRenamedOptionModule` and `lib.mkChangedOptionModule` to warn the user to update their configuration. Without this option, evaluation fails on these functions with a missing attribute error.

You can reproduce the error by using a renamed option, like `postgres.enable` (instead of `services.postgres.enable`).


sandydoo:
`warning: applying 'toString' to path '/home/sandydoo/code/cachix/devenv/examples/postgres/devenv.nix' and then accessing it is deprecated, at «stdin»:0`
This extra warning looks like a bug in nixpkgs. It's converting the file path to a string to display it in the next warning, but clearly the `toString` machinery is throwing errors in this case.

Issue #621, opened 2023-05-24
Can't automatically `poetry install --sync` or `poetry install --with dev`


sersorrel:
**Describe the bug**
I use Poetry's dependency groups to separate my project's development dependencies (e.g. mypy, pytest) from runtime dependencies (e.g. flask). To get these installed by Poetry I have to run `poetry install --with dev`, but there's no way to configure devenv to automatically include the `--with dev` when setting `languages.python.poetry.enable = true`, which seriously limits its utility.

Additionally, if I remove a package from my pyproject.toml, it won't be removed from my Poetry-managed virtual environment until I `poetry install --sync` (or in reality, `poetry install --sync --with dev`, else I lose all the dev packages). Again, there's no way to make devenv run this automatically; this results in previously-installed packages hanging around unexpectedly.

**To reproduce**
https://gist.github.com/sersorrel/a5d50dbaed4c0b2566f448ebf15c6d2c

**Version**
devenv: 0.6.2

domenkozar:
@bobvanderlinden do you think it would be possible to support this?

mirkolenz:
I have a similar issue: At the moment, it is not possible to install extras with devenv. Poetry allows one to specify both extras and groups, so it would be great if both are supported 😄

bobvanderlinden:
It is already supported to some extend:

```
languages.python.poetry.install.arguments = [ "--with" "dev" ];
```

I do agree it would be nice to have some convenience options like `languages.python.poetry.install.with = [ "dev" ];` and maybe make that the default.

> Additionally, if I remove a package from my pyproject.toml, it won't be removed from my Poetry-managed virtual environment until I poetry install --sync (or in reality, poetry install --sync --with dev, else I lose all the dev packages).

That's a different problem. I'm guessing you manually remove the package from `pyproject.toml`? The devenv module only looks at the lock file. So, instead of manually editing pyproject.toml, you need to use the command-line commands like:

`poetry add packagename` and `poetry remove packagename`

Otherwise the lock file won't be updated.
If you make changes to `pyproject.toml` manually you'll have to run `poetry lock --no-update` to update the lock file, but not update any packages.

I don't think it's a good idea for devenv to keep track of pyproject.toml. devenv would need to update poetry.lock implicitly, which usually is a git-tracked file. That would be surprising to me.

PR #631, opened 2023-05-30
python: poetry: add groups and extras options

bobvanderlinden:
Resolves https://github.com/cachix/devenv/issues/621

This adds two options.

`groups = [ <group> ]` is the equivalent of using `poetry install --with <group>`.
It only supports adding groups, as I have not looked into supporting `--only` nor `--without`. Maybe someone else could give an opinion of this, as I have never used these cli options before :sweat_smile:.

`extras = [ <extra> ]` is the equivalent of using `poetry install --extras <extra>`.

I have now also made sure that the command line option of `poetry install` are included in the checksum that determines whether `poetry install` should be re-run. If any of the arguments changes, it will run poetry install the next time devenv shell initializes.

### DIFF
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -118,6 +118,11 @@ in
           default = false;
           description = "Whether `poetry install` should avoid outputting messages during devenv initialisation.";
         };
+        groups = lib.mkOption {
+          type = lib.types.listOf lib.types.str;
+          default = [ ];
+          description = "Which dependency-groups to install. See `--with`.";
+        };
       };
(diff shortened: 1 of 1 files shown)


bobvanderlinden:
@sersorrel @mirkolenz could you have a look?

mirkolenz:
@bobvanderlinden Thank you for creating this PR! I especially like the fact that you incorporated the generated install command into the hash 😄

bobvanderlinden:
Np :smile: Since I'm not using it at the moment, let me whether you run into problems

Issue #530, opened 2023-04-12
rust: allow adding more targets


shyim:
I want to compile from mac a arm64 binary. Therefore we need to add a option to allow more targets. This seems to be possible with fenix and I have no glue how to configure this.

zoechi:
I want to compile `wasm32-unknown-unknown"

domenkozar:
See https://github.com/cachix/devenv/pull/900

MForster:
> I want to compile `wasm32-unknown-unknown`

FWIW, this workaround seems to work for me:

```nix
{ pkgs, fenix, ... }: {
  languages.rust = let ff = fenix.packages.x86_64-linux; in {
    enable = true;
    channel = "nightly";
    toolchain.rustc = ff.combine [
      ff.minimal.rustc
      ff.targets.wasm32-unknown-unknown.latest.rust-std
    ];
  };
  # ...
}

domenkozar:
Example is now on devenv.sh :)

PR #640, opened 2023-06-02
Make rust.nix accept toolchain options.

nothingnesses:
This makes the Rust option accept toolchain override options in a similar format to [rust-toolchain.toml](https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file), which should be more familiar to Rust users and allow adding arbitrary components and targets, and is a step forward to addressing #530.

Changes:
* Remove `languages.rust.packages` and `languages.rust.version` options.
* Make nix-community/fenix a hard-requirement for `languages.rust.enable`.
* Add `languages.rust.toolchain` option.

~~This no longer includes rust-src by default and no longer has pre-commit hooks since the components may no longer be present. How should I handle rust-src and the pre-commit hooks instead?~~

According to the nix-community/fenix maintainer, rust-analyzer will automatically use the rust-src package if it exists. I've pushed a commit which causes the pre-commit hooks to use the rust components if they are added by the user, otherwise it defaults to the ones from nixpkgs.

[Matrix conversation about the topic](https://matrix.to/#/!plrRoZsBTUYBWzvzIq:matrix.org/$YPphSP3OyUVegK01ZV5MLjECbhHiEOL8buqzCytiNJc?via=matrix.org&via=glorifiedgluer.com&via=kde.org).

### DIFF
--- a/docs/overrides/home.html
+++ b/docs/overrides/home.html
@@ -452,8 +452,11 @@ Starting processes ...
   {
     languages.rust = {
       enable = true;
-      # https://devenv.sh/reference/options/#languagesrustversion
-      version = "nightly";
+      # https://devenv.sh/reference/options/#languagesrusttoolchain
+      toolchain = {
+        channel = "nightly";
+        profile = "default";
+      };
     };
 
     pre-commit.hooks = {
--- a/examples/rust/devenv.nix
+++ b/examples/rust/devenv.nix
@@ -7,8 +7,11 @@
 
   languages.rust = {
     enable = true;
-    # https://devenv.sh/reference/options/#languagesrustversion
-    version = "latest";
+    # https://devenv.sh/reference/options/#languagesrusttoolchain
+    toolchain = {
+      channel = "nightly";
+      profile = "default";
+    };
   };
 
   pre-commit.hooks = {
--- a/examples/rust/devenv.nix
+++ b/examples/rust/devenv.nix
@@ -12,7 +12,6 @@
       channel = "nightly";
       profile = "default";
     };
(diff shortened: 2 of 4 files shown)


nothingnesses:
I've updated my branch with [a commit to address the failing tests](https://github.com/nothingnesses/devenv/commit/0a5dab77c7aa7a6ae763c26fa6a5a6a85b1e3674), but it doesn't seem like the latest commit being used in this PR.

nothingnesses:
I've added a commit to fix the failing tests, but I think the test will fail again when Rust updates which will require a new `sha256` value. I haven't managed to come up with a good way to automatically generate these.

nothingnesses:
It seems like the only failing test related to Rust is now due to the VM running out of storage space (I think the rest of the errors are related to Racket instead). I'm wondering if maybe enabling cachix for fenix as per [these instructions](https://github.com/nix-community/fenix/tree/6e6a94c4d0cac4821b6452fbae46609b89a8ddcf#fenix) and using [cachix-actions](https://github.com/marketplace/actions/cachix) would help alleviate this. @figsoda Are derivations produced by the `fromToolchainFile` function cached and do you think it's worth trying this?

figsoda:
> Are derivation outputs produced by the fromToolchainFile function cached in fenix's cachix

they are not, only outputs that are derivations (without needing a function call) are cached

> and do you think it's worth trying this?

I'm not sure if this is possible, since `fromToolchainFile` allows way too many permutations for all of them to be cached

nothingnesses:
Closing in favour of #667.

PR #642, opened 2023-06-02
Add readiness check for memcached

penguincoder:
This adds a simple bash-based readiness check. It will fail if the server is faulted for whatever reason. I chose `bash` over `nc` because it is fairly prevalent throughout all of `nixpkgs` and you are likely to have it already loaded in your store.

### DIFF
--- a/src/modules/services/memcached.nix
+++ b/src/modules/services/memcached.nix
@@ -49,6 +49,24 @@ in
   };
 
   config = lib.mkIf cfg.enable {
-    processes.memcached.exec = "${cfg.package}/bin/memcached --port=${toString cfg.port} --listen=${cfg.bind} ${lib.concatStringsSep " " cfg.startArgs}";
+    processes.memcached = {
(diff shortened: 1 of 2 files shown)


src/modules/services/memcached.nix
```diff
@@ -49,6 +49,24 @@ in
...
 
   config = lib.mkIf cfg.enable {
-    processes.memcached.exec = "${cfg.package}/bin/memcached --port=${toString cfg.port} --listen=${cfg.bind} ${lib.concatStringsSep " " cfg.startArgs}";
+    processes.memcached = {
+      exec = "${cfg.package}/bin/memcached --port=${toString cfg.port} --listen=${cfg.bind} ${lib.concatStringsSep " " cfg.startArgs}";
+
+      process-compose = {
+        readiness_probe = {
+          exec.command = ''
+            ${pkgs.bash}/bin/bash -c 'exec 3<>/dev/tcp/${cfg.bind}/${toString cfg.port}; printf "stats\nquit\n" >&3; cat <&3'
```
thenonameguy, line 58: On *nix-es process-compose already executes the commands via bash, so no need for the extra process.

thenonameguy (changes requested):

thenonameguy (approved):

PR #643, opened 2023-06-02
Add readiness check for rabbitmq

penguincoder:
This adds a readiness check very similar to what is done in the Docker world to check for rabbitmq readiness.

I have some async workers that can boot faster than rabbitmq, so it would be nice to have a healthcheck that gates my code until the process actually responds.

### DIFF
--- a/src/modules/services/rabbitmq.nix
+++ b/src/modules/services/rabbitmq.nix
@@ -143,6 +143,22 @@ in
     env.RABBITMQ_ENABLED_PLUGINS_FILE = plugin_file;
     env.RABBITMQ_NODENAME = "rabbit@localhost";
 
-    processes.rabbitmq.exec = "${cfg.package}/bin/rabbitmq-server";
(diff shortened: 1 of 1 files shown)


src/modules/services/rabbitmq.nix
```diff
@@ -143,6 +143,22 @@ in
...
     env.RABBITMQ_NODENAME = "rabbit@localhost";
 
-    processes.rabbitmq.exec = "${cfg.package}/bin/rabbitmq-server";
+    processes.rabbitmq = {
+      exec = "${cfg.package}/bin/rabbitmq-server";
+
+      process-compose = {
+        readiness_probe = {
+          exec.command = "${cfg.package}/bin/rabbitmq-diagnostics -q ping";
+          initial_delay_seconds = 10;
```
thenonameguy, line 152: This seems a tad bit too high for me.
It should be fine to have the check fail a few times in case the process starts up faster than 10 seconds.

PR #646, opened 2023-06-03
haskell: Parameterize `stack` package

sestrella:
This change makes `stack` package parameterizable so it could be customized or skipped if is set to `null`

### DIFF
--- a/src/modules/languages/haskell.nix
+++ b/src/modules/languages/haskell.nix
@@ -29,16 +29,25 @@ in
         Haskell language server to use.
       '';
     };
+
+    stack = lib.mkOption {
+      type = lib.types.nullOr lib.types.package;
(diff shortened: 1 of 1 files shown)


src/modules/languages/haskell.nix
```diff
@@ -29,16 +29,25 @@ in
...
+      description = ''
+        Haskell stack to use.
+      '';
+    };
   };
 
   config = lib.mkIf cfg.enable {
     packages = with pkgs; [
       cfg.package
       cfg.languageServer
```
sestrella, line 46: While working on this PR, I realized there is a bug in this line which is addressed here #645

sestrella: @domenkozar I just solved the merge conflicts caused by #645

PR #647, opened 2023-06-06
python.venv.requirements

domenkozar:
```nix
{
  languages.python = {
    enable = true;
    version = "3.7.16";

    venv.enable = true;
    venv.requirements = ''
      requests
    '';
  };
}
```

### DIFF
--- a/examples/python/devenv.nix
+++ b/examples/python/devenv.nix
@@ -1,6 +1,13 @@
 { pkgs, ... }:
 
 {
-  languages.python.enable = true;
-  languages.python.version = "3.7.16";
+  languages.python = {
+    enable = true;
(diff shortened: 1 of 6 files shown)


examples/python/devenv.nix
```diff
@@ -1,6 +1,11 @@
...
 
 {
-  languages.python.enable = true;
-  languages.python.version = "3.7.16";
+  languages.python = {
+    enable = true;
+    version = "3.7.16";
+
+    venv.enable = true;
+    venv.requirements = "requests";
```
thenonameguy, line 9: Maybe use a Nix multiline string example to really show the `lines` type.

PR #653, opened 2023-06-07
Use nc for healthcheck

penguincoder:
This is an alternative healthcheck for memcached. The prior one worked when executed with `bash -c 'healthcheck'` but failed silently when I removed the `bash` wrapper. This changes the healthcheck to use `echo | nc`, and I have tested this locally by taking the generated `process-compose.yaml` and modifying the command in the YAML. I have also tested using both `127.0.0.1` and `::1` for the `listen` parameter. I was not as rigorous as I should have been after I modified the last `memcached` PR from feedback.

### DIFF
--- a/src/modules/services/memcached.nix
+++ b/src/modules/services/memcached.nix
@@ -55,7 +55,7 @@ in
       process-compose = {
         readiness_probe = {
           exec.command = ''
-            3<>/dev/tcp/${cfg.bind}/${toString cfg.port}; printf "stats\nquit\n" >&3; cat <&3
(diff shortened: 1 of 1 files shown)


volker-schukai:
If memcache is fed with stat,

```bash
echo stats | ${pkgs.netcat}/bin/nc ${cfg.bind} ${toString cfg.port} > /dev/null 2>&1
``` 

the connection will remain open and if process-compose is used, 
memcache will be terminated due to missing healthcheck.

memchache should be terminated with a quit:

```
echo -e "stats\nquit" | ${pkgs.netcat}/bin/nc 127.0.0.1 11211 > /dev/null 2>&1
```

PR #663, opened 2023-06-14
Make postgres initialDatabases schema creation more robust

stig:
I have a project with ~150 migration files going back to 2016, added by many different people using different conventions. The commits in this PR are required in order for me to successfully apply all those sql files. 

The repo is private (sorry!), but I've tried to give a representative example of each fix in its commit message.

### DIFF
--- a/src/modules/services/postgres.nix
+++ b/src/modules/services/postgres.nix
@@ -37,7 +37,7 @@ let
             if [ -f "${database.schema}" ]
             then
               echo "Running file ${database.schema}"
-              cat "${database.schema}" | postgres --single -j -E ${database.name}
+              ${pkgs.gawk}/bin/awk 'NF' "${database.schema}" | postgres --single -j -E ${database.name}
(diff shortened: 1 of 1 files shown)


src/modules/services/postgres.nix
```diff
@@ -40,8 +40,8 @@ let
               cat "${database.schema}" | postgres --single -E ${database.name}
             elif [ -d "${database.schema}" ]
             then
-              echo "Running sql files in ${database.schema}"
-              cat "${database.schema}/*.sql" | postgres --single -E ${database.name}
+              echo "Running sql files in ${database.schema}
+              cat ${database.schema}/*.sql | postgres --single -E ${database.name}
```
stig, line 44: Another option, if you want the quotes, is:

```suggestion
              find -s "${database.schema}" -name "*.sql" -exec cat {} \; | postgres --single -E ${database.name}
```

Issue #662, opened 2023-06-13
Rust: `error: 'cargo-clippy' is not installed for the toolchain 'nightly-x86_64-apple-darwin'` but is available on the path


duckpuppy:
**Describe the bug**
When setting `language.rust.enable = true`, all the toolchain packages are available, including cargo-clippy. However, cargo does not seem to know it's there.

**To reproduce**
https://gist.github.com/duckpuppy/ac757f17ad717429860a79fe031d838b

```
$ cargo init
     Created binary (application) package

$ cargo clippy
error: 'cargo-clippy' is not installed for the toolchain 'nightly-x86_64-apple-darwin'
To install, run `rustup component add clippy`

$ cargo-clippy
    Checking test-rust-clippy v0.1.0 (/Users/paikens/src/personal/test-rust-clippy)
    Finished dev [unoptimized + debuginfo] target(s) in 0.49s
```

**Version**
 
devenv: 0.6.2

cdmistman:
hey @duckpuppy can you try again with latest main now that #667 has landed?

if it still doesn't work, can you try again with `languages.rust.channel = "stable";`?

PR #667, opened 2023-06-16
Custom rust toolchains

cdmistman:
Right now using custom `fenix` toolchains (ie, for use with `rust-toolchain.toml`) requires duplicating a `let`-bound `tools` list, which is inconvenient. Further, the rustup components that are installed is fixed, so eg `miri` can't be installed without adding it to top-level `packages`.

This change makes it easier to support both use-cases:
1. `components` is now an option and is used in place of the `let`-bound `tools`
2. `languages.rust.packages` has been renamed to `languages.rust.toolchain` to be more consistent with rustup's terminology, and now has a `freeformType` to allow usage of components that aren't in `nixpkgs`
3. `version` has been renamed to `channel` and is a bit more consistent with rustup's concept of a channel
4. the `rust-src` component is now properly overridden when defined in `components`

Here's a few examples (using `fenix` as an overlay):
```nix
# use a custom toolchain
languages.rust.toolchain = pkgs.fenix.latest;

# use a standard fenix toolchain
languages.rust.channel = "nightly";

# override a single component
languages.rust.toolchain.rustfmt = pkgs.fenix.latest.rustfmt;

# only use `rustc` and `cargo`
languages.rust.components = [ "rustc" "cargo" ];
```


cdmistman:
i also considered removing `version` and using `toolchain` instead, which would be assigned to [fenix's `toolchain` type](https://github.com/nix-community/fenix#toolchain) or a channel. This would allow for a very similar api (although would require nix-community/fenix#108):

```nix
languages.rust.toolchain = pkgs.fenix.stable;

languages.rust.toolchain = pkgs.fenix.fromManifestFile inputs.rustup-manifest;

# depends on https://github.com/nix-community/fenix/pull/108
languages.rust.toolchain = pkgs.fenix.fromToolchainName (importTOML ./rust-toolchain.toml).toolchain.channel;
```

let me know if a different API is desired and I can implement that instead :) Either way, making it easier to use custom toolchains would be awesome

src/modules/languages/rust.nix
```diff
@@ -49,13 +46,11 @@ in
...
   config = lib.mkMerge [
     (lib.mkIf cfg.enable {
-      packages = attrValues (getAttrs tools cfg.packages) ++ lib.optional pkgs.stdenv.isDarwin pkgs.libiconv;
+      packages = [ cfg.package ] ++ lib.optional pkgs.stdenv.isDarwin pkgs.libiconv;
 
       # enable compiler tooling by default to expose things like cc
       languages.c.enable = lib.mkDefault true;
 
-      env.RUST_SRC_PATH = cfg.packages.rust-src;
-
```
cdmistman, line 58: i didn't know what the best way to set this was, but I think this is fine anyways? `cargo build` still works and i verified `rustc --print sysroot` returns a satisfactory directory

domenkozar:
Have you seen https://github.com/cachix/devenv/pull/640

cdmistman:
i haven't! i'll note that the current implementation of that PR recreates a `rust-toolchain.toml` file from the rust options, which means
1. repos already using a `rust-toolchain.toml` file must recreate the toolchain spec in the devenv config
2. the recreated toolchain then results in an Import From Derivation

My PR addresses both; you may use a pre-existing `rust-toolchain.toml` file directly or avoid an IFD by adding the manifest file to the flake inputs.

If @nothingnesses would accept contributions to their PR I'd be glad to merge our work :)

cdmistman:
additionally, that PR cuts out rust components from `nixpkgs` - if I'd known this was I would've done a couple things differently 😅 namely, `rust.toolchain` could be a fenix toolchain type

nothingnesses:
Hi. I'd be happy to work together on this issue, but am not sure if there's anything much I can contribute further. I already like your solution better than mine and agree that recreating a "rust-toolchain.toml" file is undesirable and would prefer a way to just use an existing one (I couldn't figure out a way to do it without having to make a TOML parser). I think your solution would even allow using a different overlay right?

One thing I think I'd change is the default components. I think it'd be better just to stick with the ones included in the default rustup profile, which I've done in my PR. To handle the pre-commit hooks, I just set the components to be the ones from nixpkgs, if they weren't already included in the toolchain options/part of the selected profile; I think you should be able to do something similar to that with you solution.

nothingnesses:
Hi. I'd be happy to work together on this issue, but am not sure if there's anything much I can contribute further. I already like your solution better than mine and agree that recreating a "rust-toolchain.toml" file is undesirable and would prefer a way to just use an existing one (I couldn't figure out a way to do it without having to make a TOML parser). I think your solution would even allow using a different overlay right?

One thing I think I'd change is the default components. I think it'd be better just to stick with the ones included in the default rustup profile, which I've done in my PR. To handle the pre-commit hooks, I just set the components to be the ones from nixpkgs, if they weren't already included in the toolchain options/part of the selected profile; I think you should be able to do something similar to that with you solution.

nothingnesses:
Hi. I'd be happy to work together on this issue, but am not sure if there's anything much I can contribute further. I already like your solution better than mine and agree that recreating a "rust-toolchain.toml" file is undesirable and would prefer a way to just use an existing one (I couldn't figure out a way to do it without having to make a TOML parser). I think your solution would even allow using a different overlay right?

One thing I think I'd change is the default components. I think it'd be better just to stick with the ones included in the default rustup profile, which I've done in my PR. To handle the pre-commit hooks, I just set the components to be the ones from nixpkgs, if they weren't already included in the toolchain options/part of the selected profile; I think you should be able to do something similar to that with you solution.

cdmistman:
> I couldn't figure out a way to do it without having to make a TOML parser

`nixpkgs.lib.importTOML` works and there's also `builtins.fromTOML` :)

> I think your solution would even allow using a different overlay, right?

ah, i guess that's true! didn't think about that - yeah, https://github.com/oxalica/rust-overlay should work with this design as well

> I think it'd be better just to stick with the ones included in the default rustup profile, which I've done in my PR

Hm, yeah that should work. I guess I'd like to know if a goal of devenv is to handle language server installation? @domenkozar if you can weigh in here that'd be appreciated. Additionally, I'd like to know if you have any input on the trade-off between only accepting fenix's [`toolchain` type](https://github.com/nix-community/fenix#toolchain) (which would result in a much-friendlier API here) or if I should keep the API flexible with the single `options.package` as defined in my PR

> To handle the pre-commit hooks, I just set the components to be the ones from nixpkgs, if they weren't already included in the toolchain options/part of the selected profile; I think you should be able to do something similar to that with you solution.

yes I could, but I think it's better to use the binary from the `packages` - this has 2 benefits:
1. the components aren't built 2x (in the event of no substituters)
2. if the user is using nightly features and is using the nightly toolchain, those will work OOTB with pre-commit-hooks.

I guess I forgot that part, I'll use the bin directory of the built `package` derivation if the bin exists there - if it doesn't, it'll be `null`

cdmistman:
> Additionally, I'd like to know if you have any input on the trade-off between only accepting fenix's [toolchain type](https://github.com/nix-community/fenix#toolchain) (which would result in a much-friendlier API here) or if I should keep the API flexible with the single options.package as defined in my PR

it seems [gleam](https://github.com/cachix/devenv/blob/4f08665cdcde7e44a40edb39bdbfe8fc0b750932/src/modules/languages/gleam.nix#L31) and [ruby](https://github.com/cachix/devenv/blob/4f08665cdcde7e44a40edb39bdbfe8fc0b750932/src/modules/languages/ruby.nix#L82-L97) both have hard-dependencies on external inputs, but nonetheless I think I made a flexible API, defining both `options.package` and `options.toolchain`.

cdmistman:
not quite yet - i tried to use this in a project but didn't have time to dig into failures. i'll have time to take care of this in a few days - i'll ping back here when it works and it's ready for review.

cdmistman:
but i'll note this is one of the more complex nix modules i've written, any input on its current state/design would be appreciated :)

src/modules/languages/rust.nix
szlend: `RUST_SRC_PATH` is necessary when `rust-src` is not at the same location as as `rustc`. This is the case with the rust toolchain from `nixpkgs`. You'll run into this issue when using rust-analyzer in vscode.

domenkozar:
Maybe we should add that as a comment?

cdmistman:
> This is the case with the rust toolchain from nixpkgs. You'll run into this issue when using rust-analyzer in vscode.

I'm really tempted to just implement using `fenix`. Thoughts? Basically:
- a few other languages in devenv don't support using their nixpkgs equivalent unless explicitly used
- nixpkgs doesn't have all of rust's components
- it's easier to implement using just fenix

src/modules/languages/rust.nix
cdmistman: i thought this would work since i'm using `symlinkJoin`? the symlinks are in the same location, does `rustc` dereference them?

szlend: Unfortunately not. It's based on the physical location of `librustc_driver`. See: https://github.com/rust-lang/rust/pull/103660#issuecomment-1307721558

cdmistman: ah, quite unfortunate :/

cdmistman:
@domenkozar it's ready for review! happy to answer any questions or address any feedback

when i tried to remove the reliance on `nixpkgs`, `nix flake check` failed due to the missing fenix input. this is fine though - everything I've tested still works.

I probably haven't tested *all* the edge cases, but I've tested with the following configurations:
- `{ enable = true; }`
- `{ enable = true; version = "latest"; }`
- `{ enable = true; toolchain = <some toolchain>; }`
- `{ enable = true; components = [ <copy-pasted> "miri" ]; version = "latest"; }`
- `{ enable = true; components = [ <copy-pasted> "miri" ]; toolchain = <some toolchain>; }`

in a new cargo project, i tested for the following:
- rust-analyzer in vscode shows documentation for `println!`
- `cargo clean && cargo check` works
- pre-commit hooks work
- `cargo miri` works if appropriate

szlend:
Tbh I find the API as it is right now a little confusing. There's `package`, `toolchain`, `components`, `version` and it's not clear how they interact with each other - if at all. And then there's `rust-src` which is technically a component, but is in the root of config options. I don't really have a well thought out proposal right now but ideally it would be nice if the incompatible options could exclude each other somehow. Maybe it would be possible to have something like:

```
toolchain = pkgs.fenix.stable.toolchain;
toolchain = pkgs.fenix.nightly.withComponents config.languages.rust.components;
toolchain = pkgs.fenix.fromToolchainFile ./rust-toolchain.toml;
toolchain = pkgs.fenix.fromManifestFile inputs.rustup-manifest;
toolchain = { version = "stable"; }
toolchain = { version = "stable"; components = [...] }
```

I think it may be possible with `coerceTo` from attrset to package, but I'd need to think about it some more.

szlend:
Alternatively:

```
# just set the complete toolchain package
package = pkgs.fenix.stable.toolchain;
package = pkgs.fenix.nightly.withComponents [...];
package = pkgs.fenix.fromToolchainFile ./rust-toolchain.toml;
package = pkgs.fenix.fromManifestFile inputs.rustup-manifest;

# or have devenv generate `package` for you from `version` and `components`.
version = "stable";
components = [...];
```

cdmistman:
hm, maybe i can get rid of `package` by making nixpkgs quack like fenix. one sec...

szlend:
What do we really gain from having the `toolchain` option anyway? I read through the comments and I still don't understand since all we really do is call `withComponents` on it, so it should be equivalent to `version` + `components`.

Btw I just remembered, one thing this change makes more difficult is taking the stable toolchain and replacing rustfmt with the nightly version. That's something a lot of projects do since rustfmt lags a lot at formatting newly stabilized features.

src/modules/languages/rust.nix
```diff
@@ -18,62 +15,94 @@ in
...
+
+    toolchain = lib.mkOption {
+      # TODO: better type
+      type = lib.types.nullOr (lib.types.attrsOf lib.types.anything);
+      default = null;
+      defaultText = lib.literalExpression "fenix.packages.stable";
+      description = "The [fenix toolchain](https://github.com/nix-community/fenix#toolchain) to use.";
     };
 
     version = lib.mkOption {
```
szlend, line 63: Nit, but I wonder if we should really rename this to `channel`. You can't really specify a version with `fenix` because it's lacking the rust version manifests. `rust-overlay` has these though if we wanted to support that.

cdmistman: tbh i haven't looked at `rust-overlay` before, but i think it can provide a better api for devenv:

```nix
# use stable, beta, or nightly toolchain
channel = "stable";

# specify profile
profile = "default";

# you can still use `package` for more flexibility in `fenix` or `rust-overlay`
package = pkgs.fenix.stable.toolchain;
```

cdmistman: i did originally have this renamed to `channel` but the names aren't the same as rustup channels, which is why i decided to keep it as `version`. i agree, it's not really fitting, but it's more accurate than `channel`. after further thinking, i do think the correct solution is to switch to `rust-overlay` which is a bit more accurate in its naming

cdmistman:
> What do we really gain from having the toolchain option anyway? I read through the comments and I still don't understand since all we really do is call withComponents on it, so it should be equivalent to version + components.

in my mind, the fact that you can easily assign a fenix toolchain to an attribute is helpful

> one thing this change makes more difficult is taking the stable toolchain and replacing rustfmt with the nightly version. That's something a lot of projects do since rustfmt lags a lot at formatting newly stabilized features.

oooh yeah, i primarily use nightly toolchain anyways so haven't thought about that use-case.

---
ok, things obviously have to change some more. to flesh this out, it's generally desired:
1. more control over where components are coming from
2. more control over which components are enabled
3. easily specify which channel of the toolchain to use

what's currently implemented in `main` does 3, partially does 1, but does not support 2 at all. it'd be ideal if there was plug-n-play support for entire toolchain swapping as well - if a user wants more control over the toolchain than just "stable" or "nightly", they should be able to specify that easily. as such, here's the design i'm proposing - @szlend lmk if you have any more input:

```nix
# duh
enable = true;

# each component has 2 values. `enable` is self-explanatory...
components.cargo.enable = true;
# overrides the location of the component that would otherwise be set by `toolchain`
components.cargo.package = pkgs.cargo;

# the rustup channel "stable", "beta", or "nightly". sets the default value for `toolchain`
channel = "stable";

# attrset of component names to derivations
toolchain = pkgs.fenix.stable;
```

cdmistman:
i guess the above can be simplified even more using `//`

```nix
# no need for `components` since you can override component locations with `//`
# however, the easiest use-case is just using the same channel for all components, so there's a quick override for that
channel = "stable";

# components is the default list of components to grab from `toolchain`
components = [ "cargo" "rustc" ];

# toolchain is the attrset pointing to the location of each component
# if you only need nightly rustfmt for example, you can do this:
toolchain = pkgs.fenix.stable // { inherit (pkgs.fenix.latest) rustfmt; };
```

cdmistman:
nice: as of the latest commit, `rust-std` works without modification for `rust-overlay` and `nixpkgs`. if the user sets `toolchain = <some fenix toolchain>`, then there's an error due to the directory layout of fenix's `rust-std` output. i think `rust-overlay` is the right move here.

cdmistman:
nevermind, i didn't check my work enough and it seems that the toolchain derived from nixpkgs is being used even when `channel` is set... my nix knowledge is officially failing me, if somebody knows how to fix this I'd appreciate it, otherwise I'm not sure what to do now...

cdmistman:
ah, i see - rust-overlay only provides the toolchain attrsets when actually used as an overlay. switched back to fenix, again ready for review! hopefully this much simpler api is less confusing :)

nothingnesses:
@cdmistman It'd be nice if there was a way to add target triples as well. I originally opened my PR as I wanted a way to add and compile a project to the "wasm32-unknown-unknown" triple.

szlend:
> ```nix
> channel = "stable";
> 
> # components is the default list of components to grab from `toolchain`
> components = [ "cargo" "rustc" ];
> 
> # toolchain is the attrset pointing to the location of each component
> # if you only need nightly rustfmt for example, you can do this:
> toolchain = pkgs.fenix.stable // { inherit (pkgs.fenix.latest) rustfmt; };
> ```
I like it! Though it would be nice if you could just do this:
```nix
channel = "stable";
components = [ "cargo" "rustc" ]; 
toolchain.rustfmt = fenix.nightly.rustfmt; # toolchain is generated from channel and components, we override rustfmt here
```

To make this work you'd only need to `toolchain.<component> = mkDefault toolchainFromChannel.<component>`.

szlend:
> @cdmistman It'd be nice if there was a way to add target triples as well. I originally opened my PR as I wanted a way to add and compile a project to the "wasm32-unknown-unknown" triple. Is it possible to do that with your solution?

I think you could already do this like so:
```nix
toolchain.wasm-rust-std = fenix.targets.wasm32-unknown-unknown.latest.rust-std;
```
Though maybe we could make this simpler with `targets = [ "wasm32-unknown-unknown" ]`.

src/modules/languages/rust.nix
```diff
@@ -18,62 +15,80 @@ in
...
       env.RUSTFLAGS = [ "-L framework=${config.env.DEVENV_PROFILE}/Library/Frameworks" ];
       env.RUSTDOCFLAGS = [ "-L framework=${config.env.DEVENV_PROFILE}/Library/Frameworks" ];
       env.CFLAGS = [ "-iframework ${config.env.DEVENV_PROFILE}/Library/Frameworks" ];
     })
-    (lib.mkIf (cfg.version != null) (
+    (lib.mkIf (cfg.channel != null) (
       let
-        fenix = inputs.fenix or (throw "To use languages.rust.version, you need to add the following to your devenv.yaml:\n\n${setup}");
-        rustPackages = fenix.packages.${pkgs.stdenv.system}.${cfg.version} or (throw "languages.rust.version is set to ${cfg.version}, but should be one of: stable, beta or latest.");
+        error = "To use languages.rust.version, you need to add the following to your devenv.yaml:\n\n${setup}";
```
szlend, line 83: Error message needs to be updated and also applies to other options now.

src/modules/languages/rust.nix
```diff
@@ -18,62 +15,80 @@ in
...
+      description = ''
+        List of [Rustup components](https://rust-lang.github.io/rustup/concepts/components.html)
+        to install. Defaults to those available in `nixpkgs`.
+      '';
     };
 
-    version = lib.mkOption {
-      type = lib.types.nullOr lib.types.str;
+    channel = lib.mkOption {
+      type = lib.types.enum [ null "stable" "beta" "nightly" ];
```
szlend, line 29: What if we call the `null` channel `"nixpkgs"`?

cdmistman:
> I guess we could add something like `targets = [ "wasm32-unknown-unknown" ]`.

hm, i'm not sure i know how to do additional targets. the fenix docs say to combine with the `<target>.rust-std` but:
```sh
$ nix repl
nix-repl> :lf fenix
Added 14 variables.

nix-repl> :b outputs.packages.aarch64-darwin.targets.wasm32-wasi.stable.rust-src

This derivation produced the following outputs:
  out -> /nix/store/q8856rcs522pbzjxswa0cgk4kwpfbd5d-rust-src-stable-2023-06-01

nix-repl> :b outputs.packages.aarch64-darwin.stable.rust-src

This derivation produced the following outputs:
  out -> /nix/store/q8856rcs522pbzjxswa0cgk4kwpfbd5d-rust-src-stable-2023-06-01
```

i guess the only changes necessary to get this working are to get the nixpkgs for the target system and set `CARGO_TARGET_<triple>_LINKER` env var? i can dig in later but if somebody more knowledgeable than me can provide additional input that'd be appreciated.

src/modules/languages/rust.nix
```diff
@@ -13,67 +10,91 @@ let
...
+      pre-commit.tools.rustfmt = cfg.toolchain.rustfmt or null;
+      pre-commit.tools.clippy = cfg.toolchain.clippy or null;
     })
     (lib.mkIf (cfg.enable && pkgs.stdenv.isDarwin) {
       env.RUSTFLAGS = [ "-L framework=${config.env.DEVENV_PROFILE}/Library/Frameworks" ];
       env.RUSTDOCFLAGS = [ "-L framework=${config.env.DEVENV_PROFILE}/Library/Frameworks" ];
       env.CFLAGS = [ "-iframework ${config.env.DEVENV_PROFILE}/Library/Frameworks" ];
     })
-    (lib.mkIf (cfg.version != null) (
+    (lib.mkIf (cfg.channel != null) (
```
szlend, line 83: `cfg.channel != "nixpkgs"`

src/modules/languages/rust.nix
```diff
@@ -13,67 +10,91 @@ let
...
       # enable compiler tooling by default to expose things like cc
       languages.c.enable = lib.mkDefault true;
 
-      env.RUST_SRC_PATH = cfg.packages.rust-src;
+      # RUST_SRC_PATH is necessary when rust-src is not at the same location as
+      # as rustc. This is the case with the rust toolchain from nixpkgs.
+      env.RUST_SRC_PATH =
+        if cfg.toolchain ? rust-src
+        then "${cfg.toolchain.rust-src}/lib/rustlib/src/rust/library"
+        else pkgs.rustPlatform.rustLibSrc;
```
szlend, line 72: If it makes things easier, we only really need this for "nixpkgs".

src/modules/languages/rust.nix
```diff
@@ -13,67 +10,91 @@ let
...
-    packages = lib.mkOption {
+    components = lib.mkOption {
+      type = lib.types.listOf lib.types.str;
+      default = [ "rustc" "cargo" "clippy" "rustfmt" "rust-analyzer" ];
+      defaultText = lib.literalExpression ''[ "rustc" "cargo" "clippy" "rustfmt" "rust-analyzer" ]'';
+      description = ''
+        List of [Rustup components](https://rust-lang.github.io/rustup/concepts/components.html)
+        to install. Defaults to those available in `nixpkgs`.
+      '';
+    };
```
szlend, line 28: Hm, what's supposed to happen here if I set `components = [ "rustc" "cargo" ]`? Is this supposed to work like components or additional components? Right now it works like additional components. There's no way to omit e.g. "rust-analyzer". Not saying we need to support this, but at the very least it should be documented as such.

szlend: Disregard this, made a config type when testing.

src/modules/languages/rust.nix
cdmistman: ACK, haven't tested the latest changes yet 😅

src/modules/languages/rust.nix
cdmistman: > we only really need this for "nixpkgs"

i don't think that's true:
```nix
enable = true;
toolchain.rust-std = my-fenix-toolchain.rust-std;
```

cdmistman:
yeah, i'll leave extra targets as a non-blocker for this. if the user wants to accomplish this, they might be able to do so by either `combine`ing the `rust-std` components in the `toolchain` configuration or adding it to `packages`, and then setting the appropriate linker env var

src/modules/languages/rust.nix
szlend: Actually I just checked the docs so I may be wrong on this. Mixed it up with rust-overlay. I would leave it as is.

martinjlowm:
> Try with `rust-std`, not `rust-src`.
> 
> Either way though I feel like supporting cross-compilation is a bit much for this PR. I don't have experience with WASM, so maybe that's simpler. But getting other targets to work is far move involved than just installing the target `rust-std`. For example if you want to target x86_64-linux from aarch64-darwin, you'll need to build a cross-compiling stdenv from source, as nixpkgs doesn't have it in its binary cache.
> 
> So if someone with more experience can chip in to make a limited set of targets working, that would be great. But I don't think it should be a blocker.

I was actually hoping for this. :( We're cross compiling against aarch64-unknown-linux-gnu on x86_64-unknown-linux-gnu in CI but for the system platform on the developer's machine. I'm not too familiar with Nix, but perhaps there's a better way?

src/modules/languages/rust.nix
cdmistman: for clarity, when you say "leave it as is" you mean my current if-expression is fine?

szlend: It's `rust-src` in this instance. But yeah, I may be wrong here, sorry.

cdmistman:
> We're cross compiling against aarch64-unknown-linux-gnu on x86_64-unknown-linux-gnu in CI but for the system platform on the developer's machine

sounds like you want to make it an output in a flake. look at [fenix docs](https://github.com/nix-community/fenix) for an example of how to accomplish this.

szlend:
Like @cdmistman said, something like this might work, but I can't test cause I'd need to bootstrap gcc from source:
```nix
packages = [ <fenix>.targets.aarch64-unknown-linux-gnu.stable.rust-std ];
env.CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER =
  let
    inherit (pkgs.pkgsCross.aarch64-multiplatform.stdenv) cc;
  in
  "${cc}/bin/${cc.targetPrefix}cc";
```

Though this will break as soon as you have to link with any C dependencies.

src/modules/languages/rust.nix
```diff
@@ -13,67 +10,91 @@ let
...
+
+    channel = lib.mkOption {
+      type = lib.types.enum [ "nixpkgs" "stable" "beta" "nightly" ];
```
szlend, line 31: I just realized you need to set this to a non-nixpkgs value so you can set `toolchain`, which is kind of awkward and redundant. Not sure if this is something we can solve with the current API though, as it's kind of circular.

szlend:
LGTM, great job on this!

A bit suprised something like  `components = lib.mkAfter [ "rls" ];` doesn't work. I assumed it would merge with `mkOptionDefault`, but I guess not. Though I guess the use case is pretty niche.

szlend (approved):

cdmistman:
> I assumed it would merge with `mkOptionDefault`, but I guess not

yeah, I had thought the same as well - quite disappointing that `components` isn't easily extensible :( if someone else has an idea of how to do this it'd be greatly appreciated, but that's not a rabbit hole I'm wanting to go down 😅

domenkozar:
Could you also change `examples/rust` to instruct how to use it now?

cdmistman:
i didn't update the documentation - afaict, the docs get auto-updated but i couldn't rebuild them on my machine (aarch64-darwin, can't find libcairo). it seems GHA might rebuild them anyways

cdmistman:
i have to update the PR description and want to do a little more testing. I'm about to get on a flight but I'll be able to do all of that late tonight/tomorrow morning PST (depending on how i'm feeling tonight)

cdmistman:
updated the PR description and hopefully fixed failing CI (for rust - not going to fix `python-poetry` as it's out of scope) :)

domenkozar:
Could you also add rename hints for all the changed options?

```
lib.mkRenamedOptionModule [ "languages" "rust" "version" ] [ "languages" "rust" "channel" ])
```

domenkozar:
Great work, thank you for the patience with reviews :)

domenkozar:
I wonder if it also fixes https://github.com/cachix/devenv/issues/662 by any chance?

jhartma:
Just leaving this here for anybody who wants to get devenv working with rust wasm.

in ```devenv.yaml```
```yaml
inputs:
  nixpkgs:
  fenix:
    url: github:nix-community/fenix
    inputs:
      nixpkgs:
        follows: nixpkgs
  rust-overlay:
    url: github:oxalica/rust-overlay
    overlays:
      - default
   ```

and in ```devenv.nix``` add:
```nix
  ...
  languages = {
    nix.enable = true;
    rust = {
      enable = true;
      toolchain.rustc = (rustVersion.override {
        extensions = [ "rust-src" ];
        targets = [ "wasm32-unknown-unknown" ];
      });
    };
  };
  ...
```

martinjlowm:
Or for anyone using Flakes, Fenix and a toolchain file:

```nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
    fenix.url = "github:nix-community/fenix";
  };

  outputs = { self, nixpkgs, devenv, systems, fenix, ... } @ inputs:

    let
      system = builtins.currentSystem;
      fenixpkgs = fenix.packages.${system};
      aarch64-binutils = pkgs.pkgsCross.aarch64-multiplatform.stdenv.cc;
      x86_64-binutils = pkgs.pkgsCross.gnu64.stdenv.cc;
    in
      {
        devShell.${system} = devenv.lib.mkShell {
          inherit inputs pkgs;
          modules = [
            {
              packages = [
                pkgs.libunwind
                aarch64-binutils
              ] ++ pkgs.lib.optionals pkgs.stdenv.isDarwin (with pkgs.darwin.apple_sdk; [
                frameworks.Security
                frameworks.CoreFoundation
                x86_64-binutils
              ]);

              languages.rust = {
                enable = true;
                toolchain.rustc = fenixpkgs.fromToolchainFile { dir = ./.; };
              };
            }
          ];
        };
      };
}

```

`rust-toolchain.toml`:
```toml
[toolchain]
channel = "1.70.0"
targets = ["x86_64-unknown-linux-gnu", "aarch64-unknown-linux-gnu"]
```

domenkozar:
Maybe we could allow setting `languages.rust.targets` and wire everything up?

szlend:
> Maybe we could allow setting `languages.rust.targets` and wire everything up?

We could, but most targets wont work without a cross stdenv so that's a lot to wire up correctly and most of these toolchains need to be bootstrapped from source which can be hours of compiling, depending on your system. Though you can get pretty far with just cargo zigbuild: https://github.com/rust-cross/cargo-zigbuild

domenkozar:
That's exactly why we should turn it into a flag so that it's not needed to go though that pain.

PR #668, opened 2023-06-17
move process implementations to separate modules

bobvanderlinden:
As [discussed on Discord](https://discord.com/channels/1036369714731036712/1036369758024650792/1119268576206016576) this is a proposal implementation for splitting up the process-managers into separate modules.

I must say that this became a bit bigger than I initially anticipated, but it seemed like the refactorings were still useful and backwards-compatible.

There are a few todo's in here to propose further changes, but first wanted to get feedback on this proposal as-is.

While working on this I also felt it might be a good idea to move generating the Procfile and env-file into a `pkgs.formats.procfile` and `pkgs.formats.envfile`. That way these implementations can be reused, but do not need to reside as options (process-compose doesn't use them for instance, as do a number of other non-procfile-based process managers). Not really needed to worry about right now though.

### DIFF
--- a/src/modules/process-managers/hivemind.nix
+++ b/src/modules/process-managers/hivemind.nix
@@ -1,0 +1,22 @@
+{ pkgs, config, lib, ... }:
+let
+  cfg = config.process-managers.hivemind;
+in
+{
+  options.process-managers.hivemind = {
+    enable = lib.mkEnableOption "hivemind as process-manager";
+    package = lib.mkOption {
+      type = lib.types.package;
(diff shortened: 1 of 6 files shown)


jlesquembre:
@bobvanderlinden We need to keep the `port` and `tui` options in the CLI for process-compose:
https://github.com/cachix/devenv/blob/eae002528893ba224f88b35163ff3ec273f67784/src/modules/processes.nix#L56-L60

I created PR: 
#700

PR #687, opened 2023-06-27
env: support runtime evaluation

max-privatevoid:
An attempt at getting rid of the call to `builtins.getEnv` to determine the current working directory. Taking inspiration from the implementation in https://github.com/numtide/devshell.

Things to implement:
- [ ] Replace all remaining eval-time uses of environment variables (`${config.env....}`)

Things to invent:
- [ ] Find a way to splice environment variables into config files that don't support environment variable substitution
- [ ] Find a way to make this work for containers
- [ ] Find a way to deal with ordering (for example, `AN_EXAMPLE_DATA_DIR` gets set before `DEVENV_STATE`, thus cannot reference it with `$DEVENV_STATE/example`)

### DIFF
--- a/src/modules/integrations/devcontainer.nix
+++ b/src/modules/integrations/devcontainer.nix
@@ -56,7 +56,7 @@ in
 
   config = lib.mkIf config.devcontainer.enable {
     enterShell = ''
-      cat ${file} > ${config.env.DEVENV_ROOT}/.devcontainer.json
+      cat ${file} > "$DEVENV_ROOT/.devcontainer.json"
     '';
   };
 }
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -11,7 +11,7 @@ let
           url: github:cachix/nixpkgs-python
   '');
 
-  venvPath = "${config.env.DEVENV_STATE}/venv";
+  venvPath = "$DEVENV_STATE/venv";
 
   initVenvScript = pkgs.writeShellScript "init-venv.sh" ''
     # Make sure any tools are not attempting to use the python interpreter from any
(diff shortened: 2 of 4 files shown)


domenkozar:
Nice! I'd love to see it working :)

domenkozar:
What I'm worried the most as a blocker here is putting real paths into configuration files generated by Nix.

That's really important and I don't see how it can be done, since you need to resolve the env var as bash to be able to generate the config file.

Maybe we need to use recursive Nix or IFD?

max-privatevoid:
It might be necessary to implement a custom templater, similar to what [NixPak](https://github.com/nixpak/nixpak) does with Sloth values and the NixPak launcher. Sloth values are a mini-DSL for defining "thunks" of strings that get rendered at runtime. 

For example, the Nix expression (or a string that gets parsed with Nix code and converted to this, think envsubst)
```nix
sloth.concat [
  "unix:path"
  (sloth.env "XDG_RUNTMIE_DIR")
  "/nixpak-bus"
]
```
gets turned into an intermediate JSON representation (hypothetical, the NixPak implementation outputs something slightly less efficient since I'm bad at Go :P)
```json
{
  "items": [
    "unix:path="
    {
      "key": "XDG_RUNTIME_DIR",
      "type": "env"
    }
    "/nixpak-bus"
  ],
  "type": "concat"
}
```
which the launcher turns into a string at runtime, inspecting its environment to discover that the value of the environment variable `XDG_RUNTIME_DIR` is `/run/user/1000`:
````
unix:path=/run/user/1000/nixpak-bus
````
I imagine we could have Nix generate config files in the intermediate representation and then have a wrapper binary render them into a file in `$DEVENV_STATE` at runtime, inserting the path of the rendered file, either into another file or into the command-line arguments of a process.

domenkozar:
I don't think this is going to land, because we have made the decision that env vars should be expanded at the Nix language.

PR #692, opened 2023-06-29
[feature]: Add ability for process.exec to be multiline

benaduggan:
Small teak to allow multi-line bash strings for processes.exec

### DIFF
--- a/src/modules/processes.nix
+++ b/src/modules/processes.nix
@@ -4,7 +4,7 @@ let
   processType = types.submodule ({ config, ... }: {
     options = {
       exec = lib.mkOption {
-        type = types.str;
+        type = types.lines;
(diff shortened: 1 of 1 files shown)


src/modules/processes.nix
```diff
@@ -4,7 +4,7 @@ let
   processType = types.submodule ({ config, ... }: {
     options = {
       exec = lib.mkOption {
-        type = types.str;
+        type = types.lines;
```
domenkozar, line 7: I'm not sure we should change it to lines, that means multiple definitions will get merged, resulting into weird errors.

Can't we just keep `str`?

benaduggan: sure thing!

PR #701, opened 2023-07-05
Add RubyOnRails example

domenkozar:
cc @bobvanderlinden

### DIFF
--- a/examples/rubyonrails/.test.sh
+++ b/examples/rubyonrails/.test.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+set -ex
+rails new blog -d=postgresql
+devenv up&
+timeout 20 bash -c 'until echo > /dev/tcp/localhost/5100; do sleep 0.5; done'
+(cd blog && rails db:create)
+curl -s http://localhost:5100/ | grep "version"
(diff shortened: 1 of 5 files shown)


examples/rubyonrails/.test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+set -ex
+rails new blog -d=postgresql
+devenv up&
+timeout 20 bash -c 'until echo > /dev/tcp/localhost/5100; do sleep 0.5; done'
+pushd blog
+    rails db:create
+popd
```
bobvanderlinden, line 8: CI is failing for `pushd`. I guess `sh` doesn't support it.

```suggestion
(cd blog && rails db:create)
```

domenkozar:
macOS:

> /Users/runner/work/devenv/devenv/examples/rubyonrails/.test.sh: line 5: timeout: command not found

bobvanderlinden:
Ugh, maybe the tests should get a separate nix file that is used on top of the devenv.nix in the example. Include all of the environment that is needed for the test in there (bash, curl, etc)?

domenkozar:
I've added a section that injects the packages.

Maybe we could also add helper functions for testing, but that's for later.

PR #702, opened 2023-07-06
Point to all service definitions in the docs.

dnikolovv:
As a nix noob, it took me a while to figure out what fields are available.

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -37,4 +37,4 @@ Starting processes ...
   include-markdown "services-all.md"
 %}
 
-You can find the definitions of all services [here](https://github.com/cachix/devenv/tree/main/src/modules/services).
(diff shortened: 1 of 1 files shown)


docs/services.md
```diff
@@ -35,4 +35,6 @@ Starting processes ...
 
 {%
   include-markdown "services-all.md"
-%}
\ No newline at end of file
+%}
+
+You can find the definitions of all services [here](https://github.com/cachix/devenv/tree/main/src/modules/services).
```
domenkozar, line 40:
```suggestion
You can find all supported options for services [here](https://devenv.sh/reference/options/#servicesadminerenable).
```

dnikolovv: Ooh that's even better! Thanks!

PR #581, opened 2023-05-07
rust: evaluate $DEVENV_PROFILE at runtime instead of in nix

kaiba42:
Fixes #580. This appears to be another instance of the underlying issue causing #558 and #396.

"This fixes an infinite recursion error that originated as a result of trying to set an environment variable using the output of another environment variable." - @anoadragon453

PR #705, opened 2023-07-07
rust: add tests & complete example

averagechris:
similar to #581 but takes @bobvanderlinden 's approach used in #573

adds a test to the rust language module. I'm not exactly sure how these tests are setup to work, so let me know if there's something else I need to do.


examples/rust/.test.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+set -ex
+cargo --version
+rustc --version
+
+echo "$RUSTFLAGS" | grep -- "--jobs 1"
```
bobvanderlinden, line 6: Is there a way to build a simple rust/cargo application inside the test in order to test whether these rustflags work end-to-end?

Grepping the values doesn't say much about the usecase that it is trying to support.

bobvanderlinden (commented):
The change in the rust module looks good 👍

domenkozar:
See https://github.com/cachix/devenv/pull/707 for a more general solution to this.

examples/rust/.test.sh
averagechris: Mostly rust flags is used to inform the rust compiler of other libraries to link against. I don't usually develop on macOS, but I tried adding some dependencies that use native libraries on macOS (notify crate) and then compiling the project but it seemed to work without these flags set. It could be that you have to use something specific.

In that case, perhaps these libraries shouldn't even be linked by default since many rust programs won't need them at all even if they're being developed on macOS.

averagechris:
#707 looks great. Would we still want some or all of the simple test for the languages.rust module I added as a part of this PR? If not I'll close this, otherwise I can remove the change to the env variables.

domenkozar:
> #707 looks great. Would we still want some or all of the simple test for the languages.rust module I added as a part of this PR? If not I'll close this, otherwise I can remove the change to the env variables.

Test are very welcome :)

examples/rust/.test.sh
averagechris: I added an example rust app and compile / run it in the test. It doesn't cover the linking with the macos flags though.

domenkozar:
Looks like you need to run `pre-commit run -a`.

averagechris:
When I run that there are no changes or failing checks. I see that it's running the pre-commit checks that are enabled in the rust example. I'll look into it this evening and fix it.

averagechris:
Ok so it looks like no other examples have any pre-commit.hooks enabled, and the test runner doesn't work if an example does have a pre-commit hook enabled. I'm going to disable them in the example for now to put the rust example inline with all of the other examples, if that works for ya'll :)

examples/rust/.test.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+set -ex
+cargo --version
+rustc --version
+
+if [[ "$(uname)" == "Darwin" ]] then
```
domenkozar, line 6:
```suggestion
if [[ "$(uname)" == "Darwin" ]]; then
```

PR #706, opened 2023-07-07
Let Postgres support relative $DEVENV_ROOT

Atry:
`unix_socket_directories` is by default relative to $PGDATA, therefore `unix_socket_directories=.` should work the same as `unix_socket_directories=$PGDATA` when `$PGDATA` is absolute. However, `unix_socket_directories=.` also works with relative `$PGDATA` because it does not depend on `$PGDATA`

### DIFF
--- a/src/modules/services/postgres.nix
+++ b/src/modules/services/postgres.nix
@@ -258,7 +258,7 @@ in
     services.postgres.settings = {
       listen_addresses = cfg.listen_addresses;
       port = cfg.port;
-      unix_socket_directories = lib.mkDefault config.env.PGDATA;
+      unix_socket_directories = lib.mkDefault ".";
     };
 
     processes.postgres = {


domenkozar:
Why is `realpath` needed? We need to keep `PGHOST` at the Nix level since it's used often.

domenkozar:
https://github.com/cachix/devenv/commit/5e150be6deb62bbe8c8ce91a48a3eca932708c30

Atry:
We need `realpath` because `PGHOST` does not support relative path

Atry:
> We need to keep PGHOST at the Nix level since it's used often.

I didn't find `PGHOST` being used anywhere in Nix files https://github.com/search?q=repo%3Acachix%2Fdevenv%20PGHOST&type=code

Atry:
@domenkozar What do think about this PR?

domenkozar:
We aren't going to pursue relative DEVENV_ROOT support, that ship has sailed a long time ago. It's too much complexity to maintain it.

PR #707, opened 2023-07-08
env: fix infinite recursion errors

### LINKED ISSUE #580: Infinite recursion error on macOS for Rust language and processes config
**Describe the bug**
After setting up a new project using `nix flake init --template github:cachix/devenv` after having installed `direnv` and `nix-direnv`, the devenv shell fails to initialize if Rust language support is enabled _and_ any process is defined.

```bash
direnv: loading ~/Projects/test/.envrc
error: infinite recursion encountered

       at /nix/store/z23x3la7s4ghznq0r9h1vihrmmc4gqg1-source/src/modules/languages/rust.nix:66:35:

           65|       env.RUSTDOCFLAGS = [ "-L framework=${config.env.DEVENV_PROFILE}/Library/Frameworks" ];
           66|       env.CFLAGS = [ "-iframework ${config.env.DEVENV_PROFILE}/Library/Frameworks" ];
             |                                   ^
           67|     })
realpath: /Users/kaiba/Projects/test/.direnv/flake-profile.41320: No such file or directory
direnv: nix-direnv: renewed cache
```

**To reproduce**
```
# .envrc
if ! has nix_direnv_version || ! nix_direnv_version 2.2.1; then
fi

nix_direnv_watch_file devenv.nix
nix_direnv_watch_file devenv.lock
nix_direnv_watch_file devenv.yaml
if ! use flake . --impure
then
  echo "devenv could not be build. The devenv environment was not loaded. Make the necessary changes to devenv.nix and hit enter to try again." >&2
fi
```

```nix
# flake.nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-22.11";
  };

  outputs = { self, nixpkgs, devenv, systems, ... } @ inputs:
    let
      forEachSystem = nixpkgs.lib.genAttrs (import systems);
    in
    {
      devShells = forEachSystem
        (system:
          let
          in
          {
              inherit inputs pkgs;
              modules = [
                {
                  # https://devenv.sh/reference/options/
                  packages = [ pkgs.hello ];

                  
                  # Adding these three lines will result in an infinite recursion error
                  #####################################################################
                  processes = {
                    anything.exec = "ls";
                  };
                  languages.rust.enable = true;
                  #####################################################################

                  enterShell = ''
                    hello
                  '';
                }
              ];
            };
          });
    };
}
```

**Version**
`devenv: 0.6.2`
`nix (Nix) 2.13.3`
`macOS 13.2.1 BuildVersion 22D68`

domenkozar:
The best way to solve this for now is to set these as part of config.devenv.* namespace and reference these instead of from config.env when setting env.

Also, when using flakes integration let's evaluate the whole config when invoking `devenv up` instead of when building the devenv cli.

Fixes #580

### DIFF
--- a/src/devenv-devShell.nix
+++ b/src/devenv-devShell.nix
@@ -18,7 +18,7 @@ pkgs.writeScriptBin "devenv" ''
 
   case $command in
     up)
-      procfilescript=${config.procfileScript}
+      procfilescript=$(nix build '.#devShells.${pkgs.stdenv.system}.default.config.procfileScript' --impure)
       if [ "$(cat $procfilescript|tail -n +2)" = "" ]; then
         echo "No 'processes' option defined: https://devenv.sh/processes/"
         exit 1
--- a/src/modules/containers.nix
+++ b/src/modules/containers.nix
@@ -21,8 +21,6 @@ let
   nix2container = nix2containerInput.packages.${pkgs.stdenv.system};
   mk-shell-bin = inputs.mk-shell-bin or (throw "To build the container, you need to add the following to your devenv.yaml:\n\n${setup}");
   shell = mk-shell-bin.lib.mkShellBin { drv = config.shell; nixpkgs = pkgs; };
-  # set devenv root to be at /
-  containerEnv = config.env // { DEVENV_ROOT = ""; };
   mkEntrypoint = cfg: pkgs.writeScript "entrypoint" ''
     #!${pkgs.bash}/bin/bash
 
@@ -52,7 +50,7 @@ let
       })
     ];
     config = {
-      Env = lib.mapAttrsToList (name: value: "${name}=${lib.escapeShellArg (toString value)}") containerEnv;
+      Env = lib.mapAttrsToList (name: value: "${name}=${lib.escapeShellArg (toString value)}") config.env;
       Cmd = [ cfg.startupCommand ];
       Entrypoint = cfg.entrypoint;
     };
@@ -198,6 +196,9 @@ in
     }
     (if envContainerName == "" then { } else {
       containers.${envContainerName}.isBuilding = true;
+    })
+    (lib.mkIf config.container.isBuilding {
+      devenv.root = lib.mkForce "/";
     })
   ];
 }
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -19,7 +19,7 @@ let
     unset VIRTUAL_ENV
 
     if [ ! -L ${venvPath}/devenv-profile ] \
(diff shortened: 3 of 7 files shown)


src/devenv-devShell.nix
```diff
@@ -18,7 +18,7 @@ pkgs.writeScriptBin "devenv" ''
 
   case $command in
     up)
-      procfilescript=${config.procfileScript}
+      procfilescript=$(nix build '.#devShells.${pkgs.stdenv.system}.default.config.procfileScript' --impure)
```
enobayram, line 21: Seems like a sledgehammer for fixing recursion errors. Here are some immediate issues:
* Delays the `nix build` call to the `devenv up` execution time and assumes that one does not need any additional nix flags. The `config.procfileScript` build might trigger new builds and those might require, say, a binary cache configuration that the user already passed in while running the initial `nix develop`.
* Won't work with [the Multiple shells setup](https://devenv.sh/guides/using-with-flakes/#modifying-your-flakenix-file:~:text=use%20flake%20.%20%2D%2Dimpure-,Multiple%20shells,-Defining%20multiple%20development).
* Assumes that our working directory is still in the same flake, so makes it impossible to run a devenv from a remote flake, like `nix develop github:me/myshell --impure`.
* Degrades the user experience by kicking off another round of potentially-lengthy `nix build` execution. Can be particularly bad for a workflow that relies on `nix build`ing the `devShell` somewhere else and embedding the resulting nix store hash into a script (along with a binary cache, for example).

I strongly recommend avoiding a sledgehammer like this and instead finding a way to resolve the root cause of the recursion.

domenkozar: Those are fair points, could you open an issue? To fix this properly we need to decide on the trade offs.

The benefit of this approach is that we don't get into inf. recursive errors and that `devenv up` matches closer to what it does in devenv: evaluates and build on every CLI invocation - this is important as it's an expectation going between plain devenv and flakes integration.

enobayram: Sure, I'll gather my thoughts and open an issue as soon as I get the chance.

> `devenv up` matches closer to what it does in `devenv`

As a quick note; The more I think about this issue, the more it feels to me like this mini-`devenv` script doesn't belong in the nix-flake-based use case. When one uses nix flakes with the devenv project, one essentially replaces the `devenv` utility with the `nix` family of commands. So `devenv shell` becomes `nix develop`. Then it sounds to me like `devenv up` should become `nix run .#processes`. Extrapolating from this train of thought, maybe the right interface for nix flakes isn't `devenv.lib.mkShell`, but it's `devenv.lib.mkFlake` so that the mapping of the subcommands from `devenv` to nix flakes can be expressed all at once?

PR #716, opened 2023-07-11
Minio client and extra initialization options

stephank:
This adds `mc` to the profile if minio is enabled, and configures it to connect to the local minio using the conventional `local` alias.

I added `services.minio.clientConfig = null` as an opt-out. I imagine configuring other endpoints via Nix isn't ideal, because the config contains credentials, so this provides a way out to using your regular user config.

I also added a `after` option to allow running some additional initialization following minio startup.


domenkozar:
You can test it like:

```
$ devenv shell
$ devenv-test-example minio
```

stephank:
Looks like `devenv shell` at the toplevel tries to rebuild nix itself for me, there's no cache. But I managed to write a test without it.

stephank:
It looks like minio doesn't actually generate credentials when left empty, it just uses `minioadmin:minioadmin`, so I made options reflect that, and that makes it work without configuring credentials at all.

stephank:
Failure was because of `default = defaultClientConfig;` on an option, but that value was derived from config that was not available if minio is disabled.

I now use `mkBefore`, so that you could add aliases with just `services.minio.clientConfig.aliases.foo = { ... }`, and so that `version` is also set in that case. But unlike `mkDefault`, that means simply `clientConfig = null` doesn't work anymore, and `mkForce` is required.

Issue #658, opened 2023-06-10
Get rid of impure mode


domenkozar:
https://github.com/NixOS/nix/issues/5663

Atry:
It is possible to set `DEVENV_ROOT` to support pure mode
See #566 and #583

Atry:
Why do we need plain Nix expressions as flake inputs? Why not just generate a flake that contains current environment variables, including `PWD`, then just pass `--override-input` to use the generated flake.

domenkozar:
For devenv we can do that, but not for the flakes integration.

Atry:
How about:
```
nix develop --redirect .#devenv-state ./.devenv/state
```

https://nixos.org/manual/nix/stable/command-ref/new-cli/nix3-develop.html

Atry:
I think the current impure `devenv.root` is a dangerous design when packing images via `nix2container`, because the image content could depend on `devenv.root` and reference to non-existing directory at runtime.

Atry:
Ideally, there should be a `perWorkingDirectory` submodule under `devenv`, where we can access `root`. Environment variables that contain abstract paths should be configured under `devent.perWorkingDirectory.env`, instead of `devent.env`

domenkozar:
In #745 we now generate PWD instead of using `--impure` when using `devenv` cli.

Atry:
How about also avoid `--impure` when using flake? See https://github.com/cachix/devenv/pull/745/files#r1285246995

khaled:
Looking forward to the Python rewrite, as `--impure` seems to prevent evaluation caching from occurring.  This makes devenv annoyingly slow when e.g. entering directories with a `use flake --impure` in `.envrc`.

khaled:
> Looking forward to the Python rewrite, as `--impure` seems to prevent evaluation caching from occurring. This makes devenv annoyingly slow when e.g. entering directories with a `use flake --impure` in `.envrc`.

Unless one uses [nix-direnv](https://github.com/nix-community/nix-direnv), which caches the nix shell environment.  The point about evaluation caching stands though.

Atry:
Why is this issue closed? `--impure` is still needed according to https://devenv.sh/guides/using-with-flakes/#getting-started

Atry:
See https://github.com/cachix/devenv/pull/1018

domenkozar:
This issue was about devenv itself, for flakes it needs to be fixed in Nix itself. Let's see if #1018 will work :)

PR #718, opened 2023-07-11
Let dotenv support relative `devenv.root`

Atry:
This PR avoids `lib.pathExists`, which does not support relative path

### DIFF
--- a/src/modules/integrations/dotenv.nix
+++ b/src/modules/integrations/dotenv.nix
@@ -5,7 +5,6 @@ let
 
   dotenvPath = config.devenv.root + "/" + cfg.filename;
 
-  dotenvFound = lib.pathExists dotenvPath;
   parseLine = line:
     let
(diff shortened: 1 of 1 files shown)


Atry:
This PR can solve #658

Atry:
By the way, the original implementation using `lib.pathExists` is simply buggy even when `devenv.root` is absolute, because `lib.pathExists` is evaluated only when `flake.nix` changes when the evaluation result is cached by `direnv`, while we want to test if `.env` exists each time when entering the shell.

Atry:
More broadly, using `devenv.root` at nix evaluation time is always error prone.

Atry:
Reopen this PR to trigger CI

Atry:
@domenkozar Could you trigger the CI, please?

domenkozar:
This removes `env` population at Nix level, we can't break that.

PR #721, opened 2023-07-11
feat(rust): set `$CARGO_INSTALL_ROOT` and `$PATH` used in `cargo install`

Atry:
# Motivation

When you have multiple devenv managed Rust projects, you want them to use their own `$CARGO_INSTALL_ROOT`, otherwise `cargo install` will overwrite executables that belong to other projects.

See https://doc.rust-lang.org/cargo/commands/cargo-install.html


src/modules/languages/rust.nix
```diff
@@ -63,6 +63,19 @@ in
 
   config = lib.mkMerge [
     (lib.mkIf cfg.enable {
+
+      # Set $CARGO_INSTALL_ROOT so that executables installed by `cargo install` can be found from $PATH
+      enterShell = ''
+        export CARGO_INSTALL_ROOT=$(${
+          lib.strings.escapeShellArgs [
+            "${pkgs.coreutils}/bin/realpath"
```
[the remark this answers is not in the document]
Atry: `realpath` would not throw an error when the folder is not present because there is a `--no-symlinks` flag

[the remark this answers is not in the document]
shyim: Ahhh the realpath of Darwin throws an error, the coreutils is fine 😅

Atry:
Could you review this PR again, please? @shyim

shyim (approved):

Atry:
@domenkozar Do you want to review and merge this PR?

src/modules/languages/rust.nix
```diff
@@ -63,6 +63,19 @@ in
...
+            "--no-symlinks"
+            "${config.devenv.state}/cargo-install"
```
domenkozar, line 73: Where will this install the binaries? 

I'm a bit confused, doesn't `$CARGO_INSTALL_ROOT` tell where the binaries should be installed for when you call cargo-install?

It would also be great if this added a test into `.test.sh` to assure we can run the binary.

Atry: Yes, `$CARGO_INSTALL_ROOT` tell where the binaries should be installed for when you call `cargo install`. What confuses you?

Atry: @domenkozar Tests added

domenkozar:
@Atry I tried to resolve the merge conflict, but it created an issue with formatters. Could you run `pre-commit run -a` and commit?

Atry:
Formatted

PR #726, opened 2023-07-12
feat(php): allow disabling pre-enabled php extensions

shyim:
Tailscale + PHP imap is right now evil see https://github.com/php/php-src/issues/11673

Right now it's not possible to disable "builtin" extensions. So I added a additional option to do this

### DIFF
--- a/src/modules/languages/php.nix
+++ b/src/modules/languages/php.nix
@@ -18,9 +18,11 @@ let
 
   phps = inputs.phps or (throw "To use languages.php.version, you need to add the following to your devenv.yaml:\n\n${setup}");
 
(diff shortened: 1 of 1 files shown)


domenkozar:
Maybe we should exclude it from the default extensions instead?

shyim:
You mean globally disabling `imap`? 🤔

domenkozar:
This computation can be quite expensive, but I hope noone will have a ton of disabled extensions (or will fix the perf).

shyim:
Do you have an better idea how to do this more performant? It's kinda list contains 🙈

PR #728, opened 2023-07-13
Mailpit: persist db, add sendmail binary, add example

stephank:
cc @shyim 

I understand the argument handling was based on the mailhog module, and noticed it didn't uses `escapeShellArg(s)`. Not sure if that was intended?


shyim (approved):
Looks good

shyim:
I had no reason. Thanks for the adjustments! 💙

stephank:
Typo fixed

PR #730, opened 2023-07-14
processes: eliminate shell wrapper

### LINKED ISSUE #595: When using Overmind, PHP-FPM is not stopped when services are stopped
**Describe the bug**

When using Caddy & PHP as [shown in the examples](https://github.com/cachix/devenv/tree/main/examples/caddy-php), I've noticed that PHP-FPM is not killed when exiting the `devenv up` process. This *only* happens when using `process.implementation = "overmind"` though, not any of the other process managers.

After pressing Ctrl+C to terminate the processes, the following output is printed:

```
Stopping processes...
caddy      | Interrupting...
phpfpm-web | Interrupting...
caddy      | 2023/05/15 08:14:12.906	INFO	shutting down	{"signal": "SIGINT"}
caddy      | 2023/05/15 08:14:12.906	WARN	exiting; byeee!! 👋	{"signal": "SIGINT"}
caddy      | 2023/05/15 08:14:12.906	INFO	tls.cache.maintenance	stopped background certificate maintenance	{"cache": "0xc0001f76c0"}
caddy      | 2023/05/15 08:14:12.906	INFO	admin	stopped previous server	{"address": "localhost:2019"}
caddy      | 2023/05/15 08:14:12.906	INFO	shutdown complete	{"signal": "SIGINT", "exit_code": 0}
caddy      | Killing...
phpfpm-web | Killing...
caddy      | Exited
phpfpm-web | Exited
Processes stopped.
```

So Overmind claims to have killed the process, but `ps -ef | grep php-fpm` shows it is still running.

This then blocks running `devenv up` again, as PHP-FPM will notice that another instance is already using its socket and die.

**To reproduce**

I've created a [reproduction repo](https://github.com/cumber/devenv-overmind-phpfpm). It's just the Caddy+PHP example with `process.implementation = "overmind"` added.

**Version**
 
devenv: 0.6.2

stephank:
This appears to fix #595. Overmind [creates a shell wrapper](https://github.com/DarthSim/overmind/blob/7d970570e9208812eaf3782d695bc13700bee48e/start/command.go#L166-L181), but that shell process sticks and interferes with sending signals to PHP-FPM. The `exec` here appears to otherwise work fine with other process managers (tested with caddy-php example on macos)

### DIFF
--- a/src/modules/processes.nix
+++ b/src/modules/processes.nix
@@ -120,7 +120,7 @@ in
 
     procfile =
       pkgs.writeText "procfile" (lib.concatStringsSep "\n"
-        (lib.mapAttrsToList (name: process: "${name}: ${pkgs.writeShellScript name process.exec}")
+        (lib.mapAttrsToList (name: process: "${name}: exec ${pkgs.writeShellScript name process.exec}")
           config.processes));
 
     procfileEnv =
--- a/src/modules/languages/php.nix
+++ b/src/modules/languages/php.nix
@@ -337,7 +337,7 @@ in
       processes = mapAttrs'
         (pool: poolOpts:
           nameValuePair "phpfpm-${pool}" {
-            exec = "${startScript pool poolOpts}/bin/start-phpfpm";
+            exec = startScript pool poolOpts;
           }
         )
         cfg.fpm.pools;


src/modules/processes.nix
```diff
@@ -120,7 +120,7 @@ in
 
     procfile =
       pkgs.writeText "procfile" (lib.concatStringsSep "\n"
-        (lib.mapAttrsToList (name: process: "${name}: ${pkgs.writeShellScript name process.exec}")
+        (lib.mapAttrsToList (name: process: "${name}: exec ${pkgs.writeShellScript name process.exec}")
```
shyim, line 123: I had this same idea, but does this not break this feature? https://github.com/cachix/devenv/pull/692

domenkozar: Maybe we should also do `"exec ${process.exec}"`?

domenkozar: It does not, as `writeShellScript` will generate a store path with the script

PR #735, opened 2023-07-17
python: poetry: use .venv in DEVENV_ROOT instead of DEVENV_STATE

bobvanderlinden:
Currently the poetry integration suffers from some problems regarding the `.venv` location and its symlink.

Devenv creates the following files:

* `.devenv/state/venv`: The virtual environment for Python
* `.venv` a symlink to `.devenv/state/venv`

Poetry cannot directly use `.devenv/state/venv`, as it is hardcoded to use `.venv` in the project directory. The symlink usually directs poetry nicely to the `.devenv/state/venv`. However, when poetry itself finds out an incorrect python interpreter is in the venv compared to what is in `pyproject.toml` it will delete `.venv` (the symlink) and place its own `.venv` there instead. Now there are 2 venvs: `.devenv/state/venv` and `.venv`. `.venv` is the one poetry uses and `.devenv/state/venv` is the one devenv activates.

This is an annoying problem and doesn't need to be there if we were to use the hardcoded `.venv` path in devenv as well.

I have looked for a way to indicate to poetry to use a different venv path, but [it seems that it is not and will not be possible for now](https://github.com/python-poetry/poetry/issues/1579).


bobvanderlinden:
Yes, it would be great if the issue could be reconsidered.

> Related, should we also trigger venv rebuild when `pyproject.toml` changes?

Hmm, that is possible, but then it would possibly automatically alter the poetry.lock file. Not sure if that's desired.

bobvanderlinden:
I tested this on a few local projects. It seems to be working fine. `.devenv/state/venv` is not removed, but I'm not sure whether it should.

src/modules/languages/python.nix
```diff
@@ -63,8 +55,8 @@ let
       # Avoid running "poetry install" for every shell.
       # Only run it when the "poetry.lock" file or python interpreter has changed.
       # We do this by storing the interpreter path and a hash of "poetry.lock" in venv.
-      local ACTUAL_POETRY_CHECKSUM="${cfg.package.interpreter}:$(${pkgs.nix}/bin/nix-hash --type sha256 poetry.lock):''${POETRY_INSTALL_COMMAND[@]}"
-      local POETRY_CHECKSUM_FILE="${venvPath}/poetry.lock.checksum"
+      local ACTUAL_POETRY_CHECKSUM="${cfg.package.interpreter}:$(${pkgs.nix}/bin/nix-hash --type sha256 pyproject.toml):$(${pkgs.nix}/bin/nix-hash --type sha256 poetry.lock):''${POETRY_INSTALL_COMMAND[@]}"
```
domenkozar, line 58: Won't this fail if pyproject.toml doesn't exist? Maybe that's never the case?

bobvanderlinden: The check for pyproject.toml is already here: https://github.com/bobvanderlinden/devenv/blob/fbf8406b622d214890c166676993cf56702a8931/src/modules/languages/python.nix#L80
It won't run any of the poetry functions when it isn't available.

domenkozar:
I've updated and it was picking up the old environment. We should at least issue a warning, since the only way to recover is to delete it.

bobvanderlinden:
Hmm, even if the symlink is still being used, since both devenv activates from .venv and poetry uses .venv, both will use the symlink and there shouldn't be a problem. A warning could be an option to notify the user to clean the devenv state, but functionally I think it cannot hurt?

PR #742, opened 2023-07-20
Set VENV_PATH properly

Atry:
We should switch to `.venv` since #735

### DIFF
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -16,7 +16,7 @@ let
     # existing virtual environment. For instance if devenv was started within an venv.
     unset VIRTUAL_ENV
 
-    VENV_PATH="${config.env.DEVENV_STATE}/venv"
+    VENV_PATH="${config.env.DEVENV_ROOT}/.venv"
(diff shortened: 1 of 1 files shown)


bobvanderlinden:
It wasn't my intention to change the venv path for `languages.python.venv.enable`. I changed it for poetry, because poetry isn't compatible with other paths. What was the motivation for changing this path as well?

Atry:
Why not have a unified venv path? It would be a surprising behavior when users found the Python environment changes when they enable Poetry.

bobvanderlinden:
.devenv/state is the directory to place stateful files like venv. Poetry is just limited in the sense that I wasn't able to reliably steer poetry to use the state directory. Venv itself doesn't have that problem.

With he change of poetry s directory I didn't intent to change the behavior of venv as well 😅

Do we attempt to opt for using state for stateful files or do we opt for unifying the venv directories of different python env systems? 🤷

Maybe related: if poetry suddenly does support placing its venv in a different location, do we change all venv locations back to devenv state?

Issue #696, opened 2023-07-01
Postgres with Timescaledb: Create Extension TimescaleDB failed


bangau1:
**Describe the bug**
When I setup `timescaledb` extension and initialScript to `create extension timescaledb`, the `devenv up` failed with the postgres error. The log
```
20:07:50 postgres.1 | 2023-07-01 20:07:50.836 WIB [89376] FATAL:  extension "timescaledb" must be preloaded
20:07:50 postgres.1 | 2023-07-01 20:07:50.836 WIB [89376] HINT:  Please preload the timescaledb library via shared_preload_libraries.
20:07:50 postgres.1 |   
20:07:50 postgres.1 |   This can be done by editing the config file at: /Users/agung/Works/hiveos-backend/.devenv/state/postgres/postgresql.conf
20:07:50 postgres.1 |   and adding 'timescaledb' to the list in the shared_preload_libraries config.
20:07:50 postgres.1 |           # Modify postgresql.conf:
20:07:50 postgres.1 |           shared_preload_libraries = 'timescaledb'
20:07:50 postgres.1 |   
20:07:50 postgres.1 |   Another way to do this, if not preloading other libraries, is with the command:
20:07:50 postgres.1 |           echo "shared_preload_libraries = 'timescaledb'" >> /Users/agung/Works/hiveos-backend/.devenv/state/postgres/postgresql.conf 
20:07:50 postgres.1 |   
20:07:50 postgres.1 |   (Will require a database restart.)
20:07:50 postgres.1 |   
20:07:50 postgres.1 |   If you REALLY know what you are doing and would like to load the library without preloading, you can disable this check with: 
20:07:50 postgres.1 |           SET timescaledb.allow_install_without_preload = 'on';
20:07:50 postgres.1 | 2023-07-01 20:07:50.836 WIB [89376] STATEMENT:  CREATE EXTENSION IF NOT EXISTS timescaledb;

```

**To reproduce**
My devenv.nix
```
{ pkgs, lib, ... }:

{
  # https://devenv.sh/basics/
  env.GREET = "devenv";

  # https://devenv.sh/packages/
  packages = [
    pkgs.git
    pkgs.go
  ];

  # https://devenv.sh/scripts/
  scripts.hello.exec = "echo hello from $GREET";

  enterShell = ''
    hello
    git --version
  '';

  # https://devenv.sh/languages/
  # languages.nix.enable = true;

  # https://devenv.sh/pre-commit-hooks/
[...]

PR #746, opened 2023-07-21
[FEAT] Do not use single-user mode when initializing postgres

penguincoder:
This makes the initialization process for `postgres` use the "multi-user" mode instead of "single-user" mode. This change allows for creating extensions automatically on first boot (like `timescaledb`).

[Initial log output](https://raw.githubusercontent.com/penguincoder/devenv-timescaledb-test/main/log-output.txt)
[Minimal reproduction git repo](https://github.com/penguincoder/devenv-timescaledb-test)

You should be able to clone the reproduction repo and then do a `devenv shell` then `devenv up` and it should have a bunch of TimescaleDB in the output. I haven't tested this much more extensively than that, but it does appear to be reliable.

The process more or less mimics the pattern from the Docker library, but it's not much more than using `pg_ctl` to start the server at the beginning of the process (just _after_ `initdb` has been called and the `postgresql.conf` has been written in place) and then shutting the temporary server down using `pg_ctl` again.

This should resolve cachix/devenv#696

### DIFF
--- a/src/modules/services/postgres.nix
+++ b/src/modules/services/postgres.nix
@@ -82,21 +82,21 @@ let
     set -euo pipefail
     export PATH=${postgresPkg}/bin:${pkgs.coreutils}/bin
 
-    SEED_DATABASES="false"
+    POSTGRES_RUN_INITIAL_SCRIPT="false"
     if [[ ! -d "$PGDATA" ]]; then
       initdb ${lib.concatStringsSep " " cfg.initdbArgs}
-      SEED_DATABASES="true"
+      POSTGRES_RUN_INITIAL_SCRIPT="true"
       echo
-      echo "PostgreSQL init process complete; ready for start up."
+      echo "PostgreSQL initdb process complete."
       echo
     fi
 
     # Setup config
     cp ${configFile} "$PGDATA/postgresql.conf"
 
-    if [[ "$SEED_DATABASES" = "true" ]]; then
+    if [[ "$POSTGRES_RUN_INITIAL_SCRIPT" = "true" ]]; then
       echo
-      echo "PostgreSQL recently initialized; going to seed databases."
+      echo "PostgreSQL is setting up the initial database."
       echo
       OLDPGHOST="$PGHOST"
       PGHOST="$DEVENV_STATE/$(mktemp -d "pg-init-XXXXXX")"
@@ -104,9 +104,6 @@ let
 
       function remove_tmp_pg_init_sock_dir() {
         if [[ -d "$1" ]]; then
-          echo
-          echo "Removing temporary socket directory for initialization/seeding: $1"
-          echo
           rm -rf "$1"
         fi
       }
@@ -125,6 +122,7 @@ let
       echo "PostgreSQL Database directory appears to contain a database; Skipping initialization"
       echo
     fi
+    unset POSTGRES_RUN_INITIAL_SCRIPT
   '';
   startScript = pkgs.writeShellScriptBin "start-postgres" ''
     set -euo pipefail
--- a/src/modules/services/postgres.nix
+++ b/src/modules/services/postgres.nix
@@ -31,7 +31,6 @@ let
             echo "Creating database: ${database.name}"
             echo 'create database "${database.name}";' | psql --dbname postgres
 
-
             ${lib.optionalString (database.schema != null) ''
             echo "Applying database schema on ${database.name}"
             if [ -f "${database.schema}" ]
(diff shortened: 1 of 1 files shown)


penguincoder:
There will be some minor error output for existing repositories. If the initial script performs a destructive change, though, it can/will cause downstream failures. If you have any suggestions on a reliable method of checking if postgres has had the initialization scripts run against it, I would be happy to help implement it.

domenkozar:
Thanks! I'll try this as soon as I can :heart:

penguincoder:
Testing this is kind of hard; I had to make a git repo and manually point `devenv` source to my fork/branch in `devenv.yaml` in order to try this out. I couldn't figure out how to get `devenv` to just use the nix code sitting on my disk. If there is a better way, I'd love to document how to do that, too.

thenonameguy:
Thanks for taking on this task! 

Here is [how you can turn any example into a local test](https://github.com/cachix/devenv/blob/main/.github/workflows/buildtest.yml#L93-L99).

One thing to keep in mind is for more sophisticated task runners (like process-compose) that readiness probes might evaluate to true while the temporary DB is started up. This should be avoided. Please use the `process-compose` example along with `select pg_sleep(60);` to verify.

src/modules/services/postgres.nix
```diff
@@ -85,13 +85,30 @@ let
...
     # Setup config
     cp ${configFile} "$PGDATA/postgresql.conf"
+
+    if [[ ! -f "$PGDATA/.configured" ]]; then
+      touch "$PGDATA/.configured"
+      echo
+      echo "PostgreSQL appears unconfigured; going to run initialization scripts."
+      echo
+
+      pg_ctl -D "$PGDATA" -w start
```
thenonameguy, line 102: https://github.com/docker-library/postgres/blob/master/16/alpine3.18/docker-entrypoint.sh#L263-L265

thenonameguy: This, plus changing the default listen address to `127.0.0.1` (which should be safe), alongside changing the readiness probe to use TCP/IP instead of unix socket should fix the readiness probe problem.

thenonameguy: Alternatively the unix socket directory could be set to a temporary directory with `mktemp`, if we don't want to break backward compatibility. In my experience most coworkers expect an IP to be bound, given most of the people migrate from docker-compose postgres to devenv.

penguincoder: I don't really want to break backwards compatibility. I made the change to listen on "no addresses", I didn't have to modify the default listener (although I generally agree with the expectation of listening on TCP) to make sure that nothing was broken.

I now create a temporary directory using `mktemp` to hold the temporary initialization server socket. It removes the directory when finished and a `trap` removes it when the script `EXIT`s.

I also removed the need for the "trailer file" to tell if the database has been configured. Instead, I use a variable and set it when you actually run `initdb`. I like this better because it would have tried to run initialization scripts on all existing databases. This could be good or bad, depending on the scripts.

Side bar: if you have several minutes worth of initialization scripts, `process-compose` will forcibly kill the `postgres` process and you'll be stuck with an incomplete database. It's not something I am going to tackle right now, but be aware that those conditions exist.

penguincoder:
I updated the test repo to use `process-compose` and it works very well. I used `pg_sleep(15)` instead of 60, because after not-quite-one-minute PC will kill and restart the `postgres` process. (See above comment about initialization scripts)

penguincoder:
The comment about local testing was 💯 spot on. Thanks, it made testing quite a bit easier.

src/modules/services/postgres.nix
```diff
@@ -83,15 +82,49 @@ let
     set -euo pipefail
     export PATH=${postgresPkg}/bin:${pkgs.coreutils}/bin
 
+    SEED_DATABASES="false"
```
domenkozar, line 85: This should be named something like `POSTGRES_RUN_INITIAL_SCRIPT`

src/modules/services/postgres.nix
```diff
@@ -83,15 +82,49 @@ let
...
     if [[ ! -d "$PGDATA" ]]; then
       initdb ${lib.concatStringsSep " " cfg.initdbArgs}
-      ${setupInitialDatabases}
-
-      ${runInitialScript}
+      SEED_DATABASES="true"
+      echo
+      echo "PostgreSQL init process complete; ready for start up."
```
domenkozar, line 90:
```suggestion
      echo "PostgreSQL initdb process complete."
```

src/modules/services/postgres.nix
```diff
@@ -83,15 +82,49 @@ let
...
+      echo "PostgreSQL recently initialized; going to seed databases."
+      echo
+      OLDPGHOST="$PGHOST"
+      PGHOST="$DEVENV_STATE/$(mktemp -d "pg-init-XXXXXX")"
+      mkdir -p "$PGHOST"
+
+      function remove_tmp_pg_init_sock_dir() {
+        if [[ -d "$1" ]]; then
+          echo
+          echo "Removing temporary socket directory for initialization/seeding: $1"
```
domenkozar, line 108: I don't think this echo is needed as it's an implementation detail.

domenkozar (changes requested):

thenonameguy (approved):

thenonameguy (approved):

PR #753, opened 2023-07-26
Fix mkcert Integration When nss Is Used

dairyisscary:
Hi everyone. Just want to begin with a huge thanks for devenv. I'm a new user
to nix in general, and devenv has been a _really_ nice to work with.

I did run into an issue with the `mkcert` integration when using Firefox. I think
this PR should fix the problem:
  * the `certutil` binary is in the `nssTools` package
  * Also includes the `certutil` binary in `PATH` during `mkcert -install` since that 
    appears to be when its actually required -- to trust the new certificate
    authority. I kept it in both lines in case I'm wrong about that or in the future
    `mkcert` changes that.


jochenmanz (approved):
Good catch, looks correct :)

PR #762, opened 2023-07-28
javascript: add npmInstall option

jwygoda:
npmInstall option runs `npm install` on package.json change.


domenkozar:
Let's call it languages.javascript.npm.install, as we might have more options in the future.

PR #769, opened 2023-08-03
Add `lua-language-server` into `languages.lua`

DrSensor:


domenkozar:
Will that pick up the right lua version from `cfg.package`?

DrSensor:
Yes as long as it's Lua 5.1 or above.

Should I make a guard check to include the langserver only if `cfg.package.version >= 5.1 || cfg.package.luaversion >= 5.1`?

> **Note**
> `pkgs.luajit` has attribute `luaversion` in it but not with `pkgs.lua`

domenkozar:
Yeah, that sounds good.

PR #778, opened 2023-08-07
Add dotenv tests

### LINKED ISSUE #772: dotenv integration captures quotes in value/improperly stores json
**Describe the bug**

When using quoted json in `.env` the resulting environmental variable includes both the quotes and json. If something like "read" in Haskell is called on that environmental variable it will predictably fail.

The reason is because of [this regex in the dotenv integration](https://github.com/theskumar/python-dotenv/issues/35):

```
(.+) *= *(.+)
```

Which results in the single quotes being in the resulting value:

```
nix-repl> builtins.match "(.+) *= *(.+)" "a='{\"a\":\"b\"}'"
builtins.match "(.+) *= *(.+)" "a='{\"a\":\"b\"}'"
[ "a" "'{\"a\":\"b\"}'" ]
```

I can get the expected behavior by optionally matching `'`, but this might not be how you want to write it:

```
nix-repl> builtins.match "(.+) *= *'?(.+)?'" "a='{\"a\":\"b\"}'"
builtins.match "(.+) *= *'?(.+)?'" "a='{\"a\":\"b\"}'"
[ "a" "{\"a\":\"b\"}" ]
```

Some related issues:

https://github.com/theskumar/python-dotenv/issues/35
https://github.com/theskumar/python-dotenv/issues/285
https://github.com/theskumar/python-dotenv/issues/381

**To reproduce**
Please provide an [Short, Self Contained, Correct (Compilable), Example](http://sscce.org) by [creating a gist](https://gist.github.com/) using `devenv.nix`, `devenv.yaml`, and optionally `devenv.lock`.

Given `.env`:

```
data='{"a":"b"}'
```

Running `echo $data` with dotenv integration enabled results in:

```
$ echo $data
'{"a":"b"}'
```

I would expect only the json value and not the surrounding single quotes to be captured.

**Version**
 
Paste the output of `$ devenv version` here.

ParetoOptimalDev:
It's failing as expected when run with `nix eval .#tests`.

This will also soon fix #772

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -152,5 +152,14 @@
             inherit config;
           };
       };
+
+      # run tests with `nix eval .#tests`
+      tests =
+        let
+          results = import ./tests/dotenv/tests.nix { inherit nixpkgs; };
+        in
+        if results == [ ]
+        then "all tests passed"
+        else throw (builtins.toJSON results);
     };
 }
--- a/src/modules/integrations/dotenv-functions.nix
+++ b/src/modules/integrations/dotenv-functions.nix
@@ -1,0 +1,13 @@
+{ lib }:
+rec {
+  parseLine = line:
+    let
+      parts = builtins.match "(.+) *= *(.+)" line;
+    in
+    if (!builtins.isNull parts) && (builtins.length parts) == 2 then
+      { name = builtins.elemAt parts 0; value = builtins.elemAt parts 1; }
+    else
+      null;
+
+  parseEnvFile = content: builtins.listToAttrs (lib.filter (x: !builtins.isNull x) (map parseLine (lib.splitString "\n" content)));
+}
--- a/src/modules/integrations/dotenv.nix
+++ b/src/modules/integrations/dotenv.nix
@@ -6,16 +6,9 @@ let
   dotenvPath = config.devenv.root + "/" + cfg.filename;
 
   dotenvFound = lib.pathExists dotenvPath;
-  parseLine = line:
-    let
-      parts = builtins.match "(.+) *= *(.+)" line;
-    in
-    if (!builtins.isNull parts) && (builtins.length parts) == 2 then
-      { name = builtins.elemAt parts 0; value = builtins.elemAt parts 1; }
-    else
-      null;
-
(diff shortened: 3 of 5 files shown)


ParetoOptimalDev:
@domenkozar This is very prelimnary, but I just wanted to get something up and working. 

I'm guessing you want to structure things like `dotenv-functions` differently, perhaps even having the test somehow use `modules/integrations/dotenv.nix` rather than introduction `modules/integrations/dotenv-functions.nix`.

Or perhaps it's a positive to have a module not requiring imports that is more pure?

Also obviously need to wire this into github actions, not sure if you had plans beyond just calling `nix eval .#tests` for that.

domenkozar:
Looks good, we can refactor once the tests pass :fire:

ParetoOptimalDev:
I'm having problems getting the regex to optionally match surrounding single quotes. I'm pretty close though with:

https://regex101.com/r/ch9iSw/1

```
^'?(.*?)'? *= *'?(.*?)'?
```

Here is the current one for reference that doesn't handle quotes in regex101: https://regex101.com/r/qgtFkK/1

ParetoOptimalDev:
> I think the issue is with the optional space matching, followed by optional quote, followed by lazy matching any character. The pattern works on the first one because it knows the beginning of the line I believe.

queue thought "what if we mark the end of the line!"

Yep, that works: https://regex101.com/r/Tc8L3x/1

ParetoOptimalDev:
Nevermind, nix and this regex test differ in result: 

https://regex101.com/r/DOlw5S/1

```
nix-repl> builtins.match "'?(.*?)'? *= *'?([^'\n]*)" "a='{\"a\":\"b\"}'"
builtins.match "'?(.*?)'? *= *'?([^'\n]*)" "a='{\"a\":\"b\"}'"
null
```

ParetoOptimalDev:
- getting the regex to work for the simple cases so far
- finish tests, likely needing to tune regex along the way
- this to become higher priority for me over other tasks

This was high priority for me while I thought I could transition to it at work from docker-compose.

However I got stuck on #782, created a reproduction, and ran out of time and had to revert to using docker-compose again.

domenkozar:
Note that I gave up on dotenv, essentially it's used for secrets and we now have https://devenv.sh/blog/2025/07/21/announcing-secretspec-declarative-secrets-management/

PR #788, opened 2023-08-14
Init CockroachDB Service

ankhers:


domenkozar:
Could you also add support for macos to cockroachdb-bin in nixpkgs?

ankhers:
I can do that, but I may only be able to get to it tomorrow.

ankhers:
NixOS/nixpkgs#250442

PR #791, opened 2023-08-16
Add "Built with Nix" badge

totoroot:
Would you like to add the "Built with Nix" badge?

I initially tried to add the badge for the Matrix channel, but the shields.io badge somehow shows it as being inaccessible with both the room name and the alias. This is probably due to the room not being world readable (chat history visible to anyone).

Matrix channel
Matrix channel


domenkozar:
The room is public, so it must be something else.

PR #821, opened 2023-09-18
varnish: allow using module collection

shyim:
I need while developing Varnish xkey module. This is in the module collection which is already packaged, but it was not possible to configure it right now

### DIFF
--- a/examples/varnish/.test.sh
+++ b/examples/varnish/.test.sh
@@ -1,0 +1,20 @@
+#!/usr/bin/env bash
+set -ex
+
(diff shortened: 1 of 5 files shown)


shyim:
@domenkozar done :)

domenkozar:
There's a missing space in varnish.nix

domenkozar:
> /home/runner/work/devenv/devenv/examples/varnish/.test.sh: 13: [[: not found

you'll need to declare it as bash script

PR #829, opened 2023-09-22
OCaml: Add the ocp-indent package

ankhers:


domenkozar:
Shouldn't it come as a pre-commit hook?

ankhers:
Honestly, I am not sure. This is a package that a couple emacs configs suggest.

* [Doom](https://docs.doomemacs.org/v21.12/#/description/packages)
* [Spacemacs](https://develop.spacemacs.org/layers/+lang/ocaml/README.html#opam-packages)

I have only just started learning OCaml, but it at least seems important to have it installed when using emacs. I do not know how other editors work. This is not a generic formatter. That would be ocamlformat, which is already included in the language integration.

If my understanding is correct (and I could be very far off here), it gives a suggestion for whitespace while you are still typing and hit return. While I think ocamlformat will only work with fully correct syntax.

PR #832, opened 2023-09-29
perl: Add support for packages attribute.

marcusramberg:
packages can be literal CPAN package names or nixpkg names like TextCSV.

### DIFF
--- a/examples/perl/devenv.nix
+++ b/examples/perl/devenv.nix
@@ -1,0 +1,9 @@
+_:
+
+{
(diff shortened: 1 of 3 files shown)


domenkozar:
You need to remove the lock files from examples so we always test the latest version

marcusramberg:
> You need to remove the lock files from examples so we always test the latest version
Done. I see varnish has one btw, is that on purpose?

domenkozar:
Thanks a lot!

Issue #833, opened 2023-10-03
Unable to import .nix file from input


ankhers:
**Describe the bug**

According to the documentation, I should be able to setup an input and then import a specific file from that input. Something like

```yaml
inputs:
  my_proj: ...
imports:
  - my_proj/foo.nix
```

However, when I try to do that I get an error saying that `my_proj/foo.nix/devenv.nix` is not found.

**To reproduce**

You can clone https://git.sr.ht/~ankhers/bar_service to see this happening. It is trying to import the `service.nix` file from https://git.sr.ht/~ankhers/foo_service

**Version**
 
```sh
❯ devenv version
devenv 1.0 (x86_64-linux)
```

domenkozar:
It will be possible after #745 using:

```
inputs:
  root:
    url: git+file://.
    flake: false
imports:
- root/foo_service
```

domenkozar:
This works nowadays :)

PR #834, opened 2023-10-03
Allow module imports from devenv inputs

ankhers:
Previously the module import was only being checked for local files (i.e., an import prefixed by ./). This now includes an additional check for when we are importing something from an input that also ends in .nix which should import that file.


ankhers:
This will resolve #833

ankhers:
Note that I did not actually test this change. I was unsure how to go about that.

domenkozar:
Could you add a test to `examples/imports/devenv.yaml` and document it at `docs/reference/yaml-options.md`?

domenkozar:
Ping :)

ankhers:
Do you have a repo that you would like to use for the import? I think it would be preferable to be something owned by the cachix org.

ankhers:
ping

domenkozar:
Thanks - added in 4e446c65d2295d12d1503424aa352d28e7e32c7b

PR #745, opened 2023-07-21
1.0: Rewrite in Python

domenkozar:
`devenv` started as a bash wrapped in Nix to explore the design space.

As a result, there were many subtle bugs and lack of features that were hard to implement.

Previous `devenv` was 240M in closure and now it's down to 126M.

### TODO

- [x] Bump to Nix 2.17
- [x] Investigate a bug when .nix files are caches (eval-cache?)
- [x] Python native libraries support is broken
- [x] macOS intel crashes sometimes
- [ ] Fix [devenv gc](https://github.com/NixOS/nix/pull/8417)
- [ ] Fix flake.nix support
- [ ] Run `devenv ci` things in favor of `devenv test`.

### Fixes

Fixes https://github.com/cachix/devenv/issues/744
Fixes #693
Fixes #658 
Fixes #244 
Fixes #436 
Fixes #629
Fixes #731
Fixes #83 
Fixes #80 
Fixes #749 
Fixes #507 
Fixes #258
Fixes #486
Fixes #745
Fixes #599 
Fixes #607

# Try it out 🚀

Add `devenv.yaml`:

```
inputs:
  devenv:
     url: github:cachix/devenv/python-rewrite 
```

## Plain Nix

```
cachix use devenv
nix-env -if https://install.devenv.sh/python-rewrite
```

## Flakes

```
nix profile install --accept-flake-config tarball+https://install.devenv.sh/python-rewrite
```

PR #844, opened 2023-10-16
[FEAT] Conditionally link devenv profile

penguincoder:
When loading a shell, devenv checks .devenv/profile is

1. not a link
2. `readlink` reports someplace other than `${profile}`

Then will remove whatever is `.devenv/profile` and re-`ln` to the current active profile. When running a single command at a time, this is usually not a problem, but if you attempt to run two commands at the same time, then the race condition of one process removing a link while another is creating the link will emit warnings to stderr.

I tested this by pointing an example to the local repo and then on `enterShell = "set -x";` to see the checks working. The resulting shell created the link on the first attempt and then skipped the `rm`/`ln` on the subsequent invocations.

### DIFF
--- a/src/modules/top-level.nix
+++ b/src/modules/top-level.nix
@@ -190,8 +190,10 @@ in
       fi
 
       mkdir -p .devenv
-      rm -f .devenv/profile
-      ln -s ${profile} .devenv/profile
+      if [ ! -L .devenv/profile ] || [ "$(${pkgs.coreutils}/bin/readlink .devenv/profile)" != "${profile}" ]
+      then
+        ln -nsf ${profile} .devenv/profile
+      fi
     '';
 
     shell = performAssertions (


domenkozar:
Could you fix it for #745

src/modules/top-level.nix
```diff
@@ -190,8 +190,11 @@ in
       fi
 
       mkdir -p .devenv
-      rm -f .devenv/profile
-      ln -s ${profile} .devenv/profile
+      if [ ! -L .devenv/profile ] || [ "$(${pkgs.coreutils}/bin/readlink .devenv/profile)" != "${profile}" ]
+      then
+        rm -f .devenv/profile
+        ln -s ${profile} .devenv/profile
```
mmlb, line 196: Maybe instead? I was hoping it was atomic but looking at strace I see `symlinkat` followed by `reneameat` so not really. At least this way its in c and not bash?
```suggestion
        ln -nsf ${profile} .devenv/profile
```

mmlb, line 196: Maybe instead? I was hoping it was atomic but looking at strace I see `symlinkat` followed by `reneameat` so not really. At least this way its in c and not bash?
```suggestion
        ln -nsf ${profile} .devenv/profile
```

penguincoder: I can do that!

penguincoder:
Fixed for `python-rewrite` in #847

Issue #466, opened 2023-03-07
.git/info/exclude support


tuncbkose:
For projects where individuals might have different workflows, `.git/info/exclude` offers a nice alternative to `.gitignore`. So it can be nice to make devenv edits there instead of `.gitignore`. Maybe a `--local` flag for `devenv init` could provide such functionality.

theutz:
I'm curious: why is this not something that you can already achieve with devenv? Do they restrict this somehow?

tuncbkose:
I don't really understand your question, but here is what I had in mind:

It is not uncommon to see `.gitignore` as a part of a project repo, where for example some files produced during building a program might be added. Since devenv related to my local development and has nothing to do with the project itself, I'd prefer to keep any related changes locally, and not commit any reference to devenv to the repo itself. Thus, if I do `devenv init`, I would need to manually modify `.gitignore` to undo changes and add them to `.git/info/exclude`. This is perfectly doable, but it would just be slightly more convenient if I could just have a flag in `devenv init` or a global config to always to this instead.

theutz:
Oh, I see! That makes more sense to me now. :) Thanks for clarifying!

PR #849, opened 2023-10-23
feat: .git/info/exclude support in init

tuncbkose:
Adds #466 via `-gl/--gitignore_local` flag (I didn't know what else to name it).

This will only work if the current directory (as opposed to a given `target`) is a git repo, and silently fall back to `.gitignore` otherwise. I chose to do so because `devenv` currently seems to write `.gitignore` only in the current directory as well.

I'm not sure how/where to add a test, but will happily make one if someone tells me.


tuncbkose:
I gave an example in the linked issue. If that is not convincing enough, I don't have a better one

domenkozar:
closing as outdated

Issue #809, opened 2023-09-03
Add `inputsFrom` option/function to Shells


DrRuhe:
Hi,

`nixpkgs.lib.mkShell` has the ability to get all inputs from a list of derivations. This is useful to not require respecification of dependencies. 

I tried adapting the code from [`nixpkgs.lib.mkShell`](https://github.com/NixOS/nixpkgs/blob/master/pkgs/build-support/mkshell/default.nix#L17-L19) and got this function:

```nix
inputs-from = inputsFrom: builtins.concatMap (inputKind: (pkgs.lib.subtractLists inputsFrom (pkgs.lib.flatten (pkgs.lib.catAttrs inputKind inputsFrom)))) ["buildInputs" "nativeBuildInputs" "propagatedBuildInputs" "propagatedNativeBuildInputs"];
```

However it still does not work properly, so I'm putting this here in the hope that 

- devenv gains this functionality by default (either by exposing a function or directly as an option for shells)
- someone notices what I did wrong

For more details on the error I am seeing:
I have a flake using flake.parts with this devenv section:
```nix
devenv.shells.default = {
          packages =
            [
               # rust toolchain and some other tools
            ]   
            ++ inputs-from (builtins.attrValues self'.checks);
        };
```
Which results in linker errors that cc is not found, so I'm assuming the function I wrote is faulty.
The Original version using `nixpkgs.lib.mkShell` worked fine:
```nix
devShells.default = pkgs.mkShell {
          inputsFrom = builtins.attrValues self'.checks;

          nativeBuildInputs =
            [
               # rust toolchain and some other tools
            ];    
        };
```

PR #861, opened 2023-11-03
Add inputsFrom option

BurNiinTRee:
This mimics the mkShell argument with the same name. Adds all buildInputs, nativeBuildInputs, etc. to the installed packages.
Closes:  #809

### DIFF
--- a/src/modules/top-level.nix
+++ b/src/modules/top-level.nix
@@ -67,6 +67,12 @@ in
       default = [ ];
     };
 
+    inputsFrom = lib.mkOption {
+      type = types.listOf types.package;
+      description = "A list of packages whose inputs will be exposed inside the developer environment";
+      default = [ ];
+    };
+
     shell = lib.mkOption {
       type = types.package;
       internal = true;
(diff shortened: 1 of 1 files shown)


src/modules/top-level.nix
```diff
@@ -194,6 +200,21 @@ in
...
+          [ "buildInputs" ]
+          [ "nativeBuildInputs" ]
+          [ "propagatedBuildInputs" ]
+          [ "propagatedNativeBuildInputs" ]
+          [ "stdenv" "initialPath" ]
+          [ "stdenv" "defaultBuildInputs" ]
+          [ "stdenv" "defaultNativeBuildInputs" ]
+        ]);
+      in
+      lib.mkAfter allInputs;
```
BurNiinTRee, line 217: This will most likely have a lot of duplicate packages when passed multiple packages, how is this handled.
Packages added like this should appear in search paths after any explicitly added packages.

I-Want-ToBelieve:
I ran into a problem when trying to install treefmt-nix with the instructions Add config.treefmt.build.wrapper to the nativeBuildInputs, this PR looks to fix it

ref: https://github.com/numtide/treefmt-nix#flake-parts

domenkozar:
We need a test for this.

domenkozar:
I don't like the name, because inputs already refer to flake inputs.

Maybe `shellFor`?

domenkozar:
Could you rebase?

domenkozar:
See https://github.com/cachix/devenv/pull/2229

PR #870, opened 2023-11-13
[service:mysql] Option to ensureUsers with "GRANT OPTION" privilege.

noots-globally:
Support for GRANT OPTION for easier setup when your app needs to create dbs on the fly.


noots-globally:
wip: haven't built these locally.... is there a possible use case for this?

noots-globally:
probably not the best approach.

Issue #876, opened 2023-11-16
Use the defaults from NixOS


Nebucatnetzer:
When I tried to server a site with services.nginx I was surprised to discover that the MIME types weren't set.
Wouldn't it make sense to use similar defaults as with NixOS or other distros for the services so that we don't have to rebuild the config from scratch?
For the MIME types this was the required line:

```nix
include ${pkgs.mailcap}/etc/nginx/mime.types;
```

domenkozar:
Yes! Happy to merge a PR.

PR #877, opened 2023-11-20
Add defaultMimeTypes to Nginx

Nebucatnetzer:
This adds support for MIME types to Nginx.
Without it the static files don't work because they are all of type "text/plain".

Closes: #876 

I basically copied it from here: https://github.com/NixOS/nixpkgs/blob/nixos-23.05/nixos/modules/services/web-servers/nginx/default.nix#L571
Is there maybe a better way to use configs from NixOS?
This way we just duplicate a lot of the code and I don't really know what else might be useful.
I just noticed the MIME types because my project was broken without them.

### DIFF
--- a/src/modules/services/nginx.nix
+++ b/src/modules/services/nginx.nix
@@ -19,6 +19,8 @@ let
       scgi_temp_path ${config.env.DEVENV_STATE}/nginx/;
       uwsgi_temp_path ${config.env.DEVENV_STATE}/nginx/;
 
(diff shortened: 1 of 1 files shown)


shyim (approved):

Nebucatnetzer:
I recently thought about adding an additional option so that we can extend the nginx config more freely if needed.
The pattern for this I think is `extraConfig` or is it settings?
I could easily expand this PR this way.

Issue #536, opened 2023-04-18
Installation takes a long time


khaled:
**Describe the bug**

Installation of devenv is slow, AFAICT because aws-sdk-cpp and nix-2.12.0pre are built from source rather than pulled from cache.

**To Reproduce**

The steps below are run in the official nix docker container, though I've observed the same behavior on aarch64-darwin (locally), and on aarch64-linux (a VM):

```
❯ docker run --rm -it nixos/nix

bash-5.1# nix --version
nix (Nix) 2.15.0
bash-5.1# nix-env -iA cachix -f https://cachix.org/api/v1/install
installing 'cachix-1.3.3'
copying path '/nix/store/9q7vpjglay90dbmck8aari340vhiw3yg-busybox-static-aarch64-unknown-linux-musl-1.36.0' from 'https://cache.nixos.org'...
...

bash-5.1# cachix use devenv
Configured https://devenv.cachix.org binary cache in /etc/nix/nix.conf

bash-5.1# nix-env -if https://github.com/cachix/devenv/tarball/latest
installing 'devenv'
these 10 derivations will be built:
  /nix/store/9qcrqk3v6d94xwwxbxpxfxbs07nmryz1-libzip-1.9.2.drv
  /nix/store/w9g502p59yp1fn2dljmvnnhans3rq9b4-aws-sdk-cpp-1.9.294.drv
...

building '/nix/store/w9g502p59yp1fn2dljmvnnhans3rq9b4-aws-sdk-cpp-1.9.294.drv'...
unpacking sources
unpacking source archive /nix/store/26lg59zjav1irl0znfx5rx3liv47n1r7-source
source root is source
patching sources
applying patch /nix/store/frvfi397b88f6c91p371z5r2940x005p-cmake-dirs.patch
patching file cmake/AWSSDKConfig.cmake
...

building '/nix/store/00dk0qx5vjl9bywy5db593a2dfdzd66f-nix-2.12.0pre20230216_7c91803.drv'...
unpacking sources
unpacking source archive /nix/store/isia7synl0mdbcvfwav5ywnrwmwbwvxv-source
source root is source
...

```

**Version**
`devenv: 0.6.2`

domenkozar:
That's why there's a step (2) in the installation to configure devenv cachix.

khaled:
@domenkozar I did perform that step, as illustrated in the log I shared :).  The source builds of aws-sdk-cpp and nix happen regardless.

darkone23:
[...]

PR #884, opened 2023-11-25
Pin releases on cachix

sandydoo:
Can be triggered manually or via another workflow, i.e. after the build succeeds.

Related #536.

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -26,13 +26,18 @@ jobs:
         name: devenv
         authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
     - run: |
-        nix profile remove '.*' 
-        nix profile install --accept-flake-config . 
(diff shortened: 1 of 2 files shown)


sandydoo:
The nscloud-arm runner is not working. `mounting /proc: Operation not permitted`. Some sort of container issue?

sandydoo:
Disabling the nix sandbox [fixes the proc issue](https://github.com/sandydoo/nscloud-repro/actions/runs/6994794179/job/19028961161).
IIRC this used to work before. Did something change on namespace's end?

domenkozar:
> Disabling the nix sandbox [fixes the proc issue](https://github.com/sandydoo/nscloud-repro/actions/runs/6994794179/job/19028961161).
> 
> IIRC this used to work before. Did something change on namespace's end some time mid-August? @domenkozar, do you still have a contact at namespace?

We should really just use our own arm runner :)

PR #885, opened 2023-11-28
fix: redis startup on WSL2

shyim:
On WSL, the locale is not set properly, therefore, newer Redis versions don't start. 
I would set it the locale to C, to make it same on any platform and not depend on local locales.

If someone wants something different, they can still configure it


shyim:
the failing tests seems unrelated to this change 😅

PR #892, opened 2023-12-06
caddy: use mainProgram for binary

shyim:
This allows exchanging caddy with a different caddy built binary but different program name. Example frankenphp in nixpkgs

I had to update the devenv.lock as in that commit that mainProgram wasn't added in nixpkgs.

### DIFF
--- a/build_cloudflare.sh
+++ b/build_cloudflare.sh
@@ -1,3 +1,5 @@
+#!/usr/bin/env bash
+
 set -xe
 poetry install
 poetry run -- mkdocs build
--- a/devcontainer/nix-entrypoint.sh
+++ b/devcontainer/nix-entrypoint.sh
@@ -4,12 +4,15 @@ set +e
 if ! pidof nix-daemon > /dev/null 2>&1; then
     start_ok=false
     if [ "$(id -u)" = "0" ]; then
+        # shellcheck source=/dev/null
         ( . /nix/var/nix/profiles/default/etc/profile.d/nix-daemon.sh; /nix/var/nix/profiles/default/bin/nix-daemon > /tmp/nix-daemon.log 2>&1 ) &
+        # shellcheck disable=SC2181
         if [ "$?" = "0" ]; then
             start_ok=true
         fi
     elif type sudo > /dev/null 2>&1; then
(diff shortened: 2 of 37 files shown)


domenkozar:
>   https://www.shellcheck.net/wiki/SC2064 -- Use single quotes, otherwise this...
       >   https://www.shellcheck.net/wiki/SC3010 -- In POSIX sh, [[ ]] is undefined.
       >
       > In examples/mailpit/.test.sh line 6:
       > trap "pkill -P $DEVENV_PID" EXIT
       >                ^---------^ SC2064 (warning): Use single quotes, otherwise this expands now rather than when signalled.
       >
       > For more information:
       >   https://www.shellcheck.net/wiki/SC2064 -- Use single quotes, otherwise this...
       >

shyim:
I fixed the failure in mailpit. I think it was because shellcheck got updated with the lock file change

domenkozar:
There's more failures

shyim:
I unified all tests scripts now:
- fixed all shell checks errors
- use always trap to stop `devenv up`
- use same shebang for all scripts

domenkozar:
last 10 log lines:
       >   https://www.shellcheck.net/wiki/SC1091 -- Not following: /nix/var/nix/profi...
       >   https://www.shellcheck.net/wiki/SC2181 -- Check exit code directly with e.g...
       >
       > In build_cloudflare.sh line 1:
       > set -xe
       > ^-- SC2148 (error): Tips depend on target shell and yours is unknown. Add a shebang or a 'shell' directive.
       >
       > For more information:
       >   https://www.shellcheck.net/wiki/SC2148 -- Tips depend on target shell and y...
       >
       For full logs, run 'nix log /nix/store/96b5m76d34iz50bar5hzqn6k467l32a2-pre-commit-run.drv'.

shyim:
So `devenv ci` now runs without errors

shyim:
Looks the gitignore change of examples broke almost all things

shyim:
That flag fixed it. Can you trigger again the pipeline?

shyim:
I had to revert the ignoring of devenv.lock. I think we would need to adjust nix itself? to not run `git add` on it. 🙈

shyim:
So I fixed clickhouse, vault, modern-c, supported-languages. There is now something failing with ruby and rails. I have there no experience.

But more jobs are green than current main branch 😅

domenkozar:
Thanks for fixing those :heart:

PR #893, opened 2023-12-08
Ignore example devenv.lock files

mmlb:
I noticed this while running `devenv-test-example scripts`, that devenv.lock files are not being tracked in git except for varnish. It looks like that was a mistake so I dropped it and added devenv.lock to .gitignore.


sandydoo (approved):

shyim (approved):
Yes it's soo annoying 👍

PR #895, opened 2023-12-10
elasticmq: init

shyim:
[Amazon SQS](https://aws.amazon.com/sqs/) compatible solution was asked at the [SymfonyCon Brussels](https://live.symfony.com/2023-brussels-con/) for Devenv. So we quickly added this after the Talk this to Devenv :)

### DIFF
--- a/examples/elasticmq/.test.sh
+++ b/examples/elasticmq/.test.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+set -ex
+
+devenv up &
+DEVENV_PID=$!
+export DEVENV_PID
+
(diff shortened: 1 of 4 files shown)


shyim:
The test will fail for now because the option is not in master.

Should I add 
```
  devenv:
    url: path:../../src/modules
```

or leave it as it is?

shyim:
> /tmp/nix-shell.KF2gz1: line 117: exec: /home/runner/work/devenv/devenv/examples/elasticmq/.test.sh: cannot execute: Permission denied

Set now the executable bit 😅

PR #898, opened 2023-12-14
fix for bitrot in flake template; devenv up requires processes.run.exec

mcdonc:
Without the fix, the examples in the devenv tutorial won't work.


mcdonc:
Retracting this PR; I think this was a local screwup.

mcdonc:
Changed my mind; this change is OK.  The tutorial instructs you to try "devenv up" and this makes the template match the tutorial.

PR #899, opened 2023-12-14
Allow setting directory for javascript/python

domenkozar:
Needed for https://github.com/abi/screenshot-to-code


totoroot:
The changes proposed here would only set it for `poetry install` and `npm install`.

We probably want to introduce a directory option to the `python`/`javascript` module in general, so that it can also be for example be used for other modules like the soon-to-be-added `languages.python.pdm` as well.

Once #993 gets merged, I'll open up a PR for a `languages.javascript.directory` as well.

PR #903, opened 2023-12-15
Add influxdb module

soebbing:
Hi everybody!

This is my first attempt at a devenv module, to add support for InfluxDB.

I'm looking forward to some feedback!


src/modules/services/influxdb.nix
```diff
@@ -0,0 +1,22 @@
...
+{
+  options.services.influxdb = {
+    enable = lib.mkEnableOption "influxdb";
+
+    package = lib.mkOption {
+      type = types.package;
+      description = "An open-source distributed time series database";
+      default = pkgs.influxdb;
+      defaultText = lib.literalExpression "pkgs.influxdb";
+    };
```
domenkozar, line 16: Usually we'd also add `config` to pass the configuration to the daemon

soebbing: I'll try to add a config! It's just kinda hard, since I have no clue of influx nor nix 😅

soebbing: But would the pr otherwise be fine, apart from the missing config?

domenkozar: Otherwise looks good! It doesn't have to have anything in the config if it works with the default one well :)

soebbing: I added a config and tried to test for a changed setting (the modified default port) to make sure it is working.

The service seems to need some sort of complete config, not only some lonely changed setting.

I wasn't sure how where to point the directories it needs write access to (`/tmp` atm), that might need to be changed to something making more sense.

domenkozar:
Two more things:

- I'd name it `config` instead of `settings` so it's consistent with the other modules
- we don't commit the lock file for examples

soebbing:
Lock file has been removed and I changed `settings` to config (hopefully in the right place). 😅 

I took the ElasticMQ as an example https://github.com/cachix/devenv/blob/main/src/modules/services/elasticmq.nix#L18 - I think, that might need some changing as well.

domenkozar:
Seems like the test failed: 

+ influx --execute 'CREATE DATABASE devenv'
Failed to connect to http://localhost:8086: Get "http://localhost:8086/ping": dial tcp [::1]:8086: connect: connection refused
Please check your connection settings and ensure 'influxd' is running.

soebbing:
Test is fixed! I forgot to set the port for the cli-command and it worked on my machine, since another influx instance was running on that port 🤦

soebbing:
At least I thought I had fixed the test. But I am unsure how to interpret these checks. InfluxDB seems to have been successful, but others failed?

soebbing:
That seems to have done the trick! https://github.com/cachix/devenv/actions/runs/7250287690/job/19760416126

PR #905, opened 2023-12-19
normalize paths when root is /

mcdonc:
When devenv.root is "/", devenv.state becomes "//.devenv/state," and similar with devenv.dotfile.  Avoid this by normalizing using builtins.toPath to normalize the path (it strips extra slashes).

This is a fork of the python-rewrite branch, for better or worse.

### DIFF
--- a/src/modules/top-level.nix
+++ b/src/modules/top-level.nix
@@ -194,8 +194,9 @@ in
         '';
       }
     ];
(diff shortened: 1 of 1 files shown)


domenkozar:
TIL

```
nix-repl> builtins.toPath "//a"
"/a"
```

PR #904, opened 2023-12-18
Better (but not completely fixed) treatment of venv

mcdonc:
- get rid of requirements check for recreation
    
    - store .devenv_interpreter because the venv
      python binary resolves to the origin
      python of the wrapper, not the wrapper
      python
    
    In practice, merging this commit will mean that
    
    a) the invocation of pip install -r will
    need to take care of any changed requirements.
    
    b) the venv, once created, will never be recreated
    unless the Python wrapper changes.  The Python
    wrapper doesn't seem to change when you add
    Python packages to your packages = list, I'm
    not sure what the other circumstances might be.
    
    In reality, this fix sucks because the venv is
    not actually based on the wrapper.  The wrapper,
    in fact, seems to never get used.  That will
    need to be addressed in some other commit, though.

Issue #909, opened 2023-12-23
Using custom Python from nixpkgs-python results in an interpreter unable to find glibc (I think)


mcdonc:
```
  languages.python = {
    enable = true;
    version = "3.11.6";
    venv = {
      enable = true;
    };
  };
```

```
inputs:
  nixpkgs:
    url: github:NixOS/nixpkgs/nixos-23.11
  nixpkgs-python:
    url: github:cachix/nixpkgs-python
  devenv:
    url: github:mcdonc/devenv/python-rewrite

allowUnfree: true
```

(this only works if you use the nixpkgs-python cachix cache or turn off sandboxing, see #901)

```
devenv shell
Building shell ...
trace: warning: For cachix.push and cachix.pull attributes to have an effect,
upgrade to devenv 1.0 or later.

Entering shell ...

Python interpreter changed, rebuilding Python venv...
/nix/store/m29jrr9cy9wg9kgy7k1dyr0gc44a5r43-python3-3.11.6-env/bin/python3.11 -m venv --upgrade-deps /home/chrism/projects/apex/eao_dash/.devenv/state/venv
/nix/store/3w7xj32jhxqhkvn66x67vql9lzmbffyw-python3-3.11.6/bin/python3.11: error while loading shared libraries: __vdso_gettimeofday: invalid mode for dlopen(): Invalid argument
bash: /home/chrism/projects/apex/eao_dash/.devenv/state/venv/.devenv_interpreter: No such file or directory
bash: /home/chrism/projects/apex/eao_dash/.devenv/state/venv/bin/activate: No such file or directory
bash: /home/chrism/projects/apex/eao_dash/.devenv/state/venv/bin/pip: No such file or directory
✨ devenv 0.6.3 is out of date. Please update to 1.0: https://devenv.sh/getting-started/#installation
(devenv) 
[chrism@thinknix512:~/projects/apex/eao_dash]$ /nix/store/m29jrr9cy9wg9kgy7k1dyr0gc44a5r43-python3-3.11.6-env/bin/python3.11
/nix/store/3w7xj32jhxqhkvn66x67vql9lzmbffyw-python3-3.11.6/bin/python3.11: error while loading shared libraries: __vdso_gettimeofday: invalid mode for dlopen(): Invalid argument
(devenv) 
```

mcdonc:
Note that working with nixpkgs-python directly works fine:

```
❯ nix build '.#"3.11.6"'
❯ ./result/bin/python -c "pass"
```

[...]

PR #912, opened 2023-12-24
Recreate Python venv and run pip install less often

mcdonc:
- Don't recreate Python venv unless Nix wrapper has changed.

- Don't run pip install -r /nix/store/xxx-requirements.txt unless requirements have changed.

This does not take into account use of ``-r`` in requirements.txt and probably should.

This patch is against the python-rewrite branch.

Supersedes #904

### DIFF
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -61,10 +61,6 @@ let
       echo ${package.interpreter} -m venv --upgrade-deps "$VENV_PATH"
       ${package.interpreter} -m venv --upgrade-deps "$VENV_PATH"
       echo "${package.interpreter}" > "$VENV_PATH/.devenv_interpreter"
-      if [ -n "$requirements" ]
-        then
-          echo "${requirements}" > "$VENV_PATH/.devenv_requirements"
-      fi
     fi
 
     source "$VENV_PATH"/bin/activate
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -45,24 +45,42 @@ let
 
     VENV_PATH="${config.env.DEVENV_STATE}/venv"
 
-    if [ "$(${readlink} "$VENV_PATH"/bin/python)" != "$(${readlink} ${package.interpreter}/bin/python)" ] \
-    || [ "$(${readlink} "$VENV_PATH"/requirements.txt)" != "$(${readlink} ${if requirements != null then requirements else "$VENV_PATH/requirements.txt"})" ]
+    profile_python="$(${readlink} ${package.interpreter})"
+    devenv_interpreter_path="$(${pkgs.coreutils}/bin/cat "$VENV_PATH/.devenv_interpreter" 2> /dev/null|| false )"
+    venv_python="$(${readlink} "$devenv_interpreter_path")"
+    requirements="${lib.optionalString (cfg.venv.requirements != null) ''${requirements}''}"
+
+    # recreate venv if necessary
+    if [ -z $venv_python ] || [ $profile_python != $venv_python ]
     then
-      if [ -d "$VENV_PATH" ]
-      then
-        echo "Python interpreter/requirements changed, rebuilding Python venv..."
-        ${pkgs.coreutils}/bin/rm -rf "$VENV_PATH"
-      fi
+      echo "Python interpreter changed, rebuilding Python venv..."
+      ${pkgs.coreutils}/bin/rm -rf "$VENV_PATH"
       ${lib.optionalString cfg.poetry.enable ''
         [ -f "${config.env.DEVENV_STATE}/poetry.lock.checksum" ] && rm ${config.env.DEVENV_STATE}/poetry.lock.checksum
       ''}
-      echo ${package.interpreter} -m venv "$VENV_PATH"
-      ${package.interpreter} -m venv "$VENV_PATH"
+      echo ${package.interpreter} -m venv --upgrade-deps "$VENV_PATH"
+      ${package.interpreter} -m venv --upgrade-deps "$VENV_PATH"
+      echo "${package.interpreter}" > "$VENV_PATH/.devenv_interpreter"
+      if [ -n "$requirements" ]
+        then
+          echo "${requirements}" > "$VENV_PATH/.devenv_requirements"
+      fi
     fi
+
     source "$VENV_PATH"/bin/activate
-    ${lib.optionalString (cfg.venv.requirements != null) ''
-      "$VENV_PATH"/bin/pip install -r ${requirements}
-    ''}
+
+    # reinstall requirements if necessary
+    if [ -n "$requirements" ]
+      then
+        devenv_requirements_path="$(${pkgs.coreutils}/bin/cat "$VENV_PATH/.devenv_requirements" 2> /dev/null|| false )"
+        devenv_requirements="$(${readlink} "$devenv_requirements_path")"
+        if [ -z $devenv_requirements ] || [ $devenv_requirements != $requirements ]
+          then
+            echo "${requirements}" > "$VENV_PATH/.devenv_requirements"
+            echo "Requirements changed, running pip install -r ${requirements}..."
+           "$VENV_PATH"/bin/pip install -r ${requirements}
+       fi
+    fi
   '';
 
   initPoetryScript = pkgs.writeShellScript "init-poetry.sh" ''


mcdonc:
To handle detecting changes in requirements specified vua ``-r``, we need to flatten requirements.

This thing almost works: https://requirements-parser.readthedocs.io/en/latest/usage.html#parsing-local-files

It chokes on envvars and relative ``-e`` inputs like ``-e .\.``

domenkozar:
Did the Python tests fail before? :thinking:

mcdonc:
> Did the Python tests fail before? 🤔

I'm glad there are so many tests to fail!  I haven't looked at each and every one, but most seem unrelated, and similar to failings like the recent Rust PR. 

Dumb question.  How can I run a subset of the tests locally?

mcdonc:
The failure https://github.com/cachix/devenv/actions/runs/7312821636/job/19926548496?pr=912  is the issue https://github.com/cachix/devenv/issues/909

mcdonc:
> > To handle detecting changes in requirements specified via `-r`, we need to flatten requirements.
> 
> Rather than using a Python lib to do this, we could parse the root requirements file and recursively concatenate all of the files referenced by `-r` into a single requirements.txt and recursively concatenate all the files referenced by `-c` into a single constraints.txt; this would negate the need to have a lib that was able to deal with envvars or -e or whatnot.

Like this:  https://gist.github.com/mcdonc/2653bbfbd33032a12464c8f64c183aff

src/modules/languages/python.nix
```diff
@@ -45,24 +45,38 @@ let
...
-        ${pkgs.coreutils}/bin/rm -rf "$VENV_PATH"
-      fi
+      echo "Python interpreter changed, rebuilding Python venv..."
+      ${pkgs.coreutils}/bin/rm -rf "$VENV_PATH"
       ${lib.optionalString cfg.poetry.enable ''
         [ -f "${config.env.DEVENV_STATE}/poetry.lock.checksum" ] && rm ${config.env.DEVENV_STATE}/poetry.lock.checksum
       ''}
-      echo ${package.interpreter} -m venv "$VENV_PATH"
-      ${package.interpreter} -m venv "$VENV_PATH"
+      echo ${package.interpreter} -m venv --upgrade-deps "$VENV_PATH"
```
bobvanderlinden, line 61: Doesn't `--upgrade-deps` introduce reproducibility issues? `.venv` will have different versions of `pip` depending on when you install it. (running it today will install X, running it tomorrow will install Y)

I'm not 100% sure this is a good idea, whether it should be the default and whether it should be behind an option.

bobvanderlinden:
> Did the Python tests fail before? 🤔

I suspect that the poetry test is also failing in the python rewrite PR:

https://github.com/cachix/devenv/actions/runs/7312821636/job/19926548605?pr=912#step:8:396

```
+ poetry run python -c 'import os; print(os.environ['\''LD_LIBRARY_PATH'\''])'
  File "<string>", line 1, in <module>
  File "<frozen os>", line 679, in __getitem__
KeyError: 'LD_LIBRARY_PATH'
```

bobvanderlinden (commented):
From what I understand, the goals of the PR should already have been met:

> Don't recreate Python venv unless Nix wrapper has changed.

Using the condition:

```
if [ "$(${readlink} "$VENV_PATH"/bin/python)" != "$(${readlink} ${package.interpreter}/bin/python)" ]
```

> Don't run pip install -r /nix/store/xxx-requirements.txt unless requirements have changed.

Using the condition:

```
[ "$(${readlink} "$VENV_PATH"/requirements.txt)" != "$(${readlink} ${if requirements != null then requirements else "$VENV_PATH/requirements.txt"})" ]
```

In what situation was this not working correctly? Do you have some reproducible steps to see the behaviour failing with the version on `main`?

mcdonc:
> > Did the Python tests fail before? 🤔
> 
> I suspect that the poetry test is also failing in the python rewrite PR:
> 
> https://github.com/cachix/devenv/actions/runs/7312821636/job/19926548605?pr=912#step:8:396
> 
> ```
> + poetry run python -c 'import os; print(os.environ['\''LD_LIBRARY_PATH'\''])'
> Traceback (most recent call last):
>   File "<string>", line 1, in <module>
>   File "<frozen os>", line 679, in __getitem__
> KeyError: 'LD_LIBRARY_PATH'
> ```

AFAICT, there is no guarantee that LD_LIBRARY_PATH will exist within the shell?

bobvanderlinden:
> > > Did the Python tests fail before? 🤔
> > 
> > 
> > I suspect that the poetry test is also failing in the python rewrite PR:
> > https://github.com/cachix/devenv/actions/runs/7312821636/job/19926548605?pr=912#step:8:396
> > ```
> > + poetry run python -c 'import os; print(os.environ['\''LD_LIBRARY_PATH'\''])'
> > Traceback (most recent call last):
> >   File "<string>", line 1, in <module>
> >   File "<frozen os>", line 679, in __getitem__
> > KeyError: 'LD_LIBRARY_PATH'
> > ```
> 
> AFAICT, there is no guarantee that LD_LIBRARY_PATH will exist within the shell?

With the current way `.so` libraries are linked, it is needed for some of the python packages to operate correctly. There probably is going something wrong in the python-rewrite branch that doesn't set the envvar when using poetry.

mcdonc:
I'm giving up on the -r/-c requirements.txt flattening due to impurity restrictions exercised by Nix.  This means the PR stands as-is delta the venv-recreate on interpreter-checking code, which can be simplified when #920 is fixed.

mcdonc:
> With the current way `.so` libraries are linked, it is needed for some of the python packages to operate correctly. There probably is going something wrong in the python-rewrite branch that doesn't set the envvar when using poetry.

We have some longstanding bugs/assumptions and multiple branches interacting, I think, which makes this complex and hard to describe, but I don't think that the python-poetry example should be relying on `LD_LIBRARY_PATH`, even though it happens to work on main.

Assumptions:

- Something somewhere should be setting LD_LIBRARY_PATH such that DLLs needed by Python libraries can be found  (true!)

Bugs:

- Nothing whatsoever (except the user) sets an LD_LIBRARY_PATH for the shell on the python-rewrite branch, nor will anything if all goes right.

(It looks like the main branch does indeed set an LD_LIBRARY_PATH in some cases, but I think it's incidental.)

Rationale:

The python-rewrite branch, unlike main, creates a wrapper script (https://github.com/cachix/devenv/blob/python-rewrite/src/modules/languages/python.nix#L13) for the Python defined in `languages.python.version`, and has two new additional options: `languages.python.libraries` and `languages.python.manylinux`, which are meant to work in concert with the wrapper.  This doesn't work currently but if #920 is fixed it will (and it must be, if we want to use a wrapper).

You can of course set `env.LD_LIBRARY_PATH` yourself and ignore `languages.python.libraries`.  This will indeed put an LD_LIBRARY_PATH in the shell environment, but this is the only time one would necessarily be present.  But if there isn't anything in `languages.python.libraries` and `languages.python.manylinux` is false, there won't necessarily be an LD_LIBRARY_PATH in Python's `os.environ` because all library path setting is presumed to be handled by the Python wrapper binary.

So I think the right fix is to change the test, because I think it's relying on an artifact of how main is working (which I don't actually understand, but I am pretty confident that the shell, even on main, might not have an LD_LIBRARY_PATH in it unless you set one explcitly.)

mcdonc:
@bobvanderlinden maybe you can help me understand the test in `examples/python-poetry/.test.sh`  (or really any of the example tests)... is its output checked by some validator or is the test just that all of these commands complete?

bobvanderlinden:
> This test will always return false currently. See https://github.com/cachix/devenv/issues/920 .

Hmm, this is probably a recent thing that broke it. It makes sense to fix it, but the current behaviour:

```
if [ "$(${readlink} "$VENV_PATH"/bin/python)" != "$(${readlink} ${package.interpreter}/bin/python)" ]
```
does still seem like it _should_ work. Just that there is a bug in nixpkgs that results in a wrong `"$VENV_PATH"/bin/python`. A workaround in devenv makes sense while the issue is being resolved in nixpkgs.

> There is never a "requirements.txt" within the venv. There might be one within $DEVENV_ROOT, but it's not guaranteed to be named "requirements.txt". We also shouldn't need to recreate the venv when the requirements change, we should just need to rerun pip install of the requirements.

Ah sorry, that line is indeed wrong. It should indeed be fixed.

> Nothing whatsoever (except the user) sets an LD_LIBRARY_PATH for Python or the shell on the python-rewrite branch.

The main branch sets `LD_LIBRARY_PATH` for the entire shell environment. `python-rewrite` sets it _inside_ the python wrapper. See https://github.com/cachix/devenv/pull/745/files#diff-f5bb89f776eda1a758db67c6c75b19ff1665b2f10998d17d1d98479835fc9e2eR16-R26. So if `languages.python.libraries` is set (which should be done for some of the python libraries that are being tested), `LD_LIBRARY_PATH` should be available inside python.

The `.test.sh` files merely need to succeed. The output does not matter, but the output can help quite a bit to get insight why a test fails. It was probably the reason to output `LD_LIBRARY_PATH` inside the test inside of python. It's likely that removing the line that is failing right now (printing `LD_LIBRARY_PATH`) will expose that loading zlib and pjsua will fail (because they aren't available in `LD_LIBRARY_PATH`). The test likely needs to move those packages from `packages` to `languages.python.packages`.

I'm not entirely sure why these tests aren't running on the python-rewrite branch. I'll try to create a PR that fixes the poetry test for the python-rewrite branch.

mcdonc:
> > Nothing whatsoever (except the user) sets an LD_LIBRARY_PATH for Python or the shell on the python-rewrite branch.
> 
> The main branch sets `LD_LIBRARY_PATH` for the entire shell environment. `python-rewrite` sets it _inside_ the python wrapper. See https://github.com/cachix/devenv/pull/745/files#diff-f5bb89f776eda1a758db67c6c75b19ff1665b2f10998d17d1d98479835fc9e2eR16-R26. So if `languages.python.libraries` is set (which should be done for some of the python libraries that are being tested), `LD_LIBRARY_PATH` should be available inside python.

 #920 is preventing it from working currently, but that's my understanding as well.

> The `.test.sh` files merely need to succeed. The output does not matter, but the output can help quite a bit to get insight why a test fails. It was probably the reason to output `LD_LIBRARY_PATH` inside the test inside of python. It's likely that removing the line that is failing right now (printing `LD_LIBRARY_PATH`) will expose that loading zlib and pjsua will fail (because they aren't available in `LD_LIBRARY_PATH`). The test likely needs to move those packages from `packages` to `languages.python.packages`.
> 
> I'm not entirely sure why these tests aren't running on the python-rewrite branch. I'll try to create a PR that fixes the poetry test for the python-rewrite branch.

Cool!

Issue #920, opened 2024-01-05
Python venv "python3.XX" links to base Python, not the Python binary wrapper 


mcdonc:
This began life as #901, but it deserves its own bug.

python-rewrite branch.

When ``languages.python.venv.enable = true;``, the resulting venv has a Python that points at the base Python instead of the binary wrapper created within ``python.nix``, even though the executable which is run to create the venv is indeed the wrapper.  This is a problem, because it defeats the purpose of ``languages.python.manylinux.enable`` and ``languages.python.libraries``.

I have not been able to convince venv itself to link to the wrapper.  I attempted to just change the ``bin/python3.11`` symlink in the venv so it points at the wrapper, but then when I try to use pip:

``` 
❯ devenv shell
• Building shell ...
✔ Building shell in 2.5s.
• Entering shell
Python interpreter changed, rebuilding Python venv...
/nix/store/ricl9i2v2707nmms9hf7xzfqsyf0xmw2-python3-3.11.6-env/bin/python3.11 -m venv --upgrade-deps /home/chrism/projects/devenv/tmp/.devenv/state/venv
(devenv) (venv) 
[chrism@optinix:~/projects/devenv/tmp]$ cd .devenv/state/venv/bin/
(devenv) (venv) 
[chrism@optinix:~/projects/devenv/tmp/.devenv/state/venv/bin]$ ln -sf /nix/store/ricl9i2v2707nmms9hf7xzfqsyf0xmw2-python3-3.11.6-env/bin/python3.11 python3.11
(devenv) (venv) 
[chrism@optinix:~/projects/devenv/tmp/.devenv/state/venv/bin]$ cat pip
#!/home/chrism/projects/devenv/tmp/.devenv/state/venv/bin/python3.11
# -*- coding: utf-8 -*-
import re
import sys
from pip._internal.cli.main import main
if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
(devenv) (venv) 
[chrism@optinix:~/projects/devenv/tmp/.devenv/state/venv/bin]$ pip install numpy
Traceback (most recent call last):
  File "/home/chrism/projects/devenv/tmp/.devenv/state/venv/bin/pip", line 5, in <module>
    from pip._internal.cli.main import main
ModuleNotFoundError: No module named 'pip'
```

If I force pip to be installed when the wrapper is created, it produces a pip that tries to install to the Nix store somehow.

```
[...]

PR #930, opened 2024-01-10
examples/python-poetry: use python libraries instead of packages

bobvanderlinden:
Requires https://github.com/cachix/devenv/issues/920

The `python-poetry` example is currently failing on the `python-rewrite` branch. The problem is that `LD_LIBRARY_PATH` is not supplied by `packages` anymore, but now is supplied using `languages.python.libraries`. This PR makes that change.

### DIFF
--- a/examples/python-poetry/devenv.nix
+++ b/examples/python-poetry/devenv.nix
@@ -1,7 +1,7 @@
 { pkgs, config, ... }:
 
 {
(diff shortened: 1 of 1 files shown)


mcdonc:
NB: this patch will have no effect until #920 is fixed but it is indeed the right patch.

PR #931, opened 2024-01-13
[Python] Fix -r/-c in venv.requirements

### LINKED ISSUE #889: language.python.venv.requirements does not support `-r` or `-c` directives
**Describe the bug**

language.python.venv.requirements cannot take paths or strings that include `-r` or `-c` directives.

**To reproduce**

1. `echo "requests" > requirements.txt`
2. `echo "-r requirements.txt" > dev.txt`

With this devenv.nix:

```nix
{ pkgs, ... }:
{
  languages.python = {
    enable = true;
    venv = {
      enable = true;
      requirements = ./dev.txt;
    };
  };
}
```
 
```
devenv shell
...
ERROR: Could not open requirements file: [Errno 2] No such file or directory: '/nix/store/requirements.txt'
...
```

However when in the shell:

```
> pip install -r dev.txt
Successfully installed certifi-2023.11.17 charset-normalizer-3.3.2 idna-3.6 requests-2.31.0 urllib3-2.1.0
```

**Version**

`devenv: 0.6.3`
 
**Workaround**

We have requirements from multiple locations added a function to interpret the requirements.txt based on the [file format](https://pip.pypa.io/en/stable/reference/requirements-file-format/)

```
{ pkgs, lib, ... }:
let
  # mergeRequirements simulates pip install command "-r"/"-c" parsing
  # requirements.txt paths.
  mergeRequirements = paths:
    with builtins;
    with lib;
    let
      evalLine = dir: line:
        let
          args = strings.splitString " " line;
          command = lists.head args;
          relative = filename: path.append dir filename;
        in
          if command == "-r" || command == "-c"
          then (map evalFile (map relative (lists.tail args)))
          else line;

      evalFile = path: map (evalLine (dirOf path)) (strings.splitString "\n" (readFile path));
      packages = (lists.flatten (map evalFile paths));
    in
      concatStringsSep "\n" packages;
in
{
  languages.python = {
    enable = true;

    venv = {
      enable = true;
      requirements = mergeRequirements [ ./dev.txt ];
    };
  };
}
```

I'm happy to propose a pull request, but would first appreciate advice for an approach.

streadway:
Fixes #889

When `pip install -r ${cfg.venv.requirements}` runs with a stored path that references other requirements files using the `-r` or `-c` syntax, the relative directory is /nix/store and pip could not find the referenced file under `config.devenv.root`.

This adds support for storing and rewriting depdendencies the requirements.txt has on other requirements or constraints files.

Example files:

  requirements.txt:
```
-r env/development.txt
```

  env/development.txt
```
pytest
-r production.txt
```

  env/production.txt
```
requests
```

Example devenv.nix path:

  language.python.venv.requirements = ./requirements.txt

Example devenv.nix string:

  language.python.venv.requirements = "-r env/development.txt"

Not supported in requirements.txt:

  * Parent paths like `-r ../other/requirements.txt`
  * Root paths like `-r /etc/requirements.txt`
  * Environment variables like `-r ${DEVENV_ROOT}/dev.txt`

### DIFF
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -3,10 +3,41 @@
 let
   cfg = config.languages.python;
 
-  requirements = pkgs.writeText "requirements.txt" (
+  writeRequirementsFile = { root, path, }:
+    let
+      process = originalLine:
+        let
+          pattern = "([[:space:]]*(-r|-c)[[:space:]]+)([^[:space:]]+)(.*)";
+          capture = builtins.match pattern originalLine;
+          leadingSpace = lib.elemAt capture 0;
+          relativePath = lib.elemAt capture 2;
+          trailingLine = lib.elemAt capture 3;
+          requiredPath = lib.path.append root relativePath;
+          replacedLine = leadingSpace + storedDependency + trailingLine;
+          storedDependency = writeRequirementsFile {
+            root = dirOf requiredPath;
+            path = requiredPath;
+          };
+        in
+        if capture == null
+        then originalLine
+        else replacedLine;
+
+      lines = lib.splitString "\n" (lib.readFile path);
+      withStorePaths = lib.concatStringsSep "\n" (map process lines);
+    in
+    pkgs.writeText (baseNameOf path) withStorePaths;
+
+  requirements = writeRequirementsFile (
     if lib.isPath cfg.venv.requirements
-    then builtins.readFile cfg.venv.requirements
-    else cfg.venv.requirements
+    then {
+      root = dirOf cfg.venv.requirements;
+      path = cfg.venv.requirements;
+    }
+    else {
+      root = /. + config.devenv.root;
+      path = pkgs.writeText "requirements.txt" cfg.venv.requirements;
+    }
   );
 
   nixpkgs-python = inputs.nixpkgs-python or (throw ''


streadway:
@mcdonc - I explored the fileset approach, but still needed to create a derivation from the collected source files and add an entrypoint for `pip install -r` for that to work. Instead this approach creates a dependency tree in the nix store.

I haven't tested if this will cause devenv to rebuild the shell if a nested requirements file is updated.

streadway:
@domenkozar for testing this change, I have a local python-venv-requirements example. Would you rather add a new example, or adjust an existing example to include the -r and -c commands in a directory of requirements files?

domenkozar:
A new example please :)

mcdonc:
How will do you think we should do change detection ala https://github.com/cachix/devenv/blob/python-rewrite/src/modules/languages/python.nix#L73 ?

streadway:
> How will do you think we should do change detection

The derivation path should change if any of the source files change, so I suspect no change to reloading logic is needed.

However, when patching and `devenv test`ing the python-rewite branch, it looks like the root has changed to a tempdir which causes this approach to fail from reading requirements at absolute paths when the initVenv script is built.

I am still understanding the nix build phases and how devenv relates to nix develop. (Newbie problems)

I will next try to create a derivation for pip install containing all source files referenced in by requirements files.

mcdonc:
> However, when patching and `devenv test`ing the python-rewite branch, it looks like the root has changed to a tempdir which causes this approach to fail from reading requirements at absolute paths when the initVenv script is built.

This is where I got to, too.

streadway:
I think the problem is resolving path under config.devenv.root from outside of devenv.nix (devenv/src/modules/languages/python.nix).

I'm pretty sure that all the paths to requirements files that could be depended on need to be provided to the requirements option.

This could be done in a few different ways.

 * Provide a new attribute to devenv.nix that processes the requirements and returns a path to the store used by the requirements option. `{ pkgs, lib, pipRequirements, ... }: {}`
 * Accept a new type (list, attrset, fileset) for the requirements option containing the requirements file paths and their entry point (what gets passed to `pip install -r`)

Here is an example flake.nix that shows the fileset approach collects requirement sources and requires a path relative to the fileset to install. This works with direnv `use flake` and `nix develop`.

```nix
{
  inputs = {
    nixpkgs = {
      url = "github:nixos/nixpkgs/nixos-unstable";
    };
    flake-utils = {
      url = "github:numtide/flake-utils";
    };
  };
  outputs = { nixpkgs, flake-utils, ... }: flake-utils.lib.eachDefaultSystem (system:
    let
      pkgs = import nixpkgs {
        inherit system;
      };

      lib = pkgs.lib;

      requirements = lib.fileset.toSource {
        root = ./.;
        fileset = lib.fileset.unions [
          ./requirements.txt
          ./requirements/dev.txt
          ./requirements/base.txt
          ./service_mocks/requirements.txt
        ];
      };

      pipInstall = "./requirements/dev.txt";

      ps = pkgs.python3Packages;

    in rec {
      devShell = pkgs.mkShell rec {
        buildInputs = [
          ps.python
          ps.venvShellHook
        ];

        venvDir = "./.venv";

        postVenvCreation = ''
          unset SOURCE_DATE_EPOCH
        '';

        postShellHook = ''
          unset SOURCE_DATE_EPOCH
        '';

        shellHook = ''
          source .venv/bin/activate

          if [ "$(readlink .venv/requirements-source)" != "${requirements}" ]; then
            if pip install -r ${requirements}/${pipInstall}; then
              ln -sfn "${requirements}" .venv/requirements-source
            fi
          fi
        '';
      };
    }
  );
}
```

streadway:
If the `venv.requirements` option is reused to provide the source for `pip install -r` then I think it's best done as an attrset. Something like:

```nix
{
  python.venv.requirements = {
    install = ./requirements/dev.txt;
    input = [ ./requirements/ ./requirements.txt ./other/requirements.txt ];
  };
}
```

For the attribute names to make sense, is there any precedence of other builders that take some source and an "entrypoint" to the source to complete a build? The attrset to create a derivation containing all requirements/constraints files could be named:

 * `{ root fileset }` Like lib.fileset.toSource
 * `{ out src }` Like mkDerivation
 * `{ requirement constraint src }` More specific to `pip -r -c` long names.

domenkozar:
> > However, when patching and `devenv test`ing the python-rewite branch, it looks like the root has changed to a tempdir which causes this approach to fail from reading requirements at absolute paths when the initVenv script is built.
> 
> This is where I got to, too.

I'm not sure anymore that changing the tempdir is the way to go, I'm still exploring how testing should work and open to suggestions.

domenkozar:
Superseeded by https://github.com/cachix/devenv/pull/964

PR #938, opened 2024-01-18
add dynamodb-local service

shyim:


src/modules/services/dynamodb-local.nix
```diff
@@ -0,0 +1,40 @@
...
+    };
+
+    port = lib.mkOption {
+      type = types.port;
+      description = "Listen address for the Dynamodb-local.";
+      default = 8000;
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
```
thenonameguy, line 37: Consider adding `process-compose` specific configuration as well:
https://github.com/cachix/devenv/blob/ff37923d4ba9e220b9826594e83fc29e1a4f4338/src/modules/services/postgres.nix#L298-L313

shyim: what would be the readiness command? kinda something like curl? as here is nothing special 🤔

thenonameguy: https://github.com/cachix/devenv/blob/ff37923d4ba9e220b9826594e83fc29e1a4f4338/src/modules/services/opensearch.nix#L158
Yeah, something like this would be great. With a lower initial_delay_second, as I guess this starts relatively fast.

thenonameguy: https://github.com/cachix/devenv/blob/ff37923d4ba9e220b9826594e83fc29e1a4f4338/src/modules/services/memcached.nix#L57-L59
Or just checking with netcat whether the port is bound. Even faster

thenonameguy (approved):

PR #939, opened 2024-01-18
Draft: Use a nixpkgs fork with the python-env-venv patch from cwp

mcdonc:
This is not meant for merge as-is; it's more of a "here's something we could do" to move along the python-rewrite branch.

- Use the patchset from https://github.com/NixOS/nixpkgs/pull/275701 in a nixpkgs fork (https://github.com/mcdonc/nixpkgs/tree/nixos-23.11-python-env-venv) to fix Python venv generation and library loading.

- Additional minor features: add a --keep-going flag to devenv test and handle ctrl-C gracefully during test runs.

Two languages don't work for reasons I can't quite figure out: swift and julia.  Neither compiles properly in my runs.

You'll want to do a "cachix use apex" before trying to do anything with this branch.  I've populated that cache with all the packages I could build from running all of the tests.

EDIT: julia now works, and I've confiirmed that the swift problem is unrelated to the Python changes (it doesn't compile on bare nixos-23.11 nixpkgs either).


mcdonc:
Here's a similar branch which proves that it also works with a fork of nixpkgs-unstable:

https://github.com/mcdonc/devenv/tree/python-rewrite-nixpkgs-unstable-env-venv-fixed-pkgs

domenkozar:
This is now part of #1005

PR #947, opened 2024-01-23
Rebase python rewrite on main

mcdonc:
I dunno man, this is a shot at rebasing python-rewrite on main.

Currently the direnv and phoenix tests fail.

The phoenix example is kinda problematic.  It has no associated nix code and the tests fail and I'm not sure how to fix it.


domenkozar:
Now part of #745, thanks a lot!

PR #954, opened 2024-01-26
Support allowBroken

jozuas:
## Motivation

Some packages are marked as broken, but can contain binaries that still work and are useful. One example of this is `ssconvert` from `pkgs.gnumeric` on Darwin. `pkgs.gnumeric` is marked as broken, but `ssconvert` works just fine, if `allowBroken = true;` is set.

## Testing

I've tested this with the aforementioned `pkgs.gnumeric` package, which is marked as broken. `allowBroken: true` in `devenv.yaml` allows to install `pkgs.gnumeric`.

### DIFF
--- a/docs/reference/yaml-options.md
+++ b/docs/reference/yaml-options.md
@@ -2,6 +2,7 @@
 | Key                          | Value                                                                         |
 | ---------------------------- | ----------------------------------------------------------------------------- |
 | allowUnfree                  | Allow unfree packages. Defaults to `false`.                                   |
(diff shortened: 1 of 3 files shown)


domenkozar:
Note that this needs to go into #745 since it's the next release of the cli.

jozuas:
Alright, I'll make a new PR from a branch based on `python-rewrite`. I thought simply changing the base would result be sufficient, but the diff got quite large.

jozuas:
Moved to https://github.com/cachix/devenv/pull/962

PR #956, opened 2024-01-28
Better error message when container copying fails

mcdonc:
When container copying fails, print the command that failed and the stderr/stdout of the command.  Without this, the user must pdb.


mcdonc:
THis PR is unnecessary, the output goes to stderr.

PR #963, opened 2024-01-28
Nonroot containers

mcdonc:
- Container environments will now have a "user" user.
- The "user" user will have a home dir at `/env`.
- All devenv files (the files referenced by `containers.<name>.copyToRoot`) will now end up in `/env` and will be owned by "user".
- The /nix/store will now be owned by the `user` user.
- `/env` is now the DEVENV_ROOT when within a container.
- the shell in a container environment is now `bashInteractive`, which respects up arrow, down arrow, search, etc.
- The default user in a container environment shell is now the "user" user. 
-  The "user" user will now run all processes and services.
- The max number of layers (a `nix2container` feature) is now an option. 

I'm still a bit unsure whether the permissions of the resulting /env files are the best we can do, but I think so.  They are 744 to the user user.

It would also be nice to not have container gen take so long.  My usage of layers here was to try to speed things up after first gen, but it seems the practical max number of layers is around 100 at least if it is to be a Docker image, and that doesn't help much because it's doing a layer per derivation I think.

### DIFF
--- a/docs/reference/options.md
+++ b/docs/reference/options.md
@@ -163,6 +163,21 @@ boolean
 
 
 
+## containers.\<name>.maxLayers
+
+The maximum number of layers created when the container is created.
+
+*Type:*
+int
+
+*Default:*
+` 1 `
+
+*Declared by:*
+ - [https://github.com/cachix/devenv/blob/main/src/modules/containers.nix](https://github.com/cachix/devenv/blob/main/src/modules/containers.nix)
+
+
+
 ## containers.\<name>.name
 
 Name of the container.
--- a/src/devenv/cli.py
+++ b/src/devenv/cli.py
@@ -681,8 +681,7 @@ def add(ctx, name, url, follows):
 )
 @click.argument("names", nargs=-1)
 @click.option("--debug", is_flag=True, help="Run tests in debug mode.")
-@click.option(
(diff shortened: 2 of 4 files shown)


mcdonc:
The permissions of the files copied from the devenv root are also jacked up.

domenkozar:
```
+ nix build --impure --accept-flake-config .#container-processes
trace: warning: `vendorSha256` is deprecated. Use `vendorHash` instead
error:
       … while calling the 'derivationStrict' builtin

         at /builtin/derivation.nix:9:12: (source not available)

       … while evaluating derivation 'image-processes.json'
         whose name attribute is located at /nix/store/gsydlhd8sya1wf5prv1p8sg4dd91gna4-source/pkgs/stdenv/generic/make-derivation.nix:352:7

       … while evaluating attribute 'buildCommand' of derivation 'image-processes.json'

         at /nix/store/gsydlhd8sya1wf5prv1p8sg4dd91gna4-source/pkgs/build-support/trivial-builders/default.nix:98:16:

           97|         enableParallelBuilding = true;
           98|         inherit buildCommand name;
             |                ^
           99|         passAsFile = [ "buildCommand" ]

       (stack trace truncated; use '--show-trace' to show the full trace)
```

domenkozar:
This is extremely cool :rocket:

PR #964, opened 2024-01-30
Flatten python requirements

mcdonc:
I figured out how to use the output of a Python script during enterShell to flatten requirements and reinstall when they've changed.

It has a backwards imcompatibilty:  "python.venv.requirements" now must be a string, and the string must be a devenv-root-relatvive path to a requirements file.  This will break anyone who uses the option.  Not sure how to signify this in the checkin without some sort of changelog file.

Unfinished:

[ ] tests


mcdonc:
This is related to https://github.com/cachix/devenv/issues/889

src/modules/languages/python.nix
```diff
@@ -176,10 +190,10 @@ in
     venv.enable = lib.mkEnableOption "Python virtual environment";
 
     venv.requirements = lib.mkOption {
-      type = lib.types.nullOr (lib.types.either lib.types.lines lib.types.path);
```
domenkozar, line 179: Could we instead do `nullOr path`? 

This breaks backwards compatibility where previously there would be a list of requirements written as string and now it's a path as string.

Is there a reason we can't support `lines` by writing to a file?

mcdonc: We can't use a path, because the file referenced by a path gets put into the store, and we lose the original name/path of the file in the user's environment.

mcdonc:
Superseded by #981

PR #975, opened 2024-02-09
Add an option to toggle defautls-extra-file for mysql cli commands

### LINKED ISSUE #973: mysql cli probably shouldn't define --defaults-file
**Describe the bug**

We try to use drush[1] in a devenv environment.
However we noticed that we couldn't run tasks that try to access the database because we got the following error:
```
Error Output:
  ================

  In SqlCommands.php line 183:

Query failed. Rerun with --debug to see any error message. /nix/store/g9w07skpvyy47a0l6xgyscdlv5dbd6f7-mariadb-server-10.11.6/bin/mysql: unknown variable 'defaults-file=/tmp/drush_Po73Fh'
```
It turned out that it was caused by the `--defaults-file` option on the mysql cli as defined by devenv[2][3].
When we used `--defaults-extra-file` in `devenv` the drush commands were working again because `--defaults-file` had precedence[4]. 

Would it make sense for devenv to use `--defaults-extra-file` instead of the current `--defaults-file`?

For the moment we override the command with the following snippet.
Would there be a better way?

```
scripts.mysql.exec = lib.mkForce ''
    exec ${config.services.mysql.package}/bin/mysql "$@"
'';
```

The other thing we noticed is  that even when we don't provide any options in `settings` it still generates an empty file and passes that with `--defaults-file`. Shouldn't the config file only be added as a parameter if `settings` actually contains any values?
Something like this?

```
if cfg.settings:
  configFile = format.generate "my.cnf" cfg.settings;

mysqlOptions = "";
if configFile:
  mysqlOptions = "--defaults-file=${configFile}";
```

[1]: https://www.drush.org/12.x/
[2]: https://github.com/cachix/devenv/blob/main/src/modules/services/mysql.nix#L11
[3]: https://github.com/cachix/devenv/blob/main/src/modules/services/mysql.nix#L258
[4]: https://dev.mysql.com/doc/refman/8.0/en/option-file-options.html#option_general_defaults-file

**To reproduce**

Will provide if required but I think it's more a thing to discuss as an actual bug.

Nebucatnetzer:
I made so that the toggle only affects the CLI user commands and not mysqld itself.

I tried to check for empty settings as well with this:
But it complains about `settings` being a set instead of a boolean, is there a way to check if a set is empty?
I only found ways to check if it has an attribute but I don't think there is an attribute that is always there when some adds custom settings.

```
configFile = if cfg.settings then format.generate "my.cnf" cfg.settings else "";
```

@shyim 

Closes #973

### DIFF
--- a/src/modules/services/mysql.nix
+++ b/src/modules/services/mysql.nix
@@ -8,8 +8,12 @@ with lib; let
   isMariaDB = getName cfg.package == getName pkgs.mariadb;
   format = pkgs.formats.ini { listsAsDuplicateKeys = true; };
   configFile = format.generate "my.cnf" cfg.settings;
-  mysqlOptions = "--defaults-file=${configFile}";
-  mysqldOptions = "${mysqlOptions} --datadir=$MYSQL_HOME --basedir=${cfg.package}";
+  mysqlOptions =
+    if !cfg.useDefaulsExtraFile then
+      "--defaults-file=${configFile}"
+    else
+      "--defaults-extra-file=${configFile}";
+  mysqldOptions = "--defaults-file=${configFile} --datadir=$MYSQL_HOME --basedir=${cfg.package}";
 
   initDatabaseCmd =
     if isMariaDB
@@ -174,6 +178,17 @@ in
       default = null;
       description = ''
         Whether to import tzdata on the first startup of the mysql server
+      '';
+    };
+
+    useDefaulsExtraFile = lib.mkOption {
+      type = types.bool;
+      default = false;
+      description = ''
+        Whether to use defaults-exta-file for the mysql command instead of defaults-file.
+        This is useful if you want to provide a config file on the command line.
+        However this can problematic if you have MySQL installed globaly because its config might leak into your environment.
+        This option does not affect the mysqld command.
       '';
     };
 


src/modules/services/mysql.nix
```diff
@@ -177,6 +181,17 @@ in
       '';
     };
 
+    useDefaulsExtraFile = lib.mkOption {
+      type = types.nullOr types.bool;
```
shyim, line 185:
```suggestion
      type = types.bool;
```

src/modules/services/mysql.nix
```diff
@@ -177,6 +181,17 @@ in
...
+      default = null;
```
shyim, line 186:
```suggestion
      default = true;
```

to keep it as before 🤔

Nebucatnetzer: That would be `false` then.

src/modules/services/mysql.nix
Nebucatnetzer: Updated it.

src/modules/services/mysql.nix
Nebucatnetzer: Updated it.

shyim (approved):

PR #979, opened 2024-02-19
use experimental installer

### LINKED ISSUE #509: Installation instructions
I think the installation instructions should favor https://determinate.systems installer rather than the default nix installer, it bundles flakes with the install, and is much easier to understand what its doing to your system. Works on macOS and Linux as far as I can tell.

I'm referring to this one;
`curl --proto '=https' --tlsv1.2 -sSf -L https://install.determinate.systems/nix | sh -s -- install`

from here;
https://zero-to-nix.com/start/install

domenkozar:
Fixes #509

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -7,18 +7,18 @@
 === "Linux"
 
     ```
-    sh <(curl -L https://nixos.org/nix/install) --daemon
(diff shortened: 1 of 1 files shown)


domenkozar:
Included in #1005

PR #981, opened 2024-02-20
Flatten python requirements 2

mcdonc:
I figured out how to use the output of a Python script during enterShell to flatten requirements and reinstall when they've changed.

It has a backwards imcompatibilty: "python.venv.requirements" now must be a string, and the string must be a devenv-root-relatvive path to a requirements file. This will break anyone who uses the option. Not sure how to signify this in the checkin without some sort of changelog file.


src/devenv/cli.py
```diff
@@ -179,6 +179,9 @@ def cli(ctx, offline, system, debugger, nix_flags, verbose):
     ctx.obj["gc_root"] = DEVENV_HOME_GC
     ctx.obj["gc_project"] = DEVENV_HOME_GC / str(int(time.time() * 1000))
 
+@cli.group()
```
domenkozar, line 182: duplicated below

mcdonc: fixed

domenkozar:
Ideally to land this we need import-from-derivation to work well. This is kind of blocked on #1548 unless we figure out a way for `requirements` to still be a `types.path` in another way.

PR #982, opened 2024-02-20
Nonroot containers 2

mcdonc:
Superseded #963

- Container environments will now have a "user" user.
- The "user" user will have a home dir at `/env`.
- All devenv files (the files referenced by `containers.<name>.copyToRoot`) will now end up in `/env` and will be owned by "user".
- The /nix/store will now be owned by the `user` user.
- `/env` is now the DEVENV_ROOT when within a container.
- the shell in a container environment is now `bashInteractive`, which respects up arrow, down arrow, search, etc.
- The default user in a container environment shell is now the "user" user. 
-  The "user" user will now run all processes and services.
- The max number of layers (a `nix2container` feature) is now an option. 

I'm still a bit unsure whether the permissions of the resulting /env files are the best we can do, but I think so.  They are 744 to the user user.

It would also be nice to not have container gen take so long.  My usage of layers here was to try to speed things up after first gen, but it seems the practical max number of layers is around 100 at least if it is to be a Docker image, and that doesn't help much because it's doing a layer per derivation I think.


mcdonc:
Wrong branch

PR #984, opened 2024-02-21
services/mysql: properly use wrappers during setup

lopter:
Otherwise config files outside devenv can get in the way and block database setup.


domenkozar:
cc @shyim

lopter:
Bump @shyim

shyim (approved):

PR #990, opened 2024-02-24
python: declarative python build via dream2nix

DavHau:


Atry:
If we are going to introduce dream2nix, why not poetry2nix, too? How would you compare dream2nix with poetry2nix?

aareman:
Looks like `dream2nix` is a generalized form of all the `X2nix` tools.

shahinism:
I see this is being implemented as part of the Python module. Given the design of dream2nix, isn't it better to have its module, so it can be used with other languages?

DavHau:
Yes, I was also thinking about a more generic approach. Maybe I'll have another shot at it some time.

domenkozar:
This was an experiment, let's close it for now.

PR #991, opened 2024-02-28
getting-started: use github url for better name inference

bobvanderlinden:
On Nix 2.20+, `nix profile` now refers to profile elements using names. These names are inferred from URLs. With the current URL it is inferred as `latest`, instead of `devenv`.

```
$ nix profile install tarball+https://install.devenv.sh/latest
$ nix profile upgrade devenv
warning: 'devenv' does not match any packages
warning: Use 'nix profile list' to see the current profile.
$ nix profile list
...
Name:               latest
```

Changing the URL would work, but for now referring to the github repo will also work:

```
$ nix profile install github:cachix/devenv/latest
$ nix profile upgrade devenv
$ nix profile list
Name:               devenv
...
```

If you'd like to look into changing the URL, you can make sure the last segment is named `devenv`. Or `devenv` is used in as the package in the flake instead of `default`. See https://github.com/NixOS/nix/blob/master/tests/unit/libexpr/flake/url-name.cc

I haven't touched the other instances of this URL, as those will likely work just fine.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -57,7 +57,7 @@ Recommended, speeds up the installation by providing binaries.
 === "Flake profiles"
 
     ```
-    nix profile install --accept-flake-config tarball+https://install.devenv.sh/latest
+    nix profile install --accept-flake-config github:cachix/devenv/latest
     ```
 === "Declaratively using flakes"
 


domenkozar:
How could we change the url to fix the current scheme? Maybe we can suffix it with `#devenv`?

bobvanderlinden:
Yes that would also be a solution. Alias default to devenv.
Another way would be to use `/latest/devenv` or `/devenv` as the url, bit that's a bit strange. I think `#devenv` would indeed be better.

domenkozar:
We can now use it from nixpkgs without indirection: https://github.com/cachix/devenv/commit/afa29528a39f06545658593a5329ab6e79f13be3

PR #993, opened 2024-03-01
Add options to Python language module

totoroot:
Recently I continued using devenv for various different projects at work, so the need to have it also manage multi-language development environments got bigger.

For a lot of projects, the structure looks similar to the following.

```
project
├── backend
├── devenv.lock
├── devenv.nix
├── devenv.yaml
└── frontend
```

In this example `backend` would for example be the subdirectory of some Python/Django component and then there would be the `frontend` subdirectory e.g. some JavaScript/Node.js component.

For that reason I would like to specify a subdirectory for the individual components.

In this PR I made the necessary changes to be able to specify a working directory via `languages.python.directory`.
While it is already possible to pass paths of requirements files to install dependencies from, it is currently not possible to specify the location of the virtual environment or to change the working directory of the poetry command.

This makes it impossible to have the `pyproject.toml` file anywhere other than at the devenv root.

I also added some more missing options to `languages.python.poetry.install`.

### DIFF
--- a/examples/python-directory/devenv.yaml
+++ b/examples/python-directory/devenv.yaml
@@ -1,0 +1,7 @@
+---
+
+inputs:
+  nixpkgs:
+    url: github:NixOS/nixpkgs/nixpkgs-unstable
+  nixpkgs-python:
+    url: github:cachix/nixpkgs-python
--- a/examples/python-directory/devenv.yaml
+++ b/examples/python-directory/devenv.yaml
@@ -1,5 +1,3 @@
----
-
 inputs:
   nixpkgs:
     url: github:NixOS/nixpkgs/nixpkgs-unstable
--- a/examples/python-poetry/devenv.nix
+++ b/examples/python-poetry/devenv.nix
@@ -1,4 +1,4 @@
-{ pkgs, lib, config, ... }:
+{ pkgs, config, ... }:
 
 {
   packages = [
@@ -15,6 +15,23 @@
 
   languages.python = {
     enable = true;
-    poetry.enable = true;
+    poetry = {
+      enable = true;
+      install = {
+        enable = true;
+        installRootPackage = false;
+        onlyInstallRootPackage = false;
+        compile = false;
+        quiet = false;
+        groups = [ ];
+        ignoredGroups = [ ];
+        onlyGroups = [ ];
+        extras = [ ];
+        allExtras = false;
+        verbosity = "no";
+      };
+      activate.enable = true;
+      package = pkgs.poetry;
+    };
   };
 }
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -190,7 +190,6 @@ in
       description = "Whether `pip install` should avoid outputting messages during devenv initialisation.";
     };
 
-
     directory = lib.mkOption {
       type = lib.types.str;
       default = ".";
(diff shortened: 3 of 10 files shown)


totoroot:
I'm planning on adding support for [pdm](https://github.com/pdm-project/pdm) soon, and for that, these changes would facilitate its use as well.

Also, I don't think we don't want to use `lib.types.path` for the directory, as that would mean copying the subdirectory's content to the Nix store. Using `lib.types.str` and the root of the top-level devenv project as default seems fine, but please let me know in case you think otherwise.

totoroot:
If nothing speaks against it, I would go ahead and try adding a similar option for the JavaScript module, since one might also want to have the `package.json` in a different directory than the devenv root

domenkozar:
See also #899, I'd really like to have a test for this before it's merged.

totoroot:
@domenkozar I modified the example for `python-poetry` to include the newly added options.

I also added an example and tests for the introduced directory option, as you requested. In this example I have also tested the `python-poetry` module with the Python directory option set. Let me know whether I also have to add them for the `python-venv` module.

totoroot:
@domenkozar I have modified the example for `python-poetry` to include the newly added options.

I also added an example and tests for the introduced directory option, as you requested. In this example I have also tested the `python-poetry` module with the Python directory option set.

totoroot:
Should I use `config.devenv.root` in the checks? :thinking:

examples/python-directory/devenv.yaml
```diff
@@ -0,0 +1,7 @@
+---
```
domenkozar, line 1:
```suggestion
```

src/modules/languages/python.nix
```diff
@@ -135,6 +135,14 @@ in
       description = "Whether `pip install` should avoid outputting messages during devenv initialisation.";
     };
 
+    directory = lib.mkOption {
```
domenkozar, line 138: does it work if you do `./.` with `types.path` and then do `config.devenv.root + cfg.directory`?

domenkozar:
Great work, only two comments and it's ready to go :)

src/modules/languages/python.nix
totoroot: I'm don't think I understand what you mean.

In https://github.com/cachix/devenv/pull/993#issuecomment-1973478436 I mentioned that using `types.str` means not copying the specified directory to the Nix store, so poetry is still able to write the lock file.

It also gives users more flexibility when setting the option as they can set it to either an absolute path or a path relative to the devenv root with something like `./directory`. I've added an example to the option.

domenkozar: I think you should be still able to use `types.path` to validate it's a path and then do `mystringpath + cfg.directory`

domenkozar:
@totoroot ping :)

domenkozar:
Can you rebase on main and I'll take over this one :)

bobvanderlinden:
I recently updated a python project's devenv and to my surprise it didn't activate the poetry environment anymore. It seems activating the poetry environment is disabled by default, where-as before I was able to import libraries from poetry without `poetry.activate.enable = true` nor `poetry run ...`. Was this change intentional?

examples/python-poetry/devenv.nix
```diff
@@ -15,6 +15,23 @@
...
+        quiet = false;
+        groups = [ ];
+        ignoredGroups = [ ];
+        onlyGroups = [ ];
+        extras = [ ];
+        allExtras = false;
+        verbosity = "no";
+      };
+      activate.enable = true;
+      package = pkgs.poetry;
```
bobvanderlinden, line 34: By default people shouldn't need to set all of these options, should the test check for options or for the default poetry environment? :sweat_smile:

totoroot:
@bobvanderlinden

> I recently updated a python project's devenv and to my surprise it didn't activate the poetry environment anymore. It seems activating the poetry environment is disabled by default, where-as before I was able to import libraries from poetry without `poetry.activate.enable = true` nor `poetry run ...`. Was this change intentional?

I also noticed poetry's activate option in devenv breaking a while ago, but I'm fairly certain this was a while after my changes with this PR got merged. If my changes caused this changed behaviour it wasn't intentional.

Unfortunately I currently got no time to investigate this, sorry.

domenkozar:
Could someone open a bug to reproduce it?

bobvanderlinden:
I'll have a look :+1:

bobvanderlinden:
https://github.com/cachix/devenv/issues/1907

PR #1003, opened 2024-03-11
feat: Add support languages.terraform.version

oscar-izval:
This PR follows the pattern of `languages.python` to allow for easy version selection in the Terraform language module using [stackbuilders/nixpkgs-terraform](https://github.com/stackbuilders/nixpkgs-terraform).

### DIFF
--- a/examples/terraform/devenv.yaml
+++ b/examples/terraform/devenv.yaml
@@ -1,0 +1,6 @@
+allowUnfree: true
+inputs:
+  nixpkgs:
+    url: github:NixOS/nixpkgs/nixpkgs-unstable
+  nixpkgs-terraform:
+    url: github:stackbuilders/nixpkgs-terraform
--- a/examples/terraform/devenv.yaml
+++ b/examples/terraform/devenv.yaml
@@ -1,6 +1,4 @@
 allowUnfree: true
 inputs:
-  nixpkgs:
-    url: github:NixOS/nixpkgs/nixpkgs-unstable
   nixpkgs-terraform:
     url: github:stackbuilders/nixpkgs-terraform
--- a/src/modules/languages/terraform.nix
+++ b/src/modules/languages/terraform.nix
@@ -35,7 +35,7 @@ in
 
   config = lib.mkIf cfg.enable {
     languages.terraform.package = lib.mkMerge [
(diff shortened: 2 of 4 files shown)


src/modules/languages/terraform.nix
```diff
@@ -13,9 +21,23 @@ in
...
+        The Terraform version to use.
+        This automatically sets the `languages.terraform.package` using [nixpkgs-terraform](https://github.com/stackbuilders/nixpkgs-terraform).
+      '';
+      example = "1.5.0 or 1.6.2";
+    };
   };
 
   config = lib.mkIf cfg.enable {
+    languages.terraform.package = lib.mkMerge [
+      (lib.mkIf (cfg.version != null) (nixpkgs-terraform.packages.${pkgs.stdenv.system}.${cfg.version} or (throw "Unsupported Terraform version, see https://github.com/stackbuilders/nixpkgs-terraform/blob/main/versions.json for the full list of supported versions.")))
```
sestrella, line 38: @oscar-izval maybe the version already exists in a new version of `nixpkgs-terraform`, in that sense, I think the first recommendation should be to try to upgrade the `nixpkgs-terraform` input first or look at `versions.json` for the full list of supported versions, alternatively. What do you think?

domenkozar:
Hey looks really good! Let's wait for #1005 as it will require changes - mostly there's now a helper function to render what inptus to add

src/modules/languages/terraform.nix
```diff
@@ -1,7 +1,15 @@
-{ pkgs, config, lib, ... }:
+{ pkgs, config, lib, inputs, ... }:
 
 let
   cfg = config.languages.terraform;
+
+  nixpkgs-terraform = inputs.nixpkgs-terraform or (throw ''
```
domenkozar, line 6: If you rebase on `main` you can now do

```nix
config.lib.getInput {
    name = "nixpkgs-terraform";
    url = "github:stackbuilders/nixpkgs-terraform";
    attribute = "languages.terraform.version";
  };
```

oscar-izval: Nice! That's definitely useful

examples/terraform/devenv.yaml
```diff
@@ -0,0 +1,6 @@
+allowUnfree: true
+inputs:
+  nixpkgs:
```
domenkozar, line 3: Let's leave out nixpkgs, so that the defaults apply

domenkozar (changes requested):

examples/terraform/devenv.yaml
oscar-izval: Done!

domenkozar:
Thank you 🎉

Issue #540, opened 2023-04-20
OSError: AF_UNIX path too long


lzhu-aurorasolar:
**Describe the bug**

I'm getting the following Traceback when trying to run `pytest` within a devenv shell. The tests run without issues outside of devenv shell either on local environment or in a Docker container.

```python
Process SyncManager-1:
Traceback (most recent call last):
  File "/nix/store/r23sw9iy2jdrh59dsq8mic6ahkcijijr-python3-3.10.10/lib/python3.10/multiprocessing/process.py", line 314, in _bootstrap
    self.run()
  File "/nix/store/r23sw9iy2jdrh59dsq8mic6ahkcijijr-python3-3.10.10/lib/python3.10/multiprocessing/process.py", line 108, in run
    self._target(*self._args, **self._kwargs)
  File "/nix/store/r23sw9iy2jdrh59dsq8mic6ahkcijijr-python3-3.10.10/lib/python3.10/multiprocessing/managers.py", line 591, in _run_server
    server = cls._Server(registry, address, authkey, serializer)
  File "/nix/store/r23sw9iy2jdrh59dsq8mic6ahkcijijr-python3-3.10.10/lib/python3.10/multiprocessing/managers.py", line 156, in __init__
    self.listener = Listener(address=address, backlog=16)
  File "/nix/store/r23sw9iy2jdrh59dsq8mic6ahkcijijr-python3-3.10.10/lib/python3.10/multiprocessing/connection.py", line 448, in __init__
    self._listener = SocketListener(address, family, backlog)
  File "/nix/store/r23sw9iy2jdrh59dsq8mic6ahkcijijr-python3-3.10.10/lib/python3.10/multiprocessing/connection.py", line 591, in __init__
    self._socket.bind(address)
OSError: AF_UNIX path too long
```

**To reproduce**
Please provide an [sscce](http://sscce.org) by [creating a gist](https://gist.github.com/) using `devenv.nix`, `devenv.yaml`, and optionally `devenv.lock`.

`devenv.nix`:

```nix
{ pkgs, ... }:

{

  name = "app";

  # https://devenv.sh/packages/
  packages = [
    pkgs.git
    pkgs.direnv
    pkgs.gdal
    pkgs.postgresql_14
    pkgs.openssl_1_1
    pkgs.postgis
    pkgs.geos
    pkgs.redis
  ];

  # https://devenv.sh/languages/
  languages.python.enable = true;
  languages.python.venv.enable = true;

[...]

PR #1005, opened 2024-03-11
Rust rewrite

domenkozar:
Building upon https://github.com/cachix/devenv/pull/745, more on why another rewrite on the announcement post.

```
❯ ./result/bin/devenv
https://devenv.sh 1.0.0: Fast, Declarative, Reproducible, and Composable Developer Environments

Usage: devenv [OPTIONS] <COMMAND>

Commands:
  init       Scaffold devenv.yaml, devenv.nix, .gitignore and .envrc.
  shell      Activate the developer environment. https://devenv.sh/basics/
  update     Update devenv.lock from devenv.yaml inputs. http://devenv.sh/inputs/
  search     Search for packages and options in nixpkgs. https://devenv.sh/packages/#searching-for-a-file
  info       Print information about this developer environment.
  up         Start processes in the foreground. https://devenv.sh/processes/
  processes  Start or stop processes.
  test       Run tests. http://devenv.sh/tests/
  container  Build, copy, or run a container. https://devenv.sh/containers/
  inputs     Add an input to devenv.yaml. https://devenv.sh/inputs/
  gc         Deletes previous shell generations. See http://devenv.sh/garbage-collection
  build      Build any attribute in devenv.nix.
  version    Print the version of devenv.
  help       Print this message or the help of the given subcommand(s)

Options:
  -v, --verbose
          Enable debug log level.
  -j, --max-jobs <MAX_JOBS>
          Maximum number of Nix builds at any time. [default: 8]
  -j, --cores <CORES>
          Maximum number CPU cores being used by a single build.. [default: 2]
  -s, --system <SYSTEM>
          [default: x86_64-linux]
  -i, --impure
          Relax the hermeticity of the environment.
  -c, --clean [<CLEAN>...]
          Ignore existing environment variables when entering the shell. Pass a list of comma-separated environment variables to let through.
  -d, --nix-debugger
          Enter Nix debugger on failure.
  -n, --nix-option <NIX_OPTION> <NIX_OPTION>
          Pass additional options to nix commands, see `man nix.conf` for full list.
  -o, --override-input <OVERRIDE_INPUT> <OVERRIDE_INPUT>
          Override inputs in devenv.yaml.
  -h, --help
          Print help
```

Fixes https://github.com/cachix/devenv/issues/744
Fixes https://github.com/cachix/devenv/issues/693
Fixes https://github.com/cachix/devenv/issues/658
Fixes https://github.com/cachix/devenv/issues/244
Fixes https://github.com/cachix/devenv/issues/436
[...]

PR #1008, opened 2024-03-16
feat: add DEVENV_RUNTIME directory

midchildan:
This PR creates a new directory, `DEVENV_RUNTIME`. Its purpose is similar to `/run` and `XDG_RUNTIME_DIR`. It's created under `/tmp` with a short path name so that Unix domain sockets can be created there without hitting path length limits.

This also changes PostgreSQL to use `DEVENV_RUNTIME` so that it can work without worrying about the path length of the project directory. It should fix the issue described in #540. For a complete fix, modules that use Unix domain sockets also need to be fixed.

### DIFF
--- a/src/modules/top-level.nix
+++ b/src/modules/top-level.nix
@@ -34,6 +34,8 @@ let
         ${lib.concatStringsSep "\n" (builtins.map formatAssertionMessage failedAssertions)}
       ''
     else lib.trivial.showWarnings config.warnings;
+
+  q = lib.escapeShellArg;
 in
 {
   options = {
@@ -122,6 +124,11 @@ in
         internal = true;
       };
 
+      runtime = lib.mkOption {
+        type = types.str;
+        internal = true;
+      };
+
       profile = lib.mkOption {
         type = types.package;
         internal = true;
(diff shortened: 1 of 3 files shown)


src/modules/top-level.nix
```diff
@@ -163,8 +170,23 @@ in
     devenv.state = config.devenv.dotfile + "/state";
     devenv.profile = profile;
 
+    # The path has to be
+    # - unique to each DEVENV_ROOT to let multiple devenv environments coexist
+    # - deterministic so that it won't change constantly
+    # - short so that unix domain sockets won't hit the path length limit
```
midchildan, line 176: I chose not to use `$TMPDIR` because it's too long on macOS.

```console
❯ echo "$TMPDIR" | wc -c
      50
```

src/modules/top-level.nix
```diff
@@ -192,6 +214,10 @@ in
       mkdir -p .devenv
       rm -f .devenv/profile
       ln -s ${profile} .devenv/profile
+
+      mkdir -p ${q config.devenv.runtime}
+      rm -f ${q config.devenv.dotfile}/run
+      ln -s ${q config.devenv.runtime} ${q config.devenv.dotfile}/run
```
midchildan, line 220: A symlink is added for convenience.

domenkozar:
Hey, this is great! Could you make a PR against #1005 and add a test into `tests/cli` to assert runtime directory is set? 

Sorry for the mess, but we're working towards 1.0

domenkozar:
@midchildan if you'd like I can port it over since I'm working on fixing all the tests :)

domenkozar:
Merged into #1005, thank you!

midchildan:
Thanks!

PR #1016, opened 2024-03-20
Touch up 1.0 release post

euphemism:
Caught me while I was sleeping 😩.

### DIFF
--- a/docs/blog/posts/devenv-v1.0-rewrite-in-rust.md
+++ b/docs/blog/posts/devenv-v1.0-rewrite-in-rust.md
@@ -1,6 +1,6 @@
 ---
 draft: false
-date: 2023-03-20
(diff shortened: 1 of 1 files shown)


docs/blog/posts/devenv-v1.0-rewrite-in-rust.md
```diff
@@ -1,6 +1,6 @@
 ---
 draft: false
-date: 2023-03-20
+date: 2024-03-20
```
euphemism, line 3: 😬. I wonder why we, as humans, struggle with the change in year. It happens pretty consistently, you'd think we would be better at handling it. I feel like March is around the time when we generally get it together and stop accidentally writing out year `n - 1`.

PR #1018, opened 2024-03-20
Avoid --impure in flake-parts based projects

Atry:
Let `.envrc` generate the `devenv.root`

### DIFF
--- a/templates/flake-parts/.envrc
+++ b/templates/flake-parts/.envrc
@@ -5,7 +5,8 @@ fi
 nix_direnv_watch_file devenv.nix
 nix_direnv_watch_file devenv.lock
 nix_direnv_watch_file devenv.yaml
-if ! use flake . --impure
+
+if ! use flake . --override-input devenv-root "file+file://"<(printf %s "$PWD")
 then
   echo "devenv could not be built. The devenv environment was not loaded. Make the necessary changes to devenv.nix and hit enter to try again." >&2
 fi
--- a/templates/flake-parts/flake.nix
+++ b/templates/flake-parts/flake.nix
@@ -2,6 +2,10 @@
   description = "Description for the project";
 
   inputs = {
+    devenv-root = {
+      url = "file+file:///dev/null";
+      flake = false;
+    };
     nixpkgs.url = "github:cachix/devenv-nixpkgs/rolling";
     devenv.url = "github:cachix/devenv";
     nix2container.url = "github:nlewo/nix2container";
@@ -14,7 +18,7 @@
     extra-substituters = "https://devenv.cachix.org";
   };
 
-  outputs = inputs@{ flake-parts, ... }:
+  outputs = inputs@{ flake-parts, devenv-root, ... }:
     flake-parts.lib.mkFlake { inherit inputs; } {
       imports = [
         inputs.devenv.flakeModule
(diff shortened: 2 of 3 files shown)


domenkozar:
We should probably make it a JSON to pass other things like `devenv.tmpdir`?

Atry:
> We should probably make it a JSON to pass other things like `devenv.tmpdir`?

I think one value per input approach would be nice, otherwise we will have to deal with JSON escape things in `.envrc`

Atry:
@domenkozar Would you like to approve the CI workflows?

templates/flake-parts/.envrc
```diff
@@ -5,7 +5,8 @@ fi
 nix_direnv_watch_file devenv.nix
 nix_direnv_watch_file devenv.lock
 nix_direnv_watch_file devenv.yaml
-if ! use flake . --impure
+
+if ! use flake . --override-input devenv-root "file+file:"<(pwd)
```
khaled, line 9: this doesn't work for me (nix 2.18.1 aarch64-darwin) - needs to be `"file+file://"<(pwd)`

templates/flake-parts/flake.nix
```diff
@@ -2,6 +2,10 @@
   description = "Description for the project";
 
   inputs = {
+    devenv-root = {
+      url = "file+file:/dev/null";
```
khaled, line 6: similar to my prev comment - this needs to be `"file+file:///dev/null"`

khaled:
Without [these changes](https://github.com/cachix/devenv/pull/1018#pullrequestreview-1955724841) I get:

```
error:
       … while evaluating the attribute 'optionalValue.value'

         at /nix/store/lwyjz70qh12nq6cb7fixl85vryzxqm3c-source/lib/modules.nix:856:5:

          855|
          856|     optionalValue =
             |     ^
          857|       if isDefined then { value = mergedValue; }

       … while evaluating a branch condition

         at /nix/store/lwyjz70qh12nq6cb7fixl85vryzxqm3c-source/lib/modules.nix:857:7:

          856|     optionalValue =
          857|       if isDefined then { value = mergedValue; }
             |       ^
          858|       else {};

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: unable to download 'file://file:/dev/fd/63': URL using bad/illegal format or missing URL (3)
```


*With* the changes I get a little further, but apparently there is still a problem:

```
$direnv allow
...
Hello, world!
ln: failed to create symbolic link '/Users/khaled/path/to/project'$'\n''/.devenv/run': No such file or directory
```

Atry:
@khaled Fixed

khaled:
Thanks @Atry!

FYI, this probably belongs in a separate bug, but the existence of the `packages.devenv-up` package definition seems to actually break `devenv up`:

```
       error: attribute 'config' missing

       at /nix/store/86gyzipyn7zzp4nrnanz0sl5r6cps63v-source/flake.nix:48:30:

           47|         packages.default = pkgs.hello;
           48|         packages.devenv-up = self'.devShells.default.config.procfileScript;
             |                              ^
           49|         devenv.shells.default = let
 ```

Atry:
@domenkozar I don't understand the CI failure message. Could you help me understanding why?

Atry:
@domenkozar I rebased this PR. Would you mind approving the CI workflows?

Atry:
@domenkozar I think the CI failure is not due to this PR

lopter:
I am wondering if it would make sense to also update the [flake integration wrapper](https://github.com/cachix/devenv/blob/8827aa19daf1fc3f53e7adcc7201933ef28f8652/src/devenv-devShell.nix) to not use `--impure` anymore and maybe re-export other nix commands from there (e.g. the flake nix command).

Atry:
@domenkozar could you trigger CI for this PR?

Issue #1007, opened 2024-03-14
MySQL configuration on startup fails when trying to set up users


M-arcus:
**Describe the bug**
MySQL configuration on startup fails when trying to set up users after updating to newest devenv.lock

**To reproduce**
```
{ pkgs, lib, ... }:
{
  services.mysql.enable = true;
  services.mysql.package = lib.mkDefault pkgs.mysql80;
  services.mysql.initialDatabases = [{ name = "db"; }];
  services.mysql.ensureUsers = [{
    name = "db";
    password = "db";
    ensurePermissions = { "*.*" = "ALL PRIVILEGES"; };
  }];
  processes.entryscript.exec = ''
    ${pkgs.coreutils}/bin/sleep infinity
  '';
}
```
works.

But adding this breaks the devenv up command:
```
  services.mysql.settings = {
    mysql = {
      user = "db";
      password = "db";
      host = "127.0.0.1";
    };
  };
```

Error log:
```
10:59:42 mysql-configure.1 | mysqld is alive
10:59:42 mysql-configure.1 | ERROR 1045 (28000): Access denied for user 'root'@'localhost' (using password: YES)
```

**Version**
 
```
$ devenv version
devenv: 0.6.3
```

M-arcus:
@schneider-felix mentioned, that this could be the cause
https://github.com/cachix/devenv/commit/71809bca1921dcc2c72e3749891ec91ed0fb78c4

M-arcus:
@domenkozar @shyim 
Looks like it's broken again, executing `mysql` does not use the configured values anymore.

domenkozar:
We really need a test for this

shyim:
https://github.com/cachix/devenv/blob/main/tests/mysql/.test.sh

PR #1021, opened 2024-03-20
services/mysql: use empty config while mysql server bootstrapping, fi…

shyim:
…xes #1007

When the user configures a mysql client password, it affects also our configuration script and this blocks the bootstrap process.

I created similar wrapper as before, but with empty config file.


domenkozar:
Let's also add a test to tests folder?

shyim:
test added :)

shyim:
image
mysql passed. it's failling because of search. btw search prints ALLL PACKAGES. therefore the ci logs page does not load 😆

PR #1056, opened 2024-03-26
rust: use mold

domenkozar:
Should we use mold by default to speed up compilation?


domenkozar:
Seems like we'll need to use sold to support macOS: https://github.com/rui314/mold/issues/1171

PR #1064, opened 2024-03-26
fix: `Env` should not be escaped

Atry:
According to https://github.com/opencontainers/image-spec/blob/main/config.md:

> Entries are in the format of `VARNAME=VARVALUE`. These values act as defaults and are merged with any specified when creating a container.

The `VARVALUE` part does not understand shell escape or quotes.

### DIFF
--- a/src/modules/containers.nix
+++ b/src/modules/containers.nix
@@ -141,7 +141,7 @@ let
       WorkingDir = "${homeDir}";
       Env = lib.mapAttrsToList
         (name: value:
-          "${name}=${lib.escapeShellArg (toString value)}"
+          "${name}=${toString value}"
         )
         config.env ++ [ "HOME=${homeDir}" "USER=${user}" ];
       Cmd = [ cfg.startupCommand ];


domenkozar:
We should really have some tests for these, can you show an example?

Atry:
Use `--entrypoint env` to skip `mk-shell-bin`'s `envScript`, which overrides the environment variables from the image manifest.

```bash
cd examples/fly.io && nix run ../.. -- container copy shell && docker run --rm -it --entrypoint env shell:latest
```
```
• Building shell container ...
• Using Cachix: devenv
warning: creating lock file '/examples/fly.io/devenv.lock'
trace: warning: `vendorSha256` is deprecated. Use `vendorHash` instead
/nix/store/i98cgbww6qifd626wzvn8jyz00va80nz-image-shell.json
✔ Building shell container in 2.0s.
warning: creating lock file '/examples/fly.io/devenv.lock'
trace: warning: `vendorSha256` is deprecated. Use `vendorHash` instead
• Running /nix/store/amca1wlni4szd4p8drmf6accq31ml58n-copy-container /nix/store/i98cgbww6qifd626wzvn8jyz00va80nz-image-shell.json false 

Copying container /nix/store/i98cgbww6qifd626wzvn8jyz00va80nz-image-shell.json to docker-daemon:shell:latest

Getting image source signatures
Copying blob 79833476e9bf done   | 
Copying blob a460459aeca2 done   | 
Copying config 3f15e603df done   | 
Writing manifest to image destination
✔ Copying shell container in 2.8s.
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=f9ea7b473441
TERM=xterm
DEVENV_DOTFILE='/env/.devenv'
DEVENV_PROFILE='/nix/store/ra30qvay59hqy3j9fxwrb0ks20l6ifbw-devenv-profile'
DEVENV_ROOT='/env'
DEVENV_RUNTIME='/tmp/devenv-078a6e4'
DEVENV_STATE='/env/.devenv/state'
HOME=/env
USER=user
```

You can see the single quotes around `DEVENV_ROOT` and other environment variables.

Atry:
This bug also affects `docker exec` because `docker `exec` does not use `Entrypoint` nor `envScript`.

``` bash
docker exec "$(docker run --detach --rm -it shell:latest sleep infinity)" env
```

```
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=661680970c87
DEVENV_DOTFILE='/env/.devenv'
DEVENV_PROFILE='/nix/store/ra30qvay59hqy3j9fxwrb0ks20l6ifbw-devenv-profile'
DEVENV_ROOT='/env'
DEVENV_RUNTIME='/tmp/devenv-078a6e4'
DEVENV_STATE='/env/.devenv/state'
HOME=/env
USER=user
```

You can see the environment variables are incorrect.

Atry:
@domenkozar Can we fix this?

domenkozar:
I'm afraid this will break other things, we for sure added escaping to fix something in the past.

Atry:
I think it is never used to fix anything in the past

Atry:
The current `Env` settings are simply broken. We did not notice it because they only take effect on `docker exec`, as `docker run` takes envionment variables from `mk-shell-bin`'s `envScript`, not the image's `Env`.

PR #1069, opened 2024-03-27
modules: exclude more environment variables

imincik:
Remove some more environment variables and update corresponding option
description.

### DIFF
--- a/src/modules/top-level.nix
+++ b/src/modules/top-level.nix
@@ -70,15 +70,17 @@ in
 
     unsetEnvVars = lib.mkOption {
       type = types.listOf types.str;
-      description = "Remove these list of env vars from being exported to keep the shell/direnv more lean.";
+      description = "A list of removed environment variables to make the shell/direnv more lean.";
       # manually determined with knowledge from https://nixos.wiki/wiki/C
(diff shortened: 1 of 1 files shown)


src/modules/top-level.nix
```diff
@@ -70,15 +70,17 @@ in
...
       default = [
         "HOST_PATH"
         "NIX_BUILD_CORES"
         "__structuredAttrs"
+        "builder"
         "buildInputs"
         "buildPhase"
-        "builder"
+        "cmakeFlags"
+        "configureFlags"
```
domenkozar, line 83: I'm not sure we should exclude `cmakeFlags`, `configureFlags` and `mesonFlags` because someone might want to add a library that provides these via hooks to `packages` to build their C program.

imincik: I thought `cmakeFlags` + other flags excluded in this PR are in the same category as other already excluded variables. 

> someone might want to add a library that provides these via hooks to packages to build their C program.

Is this going to work when for example `buildInputs` is removed from shell (`buildInputs` is already in `unsetEnvVars` list ) ?

Issue #1078, opened 2024-03-28
"path '/devenv.nix' does not exist in Git repository" when ignored by git


hqm42:
**Describe the bug**
After upgrading to from < 1 to 1.0.2 I started getting the error
`path '/devenv.nix' does not exist in Git repository`

**Context**
As I work in a constrained environment, I can not add any devenv related files into our git repositories. I am using an envelope project that syncs the devenv config files into the respective project directories:
```sh
> tree envelope
envelope
├── devenv
│   ├── repo-1
│   │   ├── devenv.nix
│   │   └── devenv.yaml
│   └── repo-2
│       ├── devenv.nix
│       └── devenv.yaml
├── repo-1
│   └── ...
├── repo-2
    └── ...
```

The devenv specific files are ignored by git not using the normal `.gitignore` but instead using the repo local `.git/info/exclude` file:
```sh
> cat repo-1/.git/info/exclude
# git ls-files --others --exclude-from=.git/info/exclude
# Lines that start with '#' are comments.
# For a project mostly in C, the following would be a good set of
# exclude patterns (uncomment them if you want to use them):
# *.[oa]
# *~
.devenv*
devenv.local.nix
.direnv
.pre-commit-config.yaml
devenv.nix
devenv.lock
devenv.yaml
.envrc
```

**To reproduce**
```sh
mkdir devenv-test
cd devenv-test
git init
devenv init
echo devenv.nix >> .gitignore
direnv reload
```

```
...
error:
error: path '/devenv.nix' does not exist in Git repository '/home/xxx/devenv-test'
```

**Version** 
devenv 1.0.2 (x86_64-darwin)

hqm42:
For now it would be enough for me to be able to downgrade to an oder devenv version, but I did not find a simple way to do that.

domenkozar:
Hmmm, I'll look into this.

zevisert:
I'm also seeing this, as I also frequently use devenv to work on customer, internal, and oss repositories that don't track devenv files. My ignore solution is a bit simpler though; it's just a global gitignorefile tracked by my dotfiles that I can include via `git config` in the repositories that I need

```console
> # if a repo isn't using devenv, just tell git to ignore its files
> git config --local --set core.excludesfile ~/.devenv.gitignore
```

```ini
# file: ~/.devenv.gitignore
# Devenv
.devenv*
devenv.local.nix

# direnv
.direnv

.envrc
devenv.lock
devenv.nix
devenv.yaml
```

[...]

Issue #257, opened 2022-12-27
Devenv slows down with big code repositories


shyim:
**Describe the bug**

When you have a big project and maybe some cache like 10 GB it gets really slowed down by git commands. Maybe that command can be ignored?

```
 |   \-+= 25589 root /usr/bin/login -fpl shyim /Applications/iTerm.app/Contents/MacOS/ShellLauncher --launch_shell
 |     \-+= 25590 shyim -zsh
 |       \-+- 25994 shyim -zsh
 |         \-+- 25995 shyim /usr/local/bin/direnv export zsh
 |           \-+- 25996 shyim /bin/bash -c eval "$("/usr/local/bin/direnv" stdlib)" && __main__ source_env $'/Users/shyim/Code/shopware/cloud/.envrc'
 |             \-+- 26011 shyim /bin/bash -c eval "$("/usr/local/bin/direnv" stdlib)" && __main__ source_env $'/Users/shyim/Code/shopware/cloud/.envrc'
 |               \-+- 26012 shyim bash /Users/shyim/.nix-profile/bin/devenv print-dev-env
 |                 \-+- 26021 shyim /nix/store/ynwgbyq3p1db8wg5hw8b192l3336g65b-nix-2.12.0pre20221221_51b770e/bin/nix --show-trace --extra-experimental-features nix-command --extra-experimental-features flakes print-dev-env --impure
 |                   \--- 26027 shyim git -C /Users/shyim/Code/shopware/cloud --git-dir .git ls-files --cached -z --others
```

shyim:
`--others` seems like REALLY slow down because it looks then also in the cache folder / node_modules and so

shyim:
Seems like introduced by custom nix https://github.com/domenkozar/nix/commit/bf9922e7b21b83450006640d2382e04e5a203692

domenkozar:
We need this to be able to support .devenv.flake.nix, probably a better way is to just to add it to the output.

shyim:
We could also generate a generic flake.nix and commit that? 🤔

domenkozar:
@shyim do you have a workaround for this or does it need to be fixed soon?

shyim:
I am thinking about adding a hybrid solution. `.envrc` uses nix flake directly and maybe the users still use `devenv up`. 
and the flake just imports existing devenv.nix

domenkozar:
Flakes verbatim copy the whole repo into the /nix/store so that's also slower in general :)

[...]

PR #1080, opened 2024-03-28
feat(dotenv): allow passing full paths to dotenv files

alirezamirsepassi:
This pull request enables the dotenv integration to accept full paths to the dotenv files. Additionally, I've changed the option `filename` to `files` and moved the normalize function to the option itself using `lib.toList`. I've also added a warning regarding the updated option name to prevent users from breaking their environments.

P.S. I chose to use `config.devenv.root` instead of `self` because `self` evaluates to the nix store source tree when used in a flake configuration. Since dotenv files are typically not committed to a git repository, they won't be present after the copy.

### DIFF
--- a/devenv.nix
+++ b/devenv.nix
@@ -231,9 +231,6 @@ EOF
     generate-doc-css = {
       enable = true;
       name = "generate-doc-css";
-      # In CI, the auto-commit action doesn't run in the shell, so it can't reuse our scripts.
-      # And the following command is curently too slow to be a pre-commit command.
-      # entry = "devenv shell devenv-generate-doc-css";
       entry = config.scripts."devenv-generate-doc-css".exec;
     };
   };


src/modules/integrations/dotenv.nix
```diff
@@ -37,13 +33,18 @@ let
...
     enable = lib.mkEnableOption ".env integration, doesn't support comments or multiline values.";
 
-    filename = lib.mkOption {
-      type = lib.types.either lib.types.str (lib.types.listOf lib.types.str);
-      default = ".env";
-      description = "The name of the dotenv file to load, or a list of dotenv files to load in order of precedence.";
+    files = lib.mkOption {
+      type = lib.types.either lib.types.path (lib.types.listOf lib.types.path);
+      apply = lib.toList;
+      default = config.devenv.root + "/.env";
```
shyim, line 46: I don't think allowing any path is good. This would make it more hard to remove `impure` nix mode, as we can access files outside of our project 🤔

alirezamirsepassi: Thanks for your response, I appreciate it. I agree with your point of view, but after doing some research, I found out that flakes [copy the source to Nix store ](https://github.com/NixOS/nix/issues/3121) before evaluating it, and there's no way around this. As a result, using a git ignored dotenv file is impossible. 😭 

Nevertheless, I do think that allowing users to choose the dotenv file path, even if it's outside of the project, is a great idea as it provides more flexibility. For example, I often have variables that I need to share across multiple projects, and a single dotenv file can be used for this purpose.

wdyt? 🤔

shyim: The project should not be copied into nix store with devenv provided nix 🤔 . But yes the gitignore has an effect 😓 .

Maybe we find a general solution to .gitignore, because of other issues like https://github.com/cachix/devenv/issues/1078

alirezamirsepassi: How does the devenv CLI create the shell btw? 🤔 I'm not very familiar with rust honestly but what I understood from [here](https://github.com/cachix/devenv/blob/main/devenv/src/main.rs#L466), is it runs `nix develop` with some options and arguments, right? 🤔 does it copy ignored files as well?

shyim: You stay in your local normal copy where you did run devenv shell. I am right now looking if to improve this 😅

alirezamirsepassi: @domenkozar hi there!

I saw that you changed `config.devenv.root` to `self` in [the rewrite in rust project](https://github.com/cachix/devenv/pull/1005), this, unfortunately, breaks my environments where I use flakes as I mentioned in the original description of this PR.  Do you happen to know the command that devenv runs to create the shell? Maybe including that command in the .envrc file along with the flakes template would make sense as in: currently the dotenv integration doesn't work in combination with flakes. wdyt? 🤔

mschnee: Also confirmed that this change has broken dotenv support in devenv when using flakes.  

```shell
flake-test via ❄ impure (devenv-shell-env) …
➜ direnv reload
direnv: loading ~/Projects/flake-test/.envrc
direnv: loading https://raw.githubusercontent.com/nix-community/nix-direnv/2.2.1/direnvrc (sha256-zelF0vLbEl5uaqrfIzbgNzJWGmLzCmYAkInj/LNxvKs=)
direnv: nix-direnv: renewed cache
💡 The dotenv file '.env' was not found.
```

Using the default flake created by `nix flake init --template github:cachix/devenv` as [documented here](https://devenv.sh/guides/using-with-flakes/).
```nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-23.11";
    devenv.inputs.nixpkgs.follows = "nixpkgs";
  };

  nixConfig = {
    extra-trusted-public-keys = "devenv.cachix.org-1:w1cLUi8dv3hnoSPGAuibQv+f9TZLr6cv/Hm9XgU50cw=";
    extra-trusted-substituters = "https://devenv.cachix.org";
  };

  outputs = { self, nixpkgs, devenv, systems, ... } @ inputs:
    let
      forEachSystem = nixpkgs.lib.genAttrs (import systems);
    in
    {
      packages = forEachSystem (system: {
        devenv-up = self.devShells.${system}.default.config.procfileScript;
      });

      devShells = forEachSystem
        (system:
          let
          in
          {
              inherit inputs pkgs;
              modules = [
                {
                  dotenv.enable = true;
                  # https://devenv.sh/reference/options/
                  packages = [ pkgs.hello ];

                  enterShell = ''
                    hello
                  '';

                  processes.run.exec = "hello";
                }
              ];
            };
          });
    };
}
```

domenkozar: @shyim can you try using this PR to see if it addresses #257

shyim: We don't use dotenv module and I don't see it why it should improve the performance 🤔

domenkozar: It would still evaluate `self` that should be impacting the performance

domenkozar:
This breaks on devenv with:

```
       error:
       error: A definition for option `dotenv.files."[definition 1-entry 1]"' is not of type `path'. Definition values:
       - In `/nix/store/virtual0000000000000000000000007-source/src/modules/integrations/dotenv.nix': ".env"
```

because the semantics of the paths are different, I'll think about it how to properly integrate it.

alirezamirsepassi:
@domenkozar I switched the type back to string and wrapped the default value in a double quotation to avoid nix copying it to the store.

Could you try again on devenv, please?

alirezamirsepassi:
@domenkozar any updates on this?

domenkozar:
Thanks for the ping, CI is running

PR #1082, opened 2024-03-28
feat: add nix caching

### LINKED ISSUE #1079: Move nix caching from direnv plugin to devenv cli
Right now we have caching inside `direnv` to make folder switching instant. It would be nice if we could move the caching into devenv generally, so that `devenv shell` and `devenv up` also improves.

shyim:
Fixes #1079

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,6 +84,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "arrayref"
+version = "0.3.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6b4930d2cb77ce62f89ee5d5289b4ac049559b1c45539271f5ed4fdc7db34545"
+
+[[package]]
+name = "arrayvec"
+version = "0.7.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
+
+[[package]]
 name = "atty"
 version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


domenkozar:
Why don't we use `print_dev_env` function when we know it's cached?

This also needs to respect `imports` in `devenv.yaml` when calculating the hash.

shyim:
`print_dev_env ` is called by direnv and direnv does his own caching 🤔 
I am not sure if we want to ditch the caching of direnv completely, as it right now works fine 🤔. and we would at least have ~30ms latency switching between projects as devenv needs to be called

domenkozar:
> `print_dev_env ` is called by direnv and direnv does his own caching 🤔 I am not sure if we want to ditch the caching of direnv completely, as it right now works fine 🤔. and we would at least have ~30ms latency switching between projects as devenv needs to be called

Could you explain the goal of this PR? Then it makes it easier to discuss possible implementations :)

shyim:
Make `devenv shell` and `devenv up` faster on existing projects without any changes. When merging the direnv caching is possible, it would be nice to have one place of caching.

domenkozar:
So kind of like direnv caching without direnv?

shyim:
Yes excactly

domenkozar:
I'll tackle this shortly, it's important to get back to blazing fast :)

shyim:
further random idea: Could we get from one `nix eval` back all environment variables + the process startup script.

Instead of right now doing three nix commands, we could do only one 🤔

domenkozar:
That's actually what I'm exploring, we'd need to reimplement print-dev-env in Rust and then we got it all.

Will post results as soon as I get something working!

therealpxc:
I use `devenv shell` in CI for jobs where it's important to be fast. Taking the lead time down from several seconds to less than a second will make a big difference for those short jobs!

domenkozar:
Hey, we're finally going to address this! I'm going to close this PR as we're taking a bit different route that makes sure any changes to Nix evaluation are properly detected and will open a PR once it's ready for testing! 

Thank you for this work as an inspiration :heart:

domenkozar:
https://github.com/cachix/devenv/pull/1468

PR #1087, opened 2024-04-01
Added test for running gleam using the starter project

abeljim:


domenkozar:
@abeljim do you want to add things like discussed in the issue?

domenkozar:
Fixed :)

Issue #1088, opened 2024-04-01
Failing tests


domenkozar:
We should fix all the tests and accept zero-test policy now with devenv-nixpkgs

As of [this commit](https://github.com/cachix/devenv/actions/runs/8450216802/job/23174203105) we have failing tests:

- `tests`: it just fails without any output, needs investigation
- `examples/terraform`: due to #792 
- `examples/phoenix`: no idea how to fix it /cc @sevillaarvin
- `examples/rubyonrails`: cc @bobvanderlinden 

To run an example: `PATH=$PWD/result/bin:$PATH ./result/bin/devenv-run-tests --only rubyonrails  examples`

totoroot:
I'm not sure why, but when running `PATH=$PWD/result/bin:$PATH ./result/bin/devenv-run-tests --only rubyonrails examples` for a second time, tests are passing just fine :thinking:

mattthym:
Hmm, tests for `rubyonrails` and `phoenix` are still failing, even after 'fixes' got merged. The database connection can't be established for both.

totoroot:
Hmm, tests for rubyonrails and phoenix are still failing, even after 'fixes' got merged. The database connection can't be established for both.

bobvanderlinden:
> Hmm, tests for rubyonrails and phoenix are still failing, even after 'fixes' got merged. The database connection can't be established for both.

At least for rubyonrails: the test isn't deterministic. It depends on something from the system that isn't in devenv/envvar. Maybe a leftover process that isn't killed after running the test? 

It seems that once it succeeds, it'll keep succeeding. Need to dig into this again (third time 😅) and maybe make sure it succeeds after a reboot.

domenkozar:
```
❯ devenv up
warning: creating lock file '/examples/rubyonrails/devenv.lock'
• Building processes ...
• Using Cachix: devenv
warning: creating lock file '/examples/rubyonrails/devenv.lock'
✔ Building processes in 1.3s.
• Starting processes ...
• Building shell ...
warning: creating lock file '/examples/rubyonrails/devenv.lock'
✔ Building shell in 1.0s.
/home/domen/dev/cachix/devenv/examples/rubyonrails/blog /home/domen/dev/cachix/devenv/examples/rubyonrails
Bundle complete! 14 Gemfile dependencies, 81 gems now installed.
Bundled gems are installed into `/home/domen/dev/cachix/devenv/examples/rubyonrails/.devenv/state/.bundle`
[...]

PR #1100, opened 2024-04-05
fix: Terraform example

sestrella:
Related with #1088


domenkozar:
That's a good compromise for now, I still plan to look into fixing #1088

PR #1110, opened 2024-04-08
devenv test: wait for processes to start

domenkozar:
- [ ] https://github.com/seanmonstar/reqwest/pull/1623
- [ ] check if this fixes rubyonrails example
- [ ] enable only when tests use process-compose


domenkozar:
For some reason even if the `self.up()` fails, the PID is still running so it takes 60s to timeout.

domenkozar:
Going to take a different route.

PR #1112, opened 2024-04-08
add-uv-pip-replacement

cpossinger:
This PR adds an option to enable the [new uv pip drop in replacement](https://github.com/astral-sh/uv) for installing requirements and creating a virtual environment.

### DIFF
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -59,18 +59,15 @@ let
       ${lib.optionalString cfg.poetry.enable ''
         [ -f "${config.env.DEVENV_STATE}/poetry.lock.checksum" ] && rm ${config.env.DEVENV_STATE}/poetry.lock.checksum
       ''}
-      ${lib.optionalString cfg.uv.enable ''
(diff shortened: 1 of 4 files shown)


src/modules/languages/python.nix
```diff
@@ -59,8 +59,17 @@ let
       ${lib.optionalString cfg.poetry.enable ''
         [ -f "${config.env.DEVENV_STATE}/poetry.lock.checksum" ] && rm ${config.env.DEVENV_STATE}/poetry.lock.checksum
       ''}
-      echo ${package.interpreter} -m venv --upgrade-deps "$VENV_PATH"
-      ${package.interpreter} -m venv --upgrade-deps "$VENV_PATH"
+      ${lib.optionalString cfg.uv.enable ''
```
domenkozar, line 62: It would be better to use `if cond then s else s` here for better readability

domenkozar (changes requested):

domenkozar:
Could you also add a test to `tests`?

domenkozar:
Thank you!

PR #1121, opened 2024-04-12
Add missing package property to the pre-commit docs

dz0ny:
pre-commit-hooks now require `package` property which was not present in the docs.


domenkozar:
It defaults to `null` if you use newest pre-commit repo rev.

dz0ny:
Can be closed then.

PR #1129, opened 2024-04-14
fix 1.0.4 regression: python venv always recreated

mcdonc:
Fix regression introduced by https://github.com/cachix/devenv/commit/6ff0c3a81ccbf79f00b5f3289f77df21114bd87e where Python venv will always be recreated unconditionally even if there is no reason to do so.


domenkozar:
We really need a test for this.

PR #1134, opened 2024-04-15
Streamline .envrc used in templates

dliberalesso:
As per ```nix-direnv```'s Readme, there is no need to explicitly watch ```flake.nix``` or ```flake.lock``` beacause it already does that.

Also, ```nix_direnv_watch_file``` has been deprecated.


dliberalesso:
It depends on the user environment, right?

The first thing to pay attention to, is that the following code enforces version 2.2.1 even if the user has a newer version of `nix-direnv`.

```sh
if ! has nix_direnv_version || ! nix_direnv_version 2.2.1; then
fi
```

Which means we should be updating these files everytime a new release of `nix-direnv` comes around. Why should we have to do that? (They are on version 3.0.4, btw).

The other thing is: the user needs to have `nix` and `direnv` already installed for this to work. Why not let the user manage their prefered `nix-direnv` too?

PR #1142, opened 2024-04-17
Add devenv-up into the shell environment

judofyr:
This brings devenv-up as a package which is brought into the shell environment. `devenv-up` is now present in the $PATH and runs it immediately and `devenv up` execs.

The main motivation here is when using a Flake directly. Previously `devenv up` would run `nix build` on demand which is (1) fragile since it makes assumptions around what the flake is actually called and (2) doesn't register it as root.

By bringing the devenv-up into shell we solve both of these problems. Be aware that we're not able to bring it into the _profile_ since this leads to a infinite recursion: The devenv-up script is designed to run "outside" of the devenv environment (since the process-compose.yaml file sets the environment variables for you).

This commit doesn't change behavior in the Rust implementation, but I believe we could now also tweak this to invoke `devenv-up` directly instead of adding it as a root.

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -47,9 +47,7 @@ devenvFlake: { flake-parts-lib, lib, inputs, ... }: {
                 { "${shellPrefix shellName}container-${containerName}" = container.derivation; }
               )
               devenv.containers
-            ) // {
-              "${shellPrefix shellName}devenv-up" = devenv.procfileScript;
-            }
+            )
           )
           config.devenv.shells;
     });
--- a/src/devenv-devShell.nix
+++ b/src/devenv-devShell.nix
@@ -10,8 +10,6 @@ pkgs.writeScriptBin "devenv" ''
   # we want subshells to fail the program
   set -e
 
-  NIX_FLAGS="--show-trace --extra-experimental-features nix-command --extra-experimental-features flakes"
-
   command=$1
   if [[ ! -z $command ]]; then
     shift
@@ -19,13 +17,7 @@ pkgs.writeScriptBin "devenv" ''
 
   case $command in
     up)
-      procfilescript=$(nix build '.#${shellPrefix (config._module.args.name or "default")}devenv-up' --no-link --print-out-paths --impure)
-      if [ "$(cat $procfilescript|tail -n +2)" = "" ]; then
-        echo "No 'processes' option defined: https://devenv.sh/processes/"
-        exit 1
-      else
-        exec $procfilescript "$@"
-      fi
+      exec devenv-up "$@"
       ;;
     version)
       echo "devenv: ${version}"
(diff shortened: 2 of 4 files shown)


judofyr:
While pondering more about this, I think this would have been pretty neat if …

* processes.nix contains `packages = [ config.procfileBin ]`. This is what brings `devenv-up` into the environment.
* `process-compose.yaml` will _not_ include the environment variables, but instead the assumption is that `devenv-up` is executed under the right environment.
* `devenv exec` makes it possible to run an arbitrary command under the environment. It's a no-op if you're already inside the environment.
* `devenv shell` is basically an alias to `devenv exec $SHELL`
* `devenv up` is basically an alias to `devenv exec devenv-up`

Not sure if this is actually possible to implement in a backwards compatible way, or if there are other problems here…

judofyr:
The following diff (on top of this branch) works fine for my setup (where I call `devenv-up` under `nix-direnv`), but I'm not sure if this breaks other scenarios?

```diff
diff --git i/src/modules/process-managers/process-compose.nix w/src/modules/process-managers/process-compose.nix
index d8acb4a..13df529 100644
--- i/src/modules/process-managers/process-compose.nix
+++ w/src/modules/process-managers/process-compose.nix
@@ -53,9 +53,6 @@ in
         is_strict = true;
         port = lib.mkDefault 9999;
         tui = lib.mkDefault true;
-        environment = lib.mapAttrsToList
-          (name: value: "${name}=${toString value}")
-          config.env;
         processes = lib.mapAttrs
           (name: value: { command = "exec ${pkgs.writeShellScript name value.exec}"; } // value.process-compose)
           config.processes;
diff --git i/src/modules/processes.nix w/src/modules/processes.nix
index bbfe61b..d590ca6 100644
--- i/src/modules/processes.nix
+++ w/src/modules/processes.nix
@@ -153,6 +153,8 @@ in
       wait
     '';
 
+    packages = [ config.procfileBin ];
+
     ci = [ config.procfileScript ];
 
     infoSections."processes" = lib.mapAttrsToList (name: process: "${name}: exec ${pkgs.writeShellScript name process.exec}") config.processes;
diff --git i/src/modules/top-level.nix w/src/modules/top-level.nix
index a8ac7b2..9df3fdc 100644
--- i/src/modules/top-level.nix
+++ w/src/modules/top-level.nix
@@ -281,7 +281,7 @@ in
     shell = performAssertions (
       (pkgs.mkShell.override { stdenv = config.stdenv; }) ({
         name = "devenv-shell";
-        packages = config.packages ++ [config.procfileBin];
+        packages = config.packages;
         shellHook = ''
           ${lib.optionalString config.devenv.debug "set -x"}
           ${config.enterShell}
```

domenkozar:
Closing as outdated

Issue #1153, opened 2024-04-19
Processes are failing to start in flake based config


nick-kadutskyi:
**Describe the bug**
Running `devenv up` in `nix develop --impure` shell fails in flake config based on default flake template:
```
❯ nix flake init --template github:cachix/devenv
wrote: /Users/nick/Developer/PE/Test/devenvsh-2/.envrc
wrote: /Users/nick/Developer/PE/Test/devenvsh-2/.gitignore
wrote: /Users/nick/Developer/PE/Test/devenvsh-2/flake.nix


.devenv should be added to .gitignore

      |   echo .devenv >> .gitignore

direnv: error /Users/nick/Developer/PE/Test/devenvsh-2/.envrc is blocked. Run `direnv allow` to approve its content
❯ nix develop --impure
warning: creating lock file '/Users/nick/Developer/PE/Test/devenvsh-2/flake.lock'
warning: Nix search path entry '/nix/var/nix/profiles/per-user/root/channels' does not exist, ignoring
Hello, world!
(devenv) Nicks-MacBook-Air:devenvsh-2 nick$ devenv up
warning: Nix search path entry '/nix/var/nix/profiles/per-user/root/channels' does not exist, ignoring
{"level":"warn","error":"open /Users/nick/Library/Application Support/process-compose/settings.yaml: no such file or directory","time":"2024-04-19T09:55:37-07:00","message":"Error reading settings file /Users/nick/Library/Application Support/process-compose/settings.yaml"}
24-04-19 09:55:37.351 FTL start UDS http server on /tmp/nix-shell.2W2rLL/devenv-76311b3/pc.sock failed error="listen unix /tmp/nix-shell.2W2rLL/devenv-76311b3/pc.sock: bind: no such file or directory"
(devenv) Nicks-MacBook-Air:devenvsh-2 nick$ hello
Hello, world!
```

When I use `flake-parts` flake template it fails in a different way. I get the same error when I run `devenv up` on fresh template or if I add some process (`processes.run.exec="hello"`) in `devenv.shells.default`:
```
❯ mkdir devenvsh-test && cd $_
❯ nix flake init --template github:cachix/devenv#flake-parts
wrote: /Users/nick/Developer/PE/Test/devenvsh-test/.envrc
wrote: /Users/nick/Developer/PE/Test/devenvsh-test/.gitignore
wrote: /Users/nick/Developer/PE/Test/devenvsh-test/flake.nix


.devenv should be added to .gitignore

[...]

PR #1156, opened 2024-04-21
Fix flake-parts template

### LINKED ISSUE #1027: Flakes: flake-parts template fails to build devenv-up
**Describe the bug**

It is not possible to run services when using Devenv 1.0 with flake parts template.

`devenv-up` package fails to build with following error:

```
nix build .#devenv-up --impure

    error: attribute 'config' missing

       at /nix/store/0wba1p6cz6s7ydm9bsx8vjcny6z9ss1p-source/flake.nix:30:30:

           29|         # needed for devenv up
           30|         packages.devenv-up = self'.devShells.default.config.procfileScript;
             |                              ^
           31|
```

**To reproduce**
```
mkdir test && cd test

git init
nix flake init -t github:cachix/devenv#flake-parts

nix build .#devenv-up --impure

       error: attribute 'config' missing

       at /nix/store/91iwalpza3pmz4lwfjh53q8w1bk55v2q-source/flake.nix:30:30:

           29|         # needed for devenv up
           30|         packages.devenv-up = self'.devShells.default.config.procfileScript;
             |                              ^
           31|
```

My Nix version:
```
nix-env --version
nix-env (Nix) 2.18.1
```

sandydoo:
Fixes #1027.
Fixes half of #1153.

~Also, updated all the templates to use `devenv/latest`, instead of `main`.~

### DIFF
--- a/templates/flake-parts/flake.nix
+++ b/templates/flake-parts/flake.nix
@@ -30,9 +30,6 @@
         # module parameters provide easy access to attributes of the same
         # system.
 
-        # needed for devenv up
-        packages.devenv-up = self'.devShells.default.config.procfileScript;
-
         # Equivalent to  inputs'.nixpkgs.legacyPackages.hello;
         packages.default = pkgs.hello;
 
@@ -57,6 +54,8 @@
           enterShell = ''
             hello
           '';
+
+          processes.hello.exec = "hello";
         };
 
       };
--- a/templates/simple/flake.nix
+++ b/templates/simple/flake.nix
@@ -37,7 +37,7 @@
                     hello
                   '';
 
-                  processes.run.exec = "hello";
+                  processes.hello.exec = "hello";
                 }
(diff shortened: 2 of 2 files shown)


templates/flake-parts/flake.nix
```diff
@@ -7,7 +7,7 @@
       flake = false;
     };
     nixpkgs.url = "github:cachix/devenv-nixpkgs/rolling";
-    devenv.url = "github:cachix/devenv";
+    devenv.url = "github:cachix/devenv/latest";
```
domenkozar, line 10: doesn't this mean that `devenv` modules won't advance any faster than cli?

sandydoo: Ah, I see what you mean. I see people getting hit by bugs on `main` on first launch. Not a great experience, but I guess we need another approach here. Will revert.

Issue #1109, opened 2024-04-08
unable to change process-compose tui


albertilagan:
**Describe the bug**
I was previously setting the TUI and PORT using a `env` file.
```
PC_TUI_ENABLED=false
PC_HTTP_PORT=9999
```

After upgrading to v1, it's not working anymore.

**Version**
 
devenv 1.0.3 (aarch64-darwin)

Issue #1104, opened 2024-04-06
`devenv up` fails with error


MatzHilven:
**Describe the bug**
When I try executing `devenv up` I get the following error.
It should start a postgres and redis process.

**To reproduce**
https://gist.github.com/MatzHilven/4201608fc91df95d2a852e3c6d386bb7

```
• Building processes ...
• Using Cachix: devenv
✔ Building processes in 1.2s.
• Starting processes ...• Building shell ...
✔ Building shell in 1.0s.
✨ devenv 1.0.2 is out of date. Please update to 1.0.3: https://devenv.sh/getting-started/#installation
panic: exec: "infocmp": executable file not found in $PATH

goroutine 13 [running]:
github.com/f1bonacc1/process-compose/src/tui.SetupTui({0xe5d8f0?, 0xc000662320?}, {0xc0000a1f80, 0x2, 0x0?})
        github.com/f1bonacc1/process-compose/src/tui/view.go:482 +0x156
github.com/f1bonacc1/process-compose/src/cmd.startTui({0xe5d8f0, 0xc000662320})
        github.com/f1bonacc1/process-compose/src/cmd/project_runner.go:92 +0x179
created by github.com/f1bonacc1/process-compose/src/cmd.runTui in goroutine 1
        github.com/f1bonacc1/process-compose/src/cmd/project_runner.go:76 +0x65
```

**Version**
 devenv 1.0.2 (x86_64-linux)

domenkozar:
Does it help if you add `pkgs.ncurses` to `packages?

MatzHilven:
> Does it help if you add `pkgs.ncurses` to `packages?
Getting the same error

```
  packages = [
    pkgs.openssl
    pkgs.libyaml
    pkgs.ncurses
];
```

MatzHilven:
I installed `ncurses` globally and it's now returning the following error:
```
panic: exit status 1

goroutine 37 [running]:
github.com/f1bonacc1/process-compose/src/tui.SetupTui({0xe5d8f0?, 0xc0004055e0?}, {0xc00026ff80, 0x2, 0x0?})
        github.com/f1bonacc1/process-compose/src/tui/view.go:482 +0x156
github.com/f1bonacc1/process-compose/src/cmd.startTui({0xe5d8f0, 0xc0004055e0})
        github.com/f1bonacc1/process-compose/src/cmd/project_runner.go:92 +0x179
[...]

PR #1161, opened 2024-04-22
process: fix process-compose settings and other process managers

sandydoo:
I started updating `process.process-compose` to allow overriding individual settings without losing any of the defaults.
But I noticed that, despite the description, those settings are never used.

### What needs to be fixed

| Name | Type | Default | Comment| 
|-------|------|-------|----------|
| `process.process-compose` | `attrs` | `{ version = "0.5"; ... }` |  Claims to control top-level settings. Only used for `tui` and `unix-socket`. |
| `process.process-compose.tui` | `any` | `true` |  Controls the `--tui` CLI option |
| `process.process-compose.unix-socket` |  `any` |`"${config.devenv.runtime}/pc.sock"` | Sets the `--unix-socket` path. |
| `process.process-compose.settings` | `yaml` | `{}` implementation in config | Claims to modify process-specific settings. Controls top-level settings in reality. |
| `process-managers.process-compose.settings.tui` | `bool` | `lib.mkDefault true` | Never used. Not a PC setting AFAIKT, only command-line. |
| `process-managers.process-compose.settings.port` |  `int` | `lib.mkDefault 9999` | Never used. Not a PC settings AFAIKT, only command-line. |

### Other issues

- `tui = false;` breaks process-compose. Fixed by serializing the bool to `true` or `false`, instead of `"1"` and `""`.
- `PC_TUI_ENABLED` doesn't exist in `process-compose`. There's `PC_DISABLE_TUI` though. _edit: nevermind, looks like this exists for backwards-compatibility with previous devenv versions. We should document this. See https://github.com/cachix/devenv/issues/1109._

### Proposal

| Name | Action |
| ------ | ------ |
| `process.process-compose` | Deprecate. If this was previously used for settings, migrate to `process-managers.process-compose.settings`. | 
| `process.process-compose.tui` | Move to `process-managers.process-compose.tui`. |
| `process.process-compose.unix-socket` | Move to `process-managers.process-compose.unixSocket`.  Or `unix-socket` 🤷. |
| `process-managers.process-compose.settings` | Fix description to say that this controls the top-level options |
| `process-managers.process-compose.settings.port` | Move to `process-managers.process-compose.port`. |
| `process-managers.process-compose.settings.tui` | Remove. Migrate if previously used. |
| `process-managers.process-compose.useUnixSocket` | Add. Set to `true` by default and controls the `--use-uds` CLI flag. Or enable when `port = 0`. |


sandydoo:
Relevant example: https://github.com/cachix/devenv/issues/1104#issuecomment-2068232549

src/modules/processes.nix
```diff
@@ -70,7 +70,7 @@ in
               default = true;
             };
             unix-socket = lib.mkOption {
-              type = types.str;
+              type = types.str; # TODO: should be path?
```
thenonameguy, line 73: Paths get copied to the /nix/store. Most of the time you don't want that, especially for devenv runtime state.

sandydoo: TIL! Nice

domenkozar:
We also should update examples and docs.

sandydoo:
An update on the changes so far.

Following the same pattern, we were thinking of renamed `process-managers.<name>.*` to `process.managers.<name>.*`. I'm not entirely sure about this one yet.

### Process manager changes

| Before | After | Comment |
| ------ | ----- | ---------- |
| `process.implementation` | `process.manager.implementation` | |
| `process.after` | `process.manager.after` | |
| `process.before` | `process.manager.before` | |
| `processManagerCommand` | `process.manager.command` | |
| — | `process.manager.args` | Allows modifying the options passed to the command. | 
| `process-managers.<name>.enable` | | Changed to `internal = true`. These are set via `process.manager.implementation`. Added an assertion to check that only one is enabled. |

### Process-compose changes

| Before | After | Comment |
| ------ | ----- |  ---------- |
| `process.process-compose.tui` | `process-managers.process-compose.tui.enable` | |
| — | `process-managers.process-compose.unixSocket.enable` |  Enables the unix socket. Default: `true`. |
| `process.process-compose.unix-socket` | `process-managers.process-compose.unixSocket.path` | |
| `process.process-compose.*` | | Rest dropped. Never used anywhere. | 
| `process-managers.process-compose.settings.port` | `process-managers.process-compose.port` | |

domenkozar:
> Following the same pattern, we were thinking of renaming process-managers.<name>.* to process.managers.<name>.*. I'm not entirely sure about this one yet.

This follows the same pattern as `container.*` and `containers.<name>.*`, same for `processes`, etc.

devenv.local.nix
```diff
@@ -0,0 +1 @@
+{ pre-commit.default_stages = ["manual"]; }
```
domenkozar, line 1: I assume this wasn't meant to be committed

domenkozar (changes requested):

devenv.local.nix
sandydoo: Oh, yeah, I removed it from main already. Looks like it crept in here.

PR #1167, opened 2024-04-23
rust-wasm-cross: fix darwin and aarch64-linux builds

sandydoo:
- Fix the linker issue on macOS. Link to upstream issue.
- Fix the build on `aarch64-linux`, where `wasm-pack` complains that it can't load a build of `wasm-opt`.

### DIFF
--- a/examples/rust-wasm-cross/.test.sh
+++ b/examples/rust-wasm-cross/.test.sh
@@ -4,7 +4,8 @@ cargo --version
 rustc --version
 
 if [[ "$(uname)" == "Darwin" ]]; then
-  echo "$RUSTFLAGS" | grep -- "-L framework=$DEVENV_PROFILE/Library/Frameworks"
+  # See env.RUSTFLAGS in devenv.nix
+  # echo "$RUSTFLAGS" | grep -- "-L framework=$DEVENV_PROFILE/Library/Frameworks"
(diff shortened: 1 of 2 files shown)


examples/rust-wasm-cross/.test.sh
```diff
@@ -4,7 +4,8 @@ cargo --version
 rustc --version
 
 if [[ "$(uname)" == "Darwin" ]]; then
-  echo "$RUSTFLAGS" | grep -- "-L framework=$DEVENV_PROFILE/Library/Frameworks"
+  # See env.RUSTFLAGS in devenv.nix
+  # echo "$RUSTFLAGS" | grep -- "-L framework=$DEVENV_PROFILE/Library/Frameworks"
```
domenkozar, line 8: we should still assert this when not doing wasm

sandydoo: This example is wasm-only though

domenkozar: Ah! Would it make sense to just remove this?

sandydoo: I guess this got copy-pasted over from the regular rust test. 🤷

sandydoo:
Linking the upstream issue, so we can update this when that's resolved: https://github.com/rust-lang/rust/issues/122333

PR #1175, opened 2024-04-29
rust: ensure rust can be disabled when using non-default channel

sandydoo:
Previously, if `channel` was set to anything but `nixpkgs`, the rust toolchain would get added to the shell regardless of the value of `languages.rust.enable`.

Also deduped the logic that figures out which fenix channel to use.

### DIFF
--- a/src/modules/languages/rust.nix
+++ b/src/modules/languages/rust.nix
@@ -80,8 +80,8 @@ in
     };
   };
 
-  config = lib.mkMerge [
-    (lib.mkIf cfg.enable (
+  config = lib.mkIf cfg.enable (lib.mkMerge [
+    (
       let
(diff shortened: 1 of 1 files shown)


src/modules/languages/rust.nix
```diff
@@ -80,8 +80,8 @@ in
     };
   };
 
-  config = lib.mkMerge [
-    (lib.mkIf cfg.enable (
+  config = lib.mkIf cfg.enable (lib.mkMerge [
```
domenkozar, line 83: I'm a bit surprised this works?

sandydoo: Why? Am I missing something? Looks like a common pattern in nixpkgs

domenkozar: Looks good, I just didn't know it's possible to nest it like this!

Issue #1180, opened 2024-05-02
Support extending files_to_watch


soundofspace:
Language modules should be able to extend files to watch. Not sure how this would fit into the current setup, but goal would be that they could register files that are important for direnv and direnvrc script to watch.

Examples would be yarn.lock, requirements.txt... Currently these are only updated when a new shell is opened and might leave the current env in a old/unexpected state. Current workaround is modifying one of the files it watches eg add whitespace to devenv.nix to make sure a refresh is triggered, but this is easy to forget and error prone.

sandydoo:
> Language modules should be able to extend files to watch

Cool idea! We could do something similar to how we handle imports.
https://github.com/cachix/devenv/blob/6ee4b6b00eb33c323a03310429b641aa90f541a4/direnvrc#L118-L122

>  Current workaround is modifying one of the files it watches

You can watch any file you like in `.envrc` with direnv's [`watch_file`](https://github.com/direnv/direnv/blob/master/man/direnv-stdlib.1.md#watch_file-path-path-). Add this to your `.envrc`:
```
watch_file yarn.lock requirements.txt

use devenv
```

> eg add whitespace to devenv.nix to make sure a refresh is triggered

N.B. `touch devenv.nix` is quick way to do the same.

soundofspace:
@sandydoo yea had a similar idea like you did, just not sure how language modules would exposes files to watch, this would somehow require devenv to have already parsed nix files. 

 > You can watch any file you like with direnv's [watch_file](https://github.com/direnv/direnv/blob/master/man/direnv-stdlib.1.md#watch_file-path-path-). Add this to your .envrc:
 
True, and this will trigger a refresh, but this if statement won't trigger, so doesn't really do much
```sh
if [[ ! -e "$profile_rc" || "$need_update" == "1" ]];
then
  log_status "updated devenv shell cache"
else
  log_status "using cached devenv shell"
fi
 ```
 > N.B. touch devenv.nix is a quick way to do the same.
 
 Thought of that as well, but main problem is that this doesn't get checked into git, hence doesn't work for teammembers (which is the main issue).

sandydoo:
> True, and this will trigger a refresh, but this if statement won't trigger, so doesn't really do much

Ah, I forgot we had caching! Maybe `nix_direnv_watch_file`?

> yea had a similar idea like you did, just not sure how language modules would exposes files to watch, this would somehow require devenv to have already parsed nix files.

[...]

PR #1186, opened 2024-05-03
direnv: allow configuring extra files to watch from devenv.nix

sandydoo:
Proof of concept. Based on ideas from https://github.com/cachix/devenv/issues/1180.


direnvrc
```diff
@@ -155,4 +163,13 @@ use_devenv() {
...
     _nix_import_env "$profile_rc"
   fi
+
+  # Watch extra files configured in devenv.nix
+  # This picks up new files and reloads the shell on initial run
+  # TODO: can compare files in files_to_watch with the files we have now
+  if [[ -f "$extra_files_watchlist" ]]; then
+    for file in $(cat "$extra_files_watchlist"); do
+      watch_file "$file"
+    done < "$filename"
```
soundofspace, line 173: Why is `< "$filename"` needed?

soundofspace:
Question on the side: will `_nix_argsum_suffix "${files_to_watch[@]}"` not get expensive depending on the amount of files and size they are?

direnvrc
```diff
@@ -123,6 +123,14 @@ use_devenv() {
     fi
   fi
 
+  local extra_files_watchlist="$flake_dir/.devenv/watchfiles.txt"
+  # Watch extra files configured in devenv.nix
+  if [[ -f "$extra_files_watchlist" ]]; then
+    for file in $(cat "$extra_files_watchlist"); do
+      files_to_watch+=("$file")
```
soundofspace, line 130: Would probably just call `nix_direnv_watch_file` here to prevent `files_to_watch` getting bigger resulting in heavier checksum and amount of profiles getting bigger. But not sure if this is really a problem.

direnvrc
```diff
@@ -155,4 +163,13 @@ use_devenv() {
     log_status "using cached devenv shell"
     _nix_import_env "$profile_rc"
   fi
+
+  # Watch extra files configured in devenv.nix
+  # This picks up new files and reloads the shell on initial run
+  # TODO: can compare files in files_to_watch with the files we have now
+  if [[ -f "$extra_files_watchlist" ]]; then
+    for file in $(cat "$extra_files_watchlist"); do
+      watch_file "$file"
```
soundofspace, line 172: Again I would call files_to_watch here and just centralise deduplication in there.

src/modules/integrations/direnv.nix
```diff
@@ -0,0 +1,25 @@
...
+
+    watchfiles = lib.mkOption {
+      internal = true;
+      type = lib.types.package;
+    };
+  };
+
+  config = {
+    direnv.watchfiles = pkgs.writeText "watchfiles.txt" (lib.concatStringsSep "\n" cfg.watchFiles);
+  };
```
soundofspace, line 24: Pretty new to nix, does not work also?
```suggestion
  cfg.watchfiles = pkgs.writeText "watchfiles.txt" (lib.concatStringsSep "\n" cfg.watchFiles);
```

soundofspace:
@sandydoo Thanks for taking the time to implement this

src/modules/integrations/direnv.nix
sandydoo: Nope, the module system expects a top-level `config` attrset that it then merges together. This is setting `watchfiles` on an attrset literally called `cfg`, which doesn't exist.

direnvrc
soundofspace: Not sure, would it also have separate nix caches? Switching between branches with different versions is not that uncommon and if nix cache could be reused it would be almost instantaneous, which would be nice. On the other hand this use case is not that super common that it needs to be a big concern. I'd say separate profiles if nix cache can be reused, if not just overwrite the existing cache?

Did some research/testing about shasum overhead, and as long as the amount of files and size of these files doesn't get completely out of control, it should have almost no impact. It would even be possible to skip checking files last edit date completely when shasumming everything (but leaving that on is probably nicer).

direnvrc
```diff
@@ -155,4 +167,9 @@ use_devenv() {
     log_status "using cached devenv shell"
     _nix_import_env "$profile_rc"
   fi
+
+  # Watch any extra files configured via `config.direnv.watchFiles` for changes
+  # This picks up new files and reloads the shell on initial run
+  # TODO: deduplicate already watched file paths?
```
soundofspace, line 173: Could easily be done with assosiative arrays but that needs bash 4. Seeing as macos only ships with version 3, that is probably a no go.

I wonder if it would makes sense to move some of this logic to the rust core where you would have much more freedom and power to do what you want. But keeping everything in simple bash has the benefit that the barrier to enter/debug is quite a bit lower.

PR #1187, opened 2024-05-05
feat: Exclude node_modules from .gitignore

k3yss:
This MR excludes the `node_module` folder from the git tree which is generated when working with the `devenv.sh` website.


k3yss:
Already exists on the latest `main`. My bad.

PR #1194, opened 2024-05-09
devenv.isTesting

domenkozar:
We now have `container.isBuilding` and `devenv.isTesting`. I wonder if there's a generalized pattern?


arilotter:
works for me, in my testing, with both process configs & env vars!

Issue #637, opened 2023-06-01
Override default language packages in nixpkgs


tennox:
Thanks for this great project! :heart: 
It's what I dreamt numtide/devshell could be & more :heart_eyes: 

I'm looking for a devenv-native, streamlined way to change nodejs version and also have the package managers use that version, for which the best solution I found (see [discussion in discourse](https://discourse.nixos.org/t/how-to-use-pnpm-with-recent-nodejs/21867/4?u=tennox)) was to use an overlay:

```nix 
{
  overlays = [
    (final: prev: {
      nodejs = cfg.package;
    })
  ];
}
```
And then all added packages use that, too:
```nix
{
  packages = with pkgs; [
    yarn
    nodePackages.pnpm
  ];
}
```
→ `pnpm node --version && yarn node --version` is the selected package

For supporting this in e.g. `languages/javascript.nix` all we'd need is a way to add an overlay in the language module - is there a way?

A possible option (but less great & flexible):
```nix
{
  languages = {
    javascript = {
      package-manager = "pnpm";
    };
  };
}
```

And then in the `languages/javascript.nix` if `packageManager` is `yarn` add `(yarn.override { nodejs = pkgs.nodejs-16_x; })`

Coming from #475 
Vaguely related to #16

shyim:
I really like the idea of having an `package-manager` option in the javascript module and on non `npm` we could use nodejs slim to remove npm.

soundofspace:
@tennox were you able to get these overlays working in devenv? I'm having a similar problem were even some package json dependencies fail because yarn is using the wrong node version. I'm currently still to novice at nix sadly to get it working myself.

soundofspace:
Thanks will give that a go. Does this also work without flakes?
Just like you I came here since this is an amazing project (first tried devbox which is also an amazing project but sadly too basic for our needs). But I'm also worry about the extra complexity this introduces everywhere, as long as it works not really that big of an easy, but things like this make it tougher to sell.

tennox:
> Thanks will give that a go. Does this also work without flakes?

I don't see a way without integrating this into devenv (which I meant to initiate with this issue, maybe I'll craft a PR 🧐)

> I'm also [worried] about the extra complexity this introduces everywhere, as long as it works not really that big of a deal, but things like this make it tougher to sell.

Exactly.

[...]

PR #1202, opened 2024-05-15
Fix: sync nodejs versions

soundofspace:
This fixes yarn using a different nodejs version (partially solves https://github.com/cachix/devenv/issues/637). This doesn't work if yarn pkg is a `nodePackage`. Pnpm also has the same problem since it's a `nodePackage`, not sure how to fix that, seems to use the first node version that is installed.

### DIFF
--- a/src/modules/languages/javascript.nix
+++ b/src/modules/languages/javascript.nix
@@ -218,7 +218,7 @@ in
     ]
     ++ lib.optional cfg.npm.enable (cfg.npm.package)
     ++ lib.optional cfg.pnpm.enable (cfg.pnpm.package)
-    ++ lib.optional cfg.yarn.enable (cfg.yarn.package)
+    ++ lib.optional cfg.yarn.enable (cfg.yarn.package.override { nodejs = cfg.package; })
     ++ lib.optional cfg.bun.enable (cfg.bun.package)
     ++ lib.optional cfg.corepack.enable (pkgs.runCommand "corepack-enable" { } ''
       mkdir -p $out/bin


src/modules/languages/javascript.nix
```diff
@@ -218,7 +218,7 @@ in
     ]
     ++ lib.optional cfg.npm.enable (cfg.npm.package)
     ++ lib.optional cfg.pnpm.enable (cfg.pnpm.package)
-    ++ lib.optional cfg.yarn.enable (cfg.yarn.package)
+    ++ lib.optional cfg.yarn.enable (cfg.yarn.package.override { nodejs = cfg.package; })
```
domenkozar, line 221: do we also need this for pnpm?

soundofspace: Yea pnpm has the same problem but this override only works because yarn package supports it. Pnpm is a nodePackage which sadly doesn't support this (same problem if you would use nodePackage.yarn instead of yarn)

soundofspace: Only way I found to fix that was using overlay to fix all nodejs to specific version

soundofspace: Underlying nix issue: https://github.com/NixOS/nixpkgs/issues/145634

soundofspace: Something like this also works: `cfg.package.pkgs.pnpm`. This installs pnpm with the correct node version, but not sure if all node pkgs support this (tested with `pkgs.nodejs_21;`), will do some more testing.

soundofspace: Doesn't work with nodejs-slim pkg so not sure what the best way forward is here. Supporting overlays seems like the only way to support all different pkgs.

soundofspace: Commented workarounds in original issue aswel so people can still use this until a proper fix is ready. Do you reckon moving away from slim nodejs is a good solution?

Issue #1024, opened 2024-03-21
pre-commit directive no longer works


dz0ny:
**Describe the bug**
Adding a simple `pre-commit` hook yield's in flake build error.

**To reproduce**
```
  pre-commit.hooks = {
    shellcheck.enable = true;
  };
```
Error:

```
error:
       … while calling the 'derivationStrict' builtin

         at /builtin/derivation.nix:9:12: (source not available)

       … while evaluating derivation 'devenv-shell'
         whose name attribute is located at /nix/store/bv68i4l7g6s5sixxm0c9lcjdxhlkg0n5-source/pkgs/stdenv/generic/make-derivation.nix:331:7

       … while evaluating attribute 'DEVENV_PROFILE' of derivation 'devenv-shell'

         at «none»:0: (source not available)

       (stack trace truncated; use '--show-trace' to show the full trace)

       error: attribute 'enabledPackages' missing

       at /nix/store/c8kfbl8b8z0gxzwy0xs29hgw1y13fy2w-source/src/modules/integrations/pre-commit.nix:26:49:

           25|     '';
           26|     packages = [ config.pre-commit.package ] ++ config.pre-commit.enabledPackages;
             |                                                 ^
           27|     enterShell = config.pre-commit.installationScript;
```

**Version**
 
[Paste the output of `$ devenv version` here.
](https://github.com/cachix/devenv/commit/39c4b522ec4522b1c75ca8d686f4f7a9d6044f4e)

domenkozar:
Did you update pre-commit input?

dz0ny:
> Did you update pre-commit input?

Yep, testin 1.0 :)

PR #1209, opened 2024-05-16
feat: support private caches

vetsin:
Fix for #1024 -- enables the usage of the cachix env var.

Note, this does *not* support the netrc, as that would seemingly be more complicated

```
$ ~/projects/devenv/result/bin/devenv test
• Overriding .devenv to .devenv.HbfeKKcHpORj
• Building tests ...
• Using Cachix: devenv, private
• Trusting private.cachix.org on first use with the public key private.cachix.org-1:blabla
✔ Building tests in 11.2s.
```

### DIFF
--- a/devenv/src/command.rs
+++ b/devenv/src/command.rs
@@ -275,13 +275,17 @@ impl App {
                 };
 
                 let mut new_known_keys: HashMap<String, String> = HashMap::new();
+                let client = reqwest::blocking::Client::new();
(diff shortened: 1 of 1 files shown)


vetsin:
@domenkozar fixed the rustfmt, didn't realize i should probably run the tests first :) if you could kindly re-trigger so i could get this downstream

vetsin:
I see failed tests, but seemingly none are because of my change? They seem flaky? Any input here?

PR #1213, opened 2024-05-17
Fix instructions for usage with flakes

sephii:
Running `devenv up` with the current instructions fail with:

```
error: flake 'git+file:///...' does not provide attribute 'packages.x86_64-linux.devenv-up', 'legacyPackages.x86_64-linux.devenv-up' or 'devenv-up
```


sandydoo:
Thanks!

PR #1226, opened 2024-05-23
dotenv: support export declarations in .env files

### LINKED ISSUE #1207: dotenv integration unable to handle .env files with export lines
**Describe the bug**

`.env` files with `export` prefixes on the lines do not get parsed and injected into the environment.

This is a pattern that people use so that they can simply run `source` on `.env`. I came across this while trying to onboard an existing project at work with devenv. 

**To reproduce**

https://gist.github.com/euphemism/25e21226f39ce33ebef940736e2a4167

An empty line is printed instead of `bar`.

**Version**
 
`devenv 1.0.5 (aarch64-darwin)`

sandydoo:
For example, `export FOO=1`.

Fixes #1207.

We can't use non-capturing groups, so this is the best I could come up with.

I couldn't find a spec for the .env format. But some of the parsers support multiline strings, escape quotes, and what not. That's going to be rough to do inside Nix...

### DIFF
--- a/src/modules/integrations/dotenv.nix
+++ b/src/modules/integrations/dotenv.nix
@@ -1,4 +1,4 @@
-{ pkgs, config, lib, self, ... }:
+{ config, lib, self, ... }:
 
 let
   cfg = config.dotenv;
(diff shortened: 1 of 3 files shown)


domenkozar:
https://hexdocs.pm/dotenvy/dotenv-file-format.html does specify some of the options

Most requested would be variable substitution.

PR #1235, opened 2024-05-27
Turn devenv into a library

ankhers:
This is a small example of how to turn the devenv binary into both a library and a binary. The reason this works is that the devenv/Cargo.toml file would allow you to specify one or more binaries, and only a single library. However, because we are using the default values for everything, we do not need to explicitly define these sections. The library entry point will be devenv/src/lib.rs.

In the lib.rs file, we specify that we also want to use the log module. So we define and export this module. So now anything that imports the devenv package can just use devenv::log::* in order to access the pieces defined inside of that module.

There is a problem including the command module since main.rs defines an App struct, and then the command module defines an implementation of that struct. The problem here being that lib.rs does not define the App struct, so when the command module has use crate::App, it is not defined as it is in the case of the devenv binary.


domenkozar:
I assume the next step is we split the cli stuff and App?

ankhers:
Yes, we will need to restructure the existing code so that we can bring in the different pieces that you are looking to use within the run tests project. 

Is he more than happy to help with this effort as well.

Issue #1237, opened 2024-05-28
Cannot create shell with custom certificates 


ptonner:
**Describe the bug**
I'm using `devenv` on a work machine, where custom certificates are injected with Zscaler. I've set `security.pki.certificates` in my nixos config, which has generally addressed the issue (I can run general `nix` commands without issue). But trying to create a shell with devenv does not work:

```sh
> RUST_BACKTRACE=full devenv shell
• Building shell ...
thread 'main' panicked at devenv/src/command.rs:284:26:
Failed to get cache: reqwest::Error { kind: Request, url: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("cachix.org")), port: None, path: "/api/v1/cache/devenv", query: None, fragment: None }, source: hyper::Error(Connect, Ssl(Error { code: ErrorCode(1), cause: Some(Ssl(ErrorStack([Error { code: 167772294, library: "SSL routines", function: "tls_post_process_server_certificate", reason: "certificate verify failed", file: "ssl/statem/statem_clnt.c", line: 1889 }]))) }, X509VerifyResult { code: 2, error: "unable to get issuer certificate" })) }
stack backtrace:
   0:     0xaaaabf90ba54 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h75dbf07dc864c0bd
   1:     0xaaaabf9472f4 - core::fmt::write::h0c7806eb64ac0461
   2:     0xaaaabf8f8f48 - std::io::Write::write_fmt::ha53e993f45f43425
   3:     0xaaaabf90b87c - std::sys_common::backtrace::print::hb1055ea5fa282e57
   4:     0xaaaabf90c38c - std::panicking::default_hook::{{closure}}::h9bc2adbd2b89efe9
   5:     0xaaaabf90c09c - std::panicking::default_hook::h55811f2d21d601c1
   6:     0xaaaabf90c944 - std::panicking::rust_panic_with_hook::h976eed390f682de4
   7:     0xaaaabf90c7fc - std::panicking::begin_panic_handler::{{closure}}::h98a1c1b557857ce3
   8:     0xaaaabf90bca4 - std::sys_common::backtrace::__rust_end_short_backtrace::h70cf8b31899cdba4
   9:     0xaaaabf90c59c - rust_begin_unwind
  10:     0xaaaabf5f6734 - core::panicking::panic_fmt::h0971f4e715269fcb
[...]

PR #1238, opened 2024-05-28
cli: read root certs from the system

sandydoo:
#1237.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -797,6 +797,20 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


sandydoo:
Probably? still an issue in certain corporate environments. Will re-open if we get more complaints.

souhaiebtar:
devenv can't be run in corporate environment because of custom ssl certificate, can we have an option to disable ssl like (--insecure of `curl`), i wanted to use this branch but i didn't know how to

PR #1242, opened 2024-05-29
nixseparatedebuginfod: new module

midchildan:
This creates a new module for [nixseparatedebuginfod](https://github.com/symphorien/nixseparatedebuginfod) to aid with debugging. It launches nixseparatedebuginfod and prepends the server URL to the `DEBUGINFOD_URLS` environment variable.

### DIFF
--- a/tests/nixseparatedebuginfod/devenv.nix
+++ b/tests/nixseparatedebuginfod/devenv.nix
@@ -1,0 +1,28 @@
+{ lib, pkgs, config, ... }:
+
+let
+  cfg = config.services.nixseparatedebuginfod;
+  inherit (pkgs.stdenv) hostPlatform;
+
+  # Use a locally built derivation so that the test wouldn't rely on cache.nixos.org
+  cbin = pkgs.stdenv.mkDerivation {
+    pname = "cbin";
+    version = "1.0";
+    src = ./.;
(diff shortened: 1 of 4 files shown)


tests/nixseparatedebuginfod/devenv.nix
```diff
@@ -0,0 +1,28 @@
...
+    separateDebugInfo = true;
+    meta.mainProgram = "example";
+  };
+in
+lib.mkIf (lib.meta.availableOn hostPlatform cfg.package) {
+  services.nixseparatedebuginfod.enable = true;
+
+  # The Nix store needs to be indexed by nixseparatedebuginfod for debug outputs from local
+  # derivations to be served. This can take a few minutes.
+  process.before = "${lib.getExe cfg.package} --index-only";
```
midchildan, line 22: I added a test that fetches debuginfo from the server, but I'm not sure if this is a good idea. nixseparatedebuginfod needs to index the Nix store before it can serve debuginfos for locally built derivations. This can "take a few minutes" according to the official README. On my machine, it took a little more than a minute so I set the timeout to 2 minutes. The actual time will likely depend on the size of the Nix store and I don't know if 2 minutes is going to be enough, though.

PR #1248, opened 2024-06-01
ci: fix failing tests

sandydoo:
Fixed examples:

- [x] examples/rust-wasm-cross
- [x] examples/phoenix
- [x] examples/python-django
- [ ] examples/opentelemety-collector
        ~won't launch on `macos-13`. firewall issue? Tried disabling firewall, still nada.~
        the readiness_probe on clickhouse is broken on macos13. curl-ing against the http port works. Only happens in GHA 🤷 `clickhouse-client` works as expected.
- [x] examples/supported-languages
       `odin` needs nixpkgs bump for darwin fixes.
        ~`unison-ucm` doesn't have aarch64-linux builds.~
- [x] examples/ruby-on-rails: flaky on macos. timeouts too short? Seems to be working now.

Fixed languages:

- [x] languages/odin made `gdb` and `ols` optional depending on system.

Broken tests:
- [x] `devenv-test-cli`
- [x] `devenv test` 
- [ ] `tests/mysql` fails on macos

Other changes:

- Changed `.setup.sh` to run inside the devenv shell. This helps test manual user setup, like initializing a project and installing dependencies.
- Implement `.patch.sh` script support in `devenv-run-tests`. This runs outside of the shell, before `.setup.sh`. This should be used to patch files, like `devenv.nix`.
- Replaced uses of `example.com` with `localhost`.


sandydoo:
thinking about it some more, the `.patch.sh` is more of a workaround for whatever the underlying issue is. Why does running the setup stuff break in `enterShell`, but not when run before the processes are launched? Race condition?

sandydoo:
Alright, I'm out of steam. The last two failures are on macOS.

1. `examples/opentelemetry-collector` — the clickhouse healthcheck fails to go green on the hosted x86 runner. I used tmate to run it manually, and as a separate process, and it works correctly in those situations. The http health check works as expected as well. Super weird!
2. `tests/mysql` — fails on our self-hosted macos runner. There's a warning about not being able to resolve the hostname. Haven't looked into it too much.

devenv-run-tests/src/main.rs
```diff
@@ -71,30 +71,55 @@ fn run_tests_in_directory(args: &Args) -> Result<Vec<TestResult>, Box<dyn std::e
...
+
+                // Run .setup.sh if it exists
+                if setup_script_path.exists() {
+                    println!("    Running {setup_script}");
+                    // TODO: use as a library
+                    let _ = std::process::Command::new("devenv")
+                        .args(devenv_override)
+                        .args(overrides.clone())
+                        .arg("shell")
+                        .arg("./.setup.sh")
```
domenkozar, line 115: So this is the same as `enterShell`, but it's not racy so that's why it works better?

sandydoo: Exactly. The problem we have now is that this sets things up in `.devenv`, but then `devenv test` uses a temporary directory. We can work around this by librarifying devenv and setting up our own temp dir.

sandydoo:
Needs #1263.

PR #1256, opened 2024-06-08
feat(scripts): allow custom interpreters

### LINKED ISSUE #907: Custom shell for scripts
I like writing my scripts in nushell, but there seems to be no way for `scripts` to emit custom shebang.

I mean, I could write my script exec like this: `"${pkgs.nushell}/bin/nu ./scripts/script-name.nu $@";`, but then I can't easily use custom packages without exposing them to the shell as well. It'd be great to be able to set `scripts.*.shell` or similar.

rawkode:
Closes #907

### DIFF
--- a/src/modules/scripts.nix
+++ b/src/modules/scripts.nix
@@ -1,8 +1,7 @@
-{
-  lib,
-  config,
-  pkgs,
-  ...
+{ lib
+, config
+, pkgs
+, ...
 }:
 
 let
--- a/src/modules/scripts.nix
+++ b/src/modules/scripts.nix
@@ -27,7 +27,7 @@ let
         description = lib.mkOption {
           type = types.str;
           description = "Description of the script.";
-          default = pkgs.bash;
+          default = "";
         };
       };
     }
--- a/src/modules/scripts.nix
+++ b/src/modules/scripts.nix
@@ -18,7 +18,7 @@ let
         package = lib.mkOption {
           type = types.package;
           description = "The package to use to run the script.";
-          default = "";
+          default = pkgs.bash;
         };
         binary = lib.mkOption {
           type = types.str;
--- a/src/modules/scripts.nix
+++ b/src/modules/scripts.nix
@@ -1,22 +1,49 @@
-{ pkgs, lib, config, ... }:
+{
+  lib,
+  config,
+  pkgs,
+  ...
+}:
 
 let
   types = lib.types;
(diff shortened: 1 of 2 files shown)


src/modules/scripts.nix
```diff
@@ -1,22 +1,49 @@
...
-        default = "";
+  scriptType = types.submodule (
+    { config, ... }:
+    {
+      options = {
+        exec = lib.mkOption {
+          type = types.str;
+          description = "Shell code to execute when the script is run.";
+        };
+        interpreter = lib.mkOption {
```
domenkozar, line 18: I feel like we should call it `package` since we're calling it that way everywhere else too.

rawkode: No problem. Are you happy with binary, or is there a convention there too? Bin?

domenkozar:
binary sounds good!

src/modules/scripts.nix
rawkode: Fixed

src/modules/scripts.nix
```diff
@@ -1,22 +1,49 @@
...
+    {
+      options = {
+        exec = lib.mkOption {
+          type = types.str;
+          description = "Shell code to execute when the script is run.";
+        };
+        package = lib.mkOption {
+          type = types.package;
+          description = "The package to use to run the script.";
+          default = "";
```
domenkozar, line 21:
```suggestion
          default = pkgs.bash;
```

domenkozar (changes requested):

rawkode:
Oops. I'm using the wrong formatter I think

domenkozar:
@rawkode seems like you need to run `pre-commit run -a`

rawkode:
@domenkozar I'm learning. Should be sorted now. Thanks for the assist.

domenkozar:
No worries, that's why I'm here to help :)

src/modules/scripts.nix
```diff
@@ -1,22 +1,48 @@
...
+        };
+        binary = lib.mkOption {
+          type = types.str;
+          description = "Override the binary name if it doesn't match package name";
+          default = config.package.pname;
+        };
+        description = lib.mkOption {
+          type = types.str;
+          description = "Description of the script.";
+          default = pkgs.bash;
```
domenkozar, line 30:
```suggestion
          default = "";
```

PR #1261, opened 2024-06-10
devenv: resolve cli arg conflicts

sandydoo:


devenv/src/main.rs
```diff
@@ -70,7 +70,6 @@ struct Cli {
 
     // TODO: --no-clean?
     #[arg(
-        short,
```
sandydoo, line 73: Not sure what to do with this one 🤔

domenkozar: I'd rather remove -c from copy-args

PR #1263, opened 2024-06-10
devenv: split out library

sandydoo:


devenv-run-tests/src/main.rs
```diff
@@ -78,31 +79,36 @@ fn run_tests_in_directory(args: &Args) -> Result<Vec<TestResult>, Box<dyn std::e
...
+                    config,
+                    global_options,
+                    Some(&cwd.join(path)),
+                    Some(tmpdir.as_ref()),
+                    logger.clone(),
+                );
+
+                devenv.create_directories();
+
+                let status = devenv.test();
```
sandydoo, line 107: The next step is to make this nicer to use.

domenkozar:
I'm going to merge https://github.com/cachix/devenv/pull/1194 next and make 1.0.6, then this can go in ASAP :)

package.nix
```diff
@@ -23,5 +23,7 @@ pkgs.rustPlatform.buildRustPackage {
 
   postInstall = ''
     wrapProgram $out/bin/devenv --set DEVENV_NIX ${inputs.nix.packages.${pkgs.stdenv.system}.nix} --prefix PATH ":" "$out/bin:${inputs.cachix.packages.${pkgs.stdenv.system}.cachix}/bin"
+    # TODO: problematic for our library...
```
sandydoo, line 26: We could use [`env!`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rustc-env) to embed a default path to nix in the library. That would remove the runtime requirement.

PR #1267, opened 2024-06-11
fix: execution of devenv version in read only directory

### LINKED ISSUE #1211: Cannot get version without writable directory
**Describe the bug**
one cannot run `devenv version` unless CWD is writable

**To reproduce**

```
cfc0bd017ecc:/tmp$ devenv version
devenv 1.0.5 (aarch64-linux)
cfc0bd017ecc:/tmp$ pwd
/tmp
cfc0bd017ecc:/tmp$ cd /
cfc0bd017ecc:/$ pwd
/
cfc0bd017ecc:/$ devenv version
thread 'main' panicked at devenv/src/main.rs:282:45:
Failed to create .devenv/gc directory: Os { code: 13, kind: PermissionDenied, message: "Permission denied" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

**Version**
 
`devenv 1.0.5 (aarch64-linux)`

k3yss:
resolves  #1211

@sandydoo

### DIFF
--- a/devenv/src/main.rs
+++ b/devenv/src/main.rs
@@ -202,9 +202,6 @@ enum Commands {
     Assemble,
 
     #[clap(hide = true)]
-    CreateDirectories,
-
-    #[clap(hide = true)]
     PrintDevEnv {
         #[arg(long)]
         json: bool,
@@ -271,7 +268,7 @@ struct App {
     has_processes: Option<bool>,
     container_name: Option<String>,
     assembled: bool,
-    dirs_created: Option<bool>,
+    dirs_created: bool,
     // all kinds of paths
     devenv_root: PathBuf,
     devenv_dotfile: PathBuf,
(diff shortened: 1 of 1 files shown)


devenv/src/main.rs
```diff
@@ -201,6 +201,9 @@ enum Commands {
     #[clap(hide = true)]
     Assemble,
 
+    #[clap(hide = true)]
+    CreateDirectories,
+
```
sandydoo, line 206: We don't need a CLI subcommand for this, do we?
```suggestion
```

devenv/src/main.rs
```diff
@@ -268,6 +271,7 @@ struct App {
     has_processes: Option<bool>,
     container_name: Option<String>,
     assembled: bool,
+    dirs_created: Option<bool>,
```
sandydoo, line 274: A `bool` would do the job here. Is there a reason to to wrap it in an `Option`?
```suggestion
    dirs_created: bool,
```

devenv/src/main.rs
```diff
@@ -407,6 +409,7 @@ fn main() -> Result<()> {
             InputsCommand::Add { name, url, follows } => app.inputs_add(&name, &url, &follows),
         },
         // hidden
+        Commands::CreateDirectories => app.create_directories(),
```
sandydoo, line 412:
```suggestion
```

devenv/src/main.rs
```diff
@@ -1079,6 +1082,35 @@ impl App {
         Ok(())
     }
 
+    fn create_directories(&mut self) -> Result<()> {
+        if self.dirs_created != Some(true) {
```
sandydoo, line 1086:
```suggestion
        if !self.dirs_created {
```

devenv/src/main.rs
```diff
@@ -1079,6 +1082,35 @@ impl App {
...
+            let devenv_state_hash = {
+                let mut hasher = sha2::Sha256::new();
+                hasher.update(self.devenv_dotfile.to_string_lossy().as_bytes());
+                let result = hasher.finalize();
+                hex::encode(result)
+            };
+            self.devenv_runtime =
+                Path::new(&self.devenv_tmp).join(format!("devenv-{}", &devenv_state_hash[..7]));
+
+            self.dirs_created = Some(true);
```
sandydoo, line 1109:
```suggestion
            self.dirs_created = true;
```

devenv/src/main.rs
```diff
@@ -288,50 +292,48 @@ fn main() -> Result<()> {
...
         cli,
         config,
         assembled: false,
         has_processes: None,
         logger,
         container_name: None,
-        devenv_root,
-        devenv_dotfile,
-        devenv_dot_gc,
+        dirs_created: None,
```
sandydoo, line 322:
```suggestion
        dirs_created: false,
```

devenv/src/main.rs
```diff
@@ -288,50 +292,48 @@ fn main() -> Result<()> {
...
-        let mut hasher = sha2::Sha256::new();
-        hasher.update(devenv_dotfile.to_string_lossy().as_bytes());
-        let result = hasher.finalize();
-        hex::encode(result)
-    };
-    let devenv_runtime = Path::new(&devenv_tmp).join(format!("devenv-{}", &devenv_state_hash[..7]));
+
+    // using dummy values for initialization
+    let devenv_root: PathBuf = "./".into();
+    let devenv_runtime = Path::new(&devenv_tmp).join(format!("devenv-{}", 1234.to_string()));
```
sandydoo, line 312: Why do we need to fake these? Surely we can fetch the current directory in most situations?

As an aside, `PathBuf::from` is probably the more rusty way of initialising a new pathbuf.

sandydoo (changes requested):

devenv/src/main.rs
k3yss: Yeah, you are right. I was just following similar patterns as `assemble`, fixed everything in the latest commit.

PR #1277, opened 2024-06-13
docs: Add instructions to use cachix on nixos

k3yss:
While compiling devenv on NixOS, I noticed that the instructions for using cachix were only provided for non-NixOS distributions. This PR addresses this issue and adds documentation for NixOS.

### DIFF
--- a/devenv/src/command.rs
+++ b/devenv/src/command.rs
@@ -359,7 +359,23 @@ impl Devenv {
                             {{
                                 cachix.enable = false;
                             }}
+
+                            NOTE: NixOS users can add the following to your configuration.nix as an alternative to the above instructions:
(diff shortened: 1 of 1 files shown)


domenkozar:
If it's NixOS we can check if `/etc/NIXOS` exists and then offer append the extra instructions.

k3yss:
Or we can simply not display the first instructions and only display OS specific instructions:

A simple bash script to demonstrate what I mean
```
if [ -f /etc/NIXOS ]; then
    echo "Running on NixOS"
    # Perform NixOS-specific tasks here
else
    echo "Not running on NixOS"
    # Perform tasks for other systems
fi
```

PR #1278, opened 2024-06-13
chore: bump latest-version in modules

aisamu:
I know it's a silly thing, but my lack of fluency with nix things led me to spend a good 10 minutes trying to figure why devenv wasn't updating 😅 

Thanks for devenv! It's superb!


domenkozar:
We're waiting for devenv 1.0.7 to hit nixpkgs, see https://status.nixos.org/

aisamu:
I see! 
Would it be useful to have a different way to communicate the version of the flake shim then? (when you invoke `devenv --version`, that is)

PR #1281, opened 2024-06-17
Allow caching

dz0ny:
I was using devenv as flake library but later switched to cli, only noticing that is magnitude slower. What is the main reason for disabling caching? 

Without caching:
```
time devenv shell exit
• Building shell ...
✔ Building shell in 5.1s.
0.76s user 0.18s system 13% cpu 7.154 total
```

With caching:
```
time devenv --eval-cache shell exit
• Building shell ...
✔ Building shell in 1.5s.
0.06s user 0.03s system 5% cpu 1.645 total
```

### DIFF
--- a/devenv/src/cli.rs
+++ b/devenv/src/cli.rs
@@ -47,6 +47,9 @@ pub struct GlobalOptions {
     )]
     pub impure: bool,
 
+    #[arg(long, global = true, help = "Use flake cache for evaluation results.")]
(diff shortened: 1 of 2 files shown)


domenkozar:
The main reason is that it brings a lot of complexity and bugs, like caching evaluation errors.

dz0ny:
Ah, well I needed this for CI. Otherwise it was bit too slow.

PR #1298, opened 2024-06-24
Assign PGHOST to runtimeDir as default 

### LINKED ISSUE #1253: Setting postgres `listen_addresses` to `*` also sets `$PGHOST` to `*`
**Describe the bug**
Hello, I was having some issues with a program that tried to connect to postgres in my devenv project.
I have set `listen_addresses` to `*` so that it could be used without needing to use the socket file.
However, this meant that devenv set the `$PGHOST` environment variable to `*`, which made the program I was using try to connect to `*` instead of `0.0.0.0` for example.

This was really confusing, since I did not know that devenv was setting the `$PGHOST`  and it was even more confusing that it would set it to something that was invalid.

**To reproduce**

```nix
{
  services = {
    postgres = {
      enable = true;
      listen_addresses = "*";
    };
  };
}
```

**Expected behavior**
I would expect devenv either to not set the env variable at all or set it to something that was still valid, like `0.0.0.0`.

**Version**
 
Using flakes
devenv: 1.0.5

k3yss:
Fixes #1253

### DIFF
--- a/src/modules/services/postgres.nix
+++ b/src/modules/services/postgres.nix
@@ -282,10 +282,7 @@ in
     packages = [ postgresPkg startScript ];
 
     env.PGDATA = config.env.DEVENV_STATE + "/postgres";
-    env.PGHOST =
-      if cfg.listen_addresses == "" || cfg.listen_addresses == "*"
-      then runtimeDir
-      else cfg.listen_addresses;
+    env.PGHOST = runtimeDir;
     env.PGPORT = cfg.port;
 
     services.postgres.settings = {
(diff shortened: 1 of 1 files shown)


domenkozar:
I'm not sure this is correct, because `*` means all IP addresses, but it's not setting up the socket in `runtimeDir`.

sandydoo:
If `*` binds to all interfaces, then could we use `localhost` or `127.0.0.1` in that case?

domenkozar:
Let's have a test for this

sandydoo (approved):

PR #1302, opened 2024-06-26
fix: --upgrade-dep not available with python < 3.9

### LINKED ISSUE #1289: --upgrade-deps doesn't work with python 3.8
**Describe the bug**
--upgrade-deps doesn't work with python 3.8

**To reproduce**
Create a python environment with python 3.8 and enable venv

**Version**
1.5

k3yss:
The `--upgrade-dep` flag was introduced in python version `3.9`, this PR disables the flag in case of python version lower than the version `3.9`

fixes #1289

### DIFF
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -63,10 +63,14 @@ let
         echo uv venv "$VENV_PATH"
         uv venv "$VENV_PATH"
       ''
-      else ''
+      else if lib.versionAtLeast cfg.version "3.9"  then ''
           echo ${package.interpreter} -m venv --upgrade-deps "$VENV_PATH"
           ${package.interpreter} -m venv --upgrade-deps "$VENV_PATH"
         ''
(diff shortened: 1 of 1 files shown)


src/modules/languages/python.nix
```diff
@@ -63,10 +63,14 @@ let
         echo uv venv "$VENV_PATH"
         uv venv "$VENV_PATH"
       ''
-      else ''
+      else if lib.versionAtLeast cfg.version "3.9"  then ''
           echo ${package.interpreter} -m venv --upgrade-deps "$VENV_PATH"
```
domenkozar, line 67: It would be better to just conditionalize the flag itself

domenkozar (changes requested):

k3yss:
Should be fine now @domenkozar

k3yss:
fixing tests

domenkozar (approved):

PR #1316, opened 2024-07-03
Bug/tailwindcss - Update devenv.nix to Use Tailwind Nix Package

use-the-fork:
The base `devenv.nix` configuration relied on NPX to execute Tailwind CSS commands. However, for users without Node.js installed locally, these commands would fail. This update replaces NPX with the Tailwind package from Nix, ensuring compatibility regardless of local Node.js installations.

### DIFF
--- a/devenv.nix
+++ b/devenv.nix
@@ -35,7 +35,7 @@
 
   processes = {
     docs.exec = "mkdocs serve";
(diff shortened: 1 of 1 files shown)


use-the-fork:
@domenkozar Here is the bug I was talking about. It's hard to work on the dev package since the pre-commit hook uses NPX and my other branches fail to commit without this. 

If this can get reviewed and integrated first, it would be great!

sandydoo (approved):

PR #1319, opened 2024-07-04
devenv: generate and ship manpages

sandydoo:
Usage:
```shell
cargo xtask generate-manpages --out-dir OPTIONAL_PATH
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,6 +3,7 @@ resolver = "2"
 members = [
     "devenv",
     "devenv-run-tests",
+    "xtask",
 ]
 
 [workspace.package]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[alias]
+xtask = "run --package xtask --"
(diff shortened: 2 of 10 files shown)


Cargo.toml
```diff
@@ -3,6 +3,7 @@ resolver = "2"
 members = [
     "devenv",
     "devenv-run-tests",
+    "xtask",
```
domenkozar, line 6: Why not `devenv-generate-manpages`?

sandydoo: Sort of a convention https://github.com/matklad/cargo-xtask?tab=readme-ov-file. You could add other tasks to this crate to avoid slowing down builds of the main package.

sandydoo: Cargo has separate crates for each task and then sets aliases for them in `.cargo/config.toml`. Others use a single crate with subcommands. Same thing.

domenkozar:
We should also make a PR to nixpkgs to generate manpages as part of the next release

PR #1339, opened 2024-07-22
python: Enter directory when creating venv

pinkwah:
This PR does several things in one go:

1. Fixes `libstdc++.so.6` issues due to poetry using unwrapped Python when used in combination with `languages.python.directory`. This was because the `poetry env use` wasn't given the source directory.

2. Fixes poetry creating incorrect shebangs because `languages.python.directory` is a relative path. This is a known issue: https://github.com/python-poetry/poetry/issues/9415

3. Cleans up the bash code by using pushd/popd.

I am a bit uncertain about the use of `pushd`/`popd`. It'd be nice to use them, but they can be replaced with a local variable.


domenkozar:
It seems like `python-venv` example is failing.

pinkwah:
Updated.

PR #1343, opened 2024-07-26
feat: Add option to select jdk with clojure

### LINKED ISSUE #1341: Make it easy to change JDK version for languages.clojure
For languages.java and languages.scala it seems to be quite simple to change the JDK version. 

However, there's no way to do that for Clojure, and using `languages.java.jdk.package` does not appear to affect the jdk version used in Clojure:

```
$ java -version
openjdk version "22" 2024-03-19
OpenJDK Runtime Environment (build 22+36-nixos)
OpenJDK 64-Bit Server VM (build 22+36-nixos, mixed mode, sharing)
$ clojure --version
Clojure CLI version 1.11.3.1463
$ clj
Clojure 1.11.1
user=> (System/getProperty "java.version")
"21"
```

k3yss:
This PR adds the functionality of selecting the `jdk` version for `clojure` programming language using overrides.

Fixes #1341

### DIFF
--- a/src/modules/languages/clojure.nix
+++ b/src/modules/languages/clojure.nix
@@ -6,26 +6,17 @@ in
 {
   options.languages.clojure = {
     enable = lib.mkEnableOption "tools for Clojure development";
-    jdk.package = lib.mkOption {
-      type = lib.types.package;
-      example = lib.literalExpression "pkgs.jdk8";
-      default = pkgs.jdk;
-      defaultText = lib.literalExpression "pkgs.jdk";
-      description = ''
-        The JDK package to use.
-        This will also become available as `JAVA_HOME`.
-      '';
-    };
   };
 
   config = lib.mkIf cfg.enable {
     packages = with pkgs; [
       (clojure.override {
-        jdk = cfg.jdk.package;
+        jdk = config.languages.java.jdk.package;
       })
       clojure-lsp
     ];
+    languages.java.enable = true;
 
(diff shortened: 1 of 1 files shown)


src/modules/languages/clojure.nix
```diff
@@ -6,12 +6,26 @@ in
...
   config = lib.mkIf cfg.enable {
     packages = with pkgs; [
-      clojure
+      (clojure.override {
+        jdk = cfg.jdk.package;
+      })
       clojure-lsp
     ];
+
+    env.JAVA_HOME = cfg.jdk.package.home;
```
domenkozar, line 29: I wonder if we should instead set `languages.java.enable = true;` and then use `config.languages.java.package`?

k3yss: Yeah, I think that's better. Changed it in the latest commit.

domenkozar: java already sets `JAVA_HOME`

src/modules/languages/clojure.nix
```diff
@@ -10,8 +10,13 @@ in
...
     packages = with pkgs; [
-      clojure
+      (clojure.override {
+        jdk = config.languages.java.jdk.package;
+      })
       clojure-lsp
     ];
+    languages.java.enable = true;
+
+    env.JAVA_HOME = config.languages.java.jdk.package.home;
```
domenkozar, line 20:
```suggestion
```

PR #1352, opened 2024-07-30
Nix language support improvements, including smaller Cachix

therealpxc:
**This fix saves 7 GB** in the closure size of my devenv projects which have Nix language support enabled.

Background: Currently, with `languages.nix.enable = true;`, all outputs of the `cachix` package in Nixpkgs get installed. Unlike the Cachix package provided by the [cachix/cachix](https://github.com/cachix/cachix) flake, this is a massive multi-output derivation that includes a huge Haskell runtime with libs in `.out` and more than a gigabyte of library documentation in `.doc`.

The original issue here raises some questions for me about how devenv profiles are constructed and whether they do or should honor `meta.outputsToInstall`-- maybe there's something to think about there, too. (The Cachix package in nixpkgs has only `"bin"` in `meta.outputsToInstall`.)

Anyway, this fix works now and is a huge quality of life improvement for me!

### DIFF
--- a/src/modules/cachix.nix
+++ b/src/modules/cachix.nix
@@ -1,4 +1,4 @@
-{ lib, config, ... }:
+{ pkgs, lib, config, ... }:
 let
   cfg = config.cachix;
 in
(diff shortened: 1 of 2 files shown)


domenkozar:
Maybe we should just use `devenv.bin`?

Issue #1355, opened 2024-07-31
[Rust] darwinFlags are causing problem with no_std crates


tianyaochou:
**Describe the bug**
When compiling no_std crate, linking stage throws error regarding the `darwinFlags`.

**To reproduce**

https://gist.github.com/tianyaochou/3ce69267f6ab91b50214b89ac57b6ab4

```
cargo build --release
   Compiling core v0.0.0 (/nix/store/ysj8iy32fnsbag8qh6jlz7i4q6l5hfam-rust-mixed/lib/rustlib/src/rust/library/core)
   Compiling compiler_builtins v0.1.109
   Compiling esp-println v0.10.0
   Compiling portable-atomic v1.7.0
   Compiling riscv v0.11.1
   Compiling esp32c6 v0.15.0
   Compiling esp-hal v0.19.0
   Compiling esp-backtrace v0.13.0
   Compiling esp-test v0.1.0 (/Users/tianyaochou/Projects/Playgrounds/esp-playground)
   Compiling rustc-std-workspace-core v1.99.0 (/nix/store/ysj8iy32fnsbag8qh6jlz7i4q6l5hfam-rust-mixed/lib/rustlib/src/rust/library/rustc-std-workspace-core)
   Compiling critical-section v1.1.2
   Compiling embedded-hal v1.0.0
   Compiling nb v1.1.0
   Compiling gcd v2.3.0
   Compiling stable_deref_trait v1.2.0
   Compiling log v0.4.22
   Compiling vcell v0.1.3
   Compiling embedded-can v0.4.1
   Compiling fugit v0.3.7
   Compiling embedded-hal-nb v1.0.0
   Compiling embedded-dma v0.2.0
   Compiling bytemuck v1.16.1
   Compiling strum v0.26.3
   Compiling rand_core v0.6.4
   Compiling enumset v1.1.5
   Compiling bitfield v0.15.0
   Compiling void v1.0.2
   Compiling esp-riscv-rt v0.9.0
   Compiling bitflags v2.6.0
   Compiling cfg-if v1.0.0
error: linking with `rust-lld` failed: exit status: 1
  |
[...]

PR #1356, opened 2024-07-31
Remove darwinFlags in rust module

tianyaochou:
Remove `darwinFlags` to fix linking error when compiling no_std crate. Closing #1355 

Only tested with Developer Frameworks already installed in the system.

I am not sure if this change will break those who only rely on Developer Frameworks in Nixpkgs and don't have Developer Frameworks installed outside Nix.

### DIFF
--- a/examples/rust-wasm-cross/devenv.nix
+++ b/examples/rust-wasm-cross/devenv.nix
@@ -27,9 +27,4 @@
   ] ++ lib.optionals pkgs.stdenv.isDarwin (with pkgs.darwin.apple_sdk; [
     frameworks.Security
   ]);
-
-  # macOS workaround:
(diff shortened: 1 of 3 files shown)


sandydoo (approved):
I've updated our rust tests and examples to work with this change.

domenkozar:
What if you need to compile against a framework library on macos?

sandydoo:
@domenkozar, I tested this out in a fresh VM. 
Nix already adds the `-iframework` CFLAG for us. 
The `-L framework` rust flag doesn't seem necessary. I tested a bunch of random OSS projects that link to frameworks — they compile without issues. Is this around from the pre-1.0 era when we didn't have setup hooks?

PR #1358, opened 2024-08-01
python: when enabling uv and venv, version is not respected

gotcha:
@domenkozar This does work for me.

However, there are no tests.

Further, being a devenv newbie, I have no idea how many best practices I am breaking.

### DIFF
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -62,8 +62,8 @@ let
         [ -f "${config.env.DEVENV_STATE}/poetry.lock.checksum" ] && rm ${config.env.DEVENV_STATE}/poetry.lock.checksum
       ''}
       ${if cfg.uv.enable then ''
-        echo uv venv "$VENV_PATH"
-        uv venv "$VENV_PATH"
+        echo uv venv -p ${package.interpreter} "$VENV_PATH"
+        uv venv -p ${package.interpreter} "$VENV_PATH"
       ''
       else ''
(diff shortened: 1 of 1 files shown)


src/modules/languages/python.nix
```diff
@@ -62,8 +62,8 @@ let
         [ -f "${config.env.DEVENV_STATE}/poetry.lock.checksum" ] && rm ${config.env.DEVENV_STATE}/poetry.lock.checksum
       ''}
       ${if cfg.uv.enable then ''
-        echo uv venv "$VENV_PATH"
-        uv venv "$VENV_PATH"
+        echo uv venv -p ${cfg.package.interpreter} "$VENV_PATH"
```
domenkozar, line 65: Should it be package.interpeter?

src/modules/languages/python.nix
```diff
@@ -48,7 +48,7 @@ let
 
     VENV_PATH="${config.env.DEVENV_STATE}/venv"
 
-    profile_python="$(${readlink} ${package.interpreter})"
+    profile_python="$(${readlink} ${cfg.package.interpreter})"
```
domenkozar, line 51: Why is this needed?

domenkozar (changes requested):

src/modules/languages/python.nix
gotcha: Thanks for the pointer. And yes, it should be ``package.interpreter``

src/modules/languages/python.nix
gotcha: Right, this is useless.

gotcha:
@domenkozar Thanks for merging.

Where should I look if I would want to add a test ?

domenkozar:
tests folder

PR #1366, opened 2024-08-08
fix: don't delete runtime dir on process down

### LINKED ISSUE #1270: second `devenv up` always crashes with overmind
**Describe the bug**

When I run `devenv up`, stop the process, and then run `devenv up` again it crashes with the following error message:

byNjf-VR

**To reproduce**

```
# Make a new ihp project
nix run nixpkgs#ihp-new test

cd test

# Open flake.nix and enable overmind by adding the following to line 40:
# process.implementation = "overmind";

# First start
devenv up

# After 30 secs stop the process using CTRL C

# Second start
devenv up

# Now the error happens:
# overmind: listen unix /tmp/devenv-d8acd82/overmind.sock: bind: no such file or directory
```

**Version**
 
`devenv: 1.0.5` via flakes 

```
devenv.url = "github:cachix/devenv?rev=800f19d1b999f89464fd8e0226abf4b3b444b0fa";
```

sandydoo:
Fixes #1270.

@domenkozar, should we be deleting the runtime dir when taking down processes? Any good reason to do this?

There are a few issues with this line:

1. It relies on trapping a signal. In the case of process-compose, I think the reason it's immune from this failure is because the trap is never triggered. We need to rethink how we run post-process actions.
2. We create the runtime dir in the CLI, but try to delete it in the shell script. This sounds like a bug magnet to me.
3. There was an issue a while back where redis would crash on shutdown because of a race condition here.

### DIFF
--- a/src/modules/processes.nix
+++ b/src/modules/processes.nix
@@ -146,7 +146,6 @@ in
         wait $backgroundPID
         ${config.process.after}
         echo "Processes stopped."
-        rm -rf ${config.devenv.runtime}
       }
 
       trap down SIGINT SIGTERM


domenkozar:
Maybe instead we should delete it on startup? To start fresh?

sandydoo:
> Maybe instead we should delete it on startup? To start fresh?

do we care?

mpscholten:
Thanks ❤️

Issue #1212, opened 2024-05-17
Is there a way to modify pg_hba.conf from within the declaration?


ElrohirGT:
Hi! I've been trying to develop a serverless API using SAM and would like it to connect to the postgres that devenv services wakes up. Long story short, I need the postgres to accept connections from all IP's, but this can only be changed from within pg_hba.conf. How would I go about defining this inside the nix file?

Here's the error it gives when trying to connect to it:
![image](https://github.com/cachix/devenv/assets/45268815/22fef128-e8c7-4ff0-82f8-026c9f3ad652)

ElrohirGT:
In the end I couldn't find a way to configure the pg_hba.conf directly from the configuratino that devenv offers. So I created a helper service instead whose whole job is to copy my custom pg_hba.conf file into the correct directoy that devenv creates. It has the caveat that I need to restart the DB service manually to make the changes take effect but it's good enough for my usecase which is running it manually in local. Here's the code for this service:

flake.nix
```nix
processes = {
              pg_setup = {
                exec = "cat pg_hba.conf > ./.devenv/state/postgres/pg_hba.conf";
                process-compose = {
                  depends_on = {
                    postgres = {
                      condition = "process_healthy";
                    };
                  };
                };
              };
            };
```
pg_hba.conf
```text
# TYPE  DATABASE        USER            ADDRESS                  METHOD
host	       all                       all                  0.0.0.0/0                     trust
```

This is a hack and I don't like having to restart the service manually because that wrecks all chances of having this in CI but it's a workaround for local development.

PR #1368, opened 2024-08-08
feat: Added custom option and config

ElrohirGT:
This PR resolves (https://github.com/cachix/devenv/issues/1212) by adding a configuration option that enables the user to start the PostgreSQL server with a custom pg_hba.conf file, this file is primarily used to enable or disable which IPs get to connect to the server both locally (from UNIX sockets) or from the network (using TCP/IP).

I had a "Fine I'll do it myself moment" and honestly after I saw the code it wasn't really that hard.

### DIFF
--- a/src/modules/services/postgres.nix
+++ b/src/modules/services/postgres.nix
@@ -85,8 +85,8 @@ let
 
   configFile = pkgs.writeText "postgresql.conf" (lib.concatStringsSep "\n"
     (lib.mapAttrsToList (n: v: "${n} = ${toStr v}") cfg.settings));
-  setupPgHbaFileScript = if cfg.customPgHbaFile != null then
-    ''cp ${cfg.customPgHbaFile} "$PGDATA/pg_hba.conf"'' else "";
+  setupPgHbaFileScript = if cfg.hbaConf != null then
+    ''cp ${cfg.hbaConf} "$PGDATA/pg_hba.conf"'' else "";
   setupScript = pkgs.writeShellScriptBin "setup-postgres" ''
     set -euo pipefail
     export PATH=${postgresPkg}/bin:${pkgs.coreutils}/bin
@@ -282,8 +282,8 @@ in
       '';
     };
 
-    customPgHbaFile = lib.mkOption {
-      type = types.nullOr types.path;
+    hbaConf = lib.mkOption {
+      type = types.nullOr types.str;
       default = null;
       description = ''
         The path to a custom pg_hba.conf file to copy into the postgres installation. This
@@ -291,7 +291,7 @@ in
         you're using flakes remember to add the file to version control!
       '';
       example = lib.literalExpression ''
-        ./my-custom/directory/to/pg_hba.conf
+        "${./my-custom/directory/to/pg_hba.conf}"
       '';
     };
   };
--- a/src/modules/services/postgres.nix
+++ b/src/modules/services/postgres.nix
@@ -85,8 +85,11 @@ let
 
   configFile = pkgs.writeText "postgresql.conf" (lib.concatStringsSep "\n"
     (lib.mapAttrsToList (n: v: "${n} = ${toStr v}") cfg.settings));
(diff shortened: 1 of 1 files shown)


src/modules/services/postgres.nix
```diff
@@ -276,6 +281,19 @@ in
         CREATE ROLE bar;
       '';
     };
+
+    customPgHbaFile = lib.mkOption {
```
domenkozar, line 285: I'd name it `hbaConf` and make it accept a string, for consistency with other options.

domenkozar:
> I had a "Fine I'll do it myself moment" and honestly after I saw the code it wasn't really that hard.

:muscle: :muscle: :muscle:

src/modules/services/postgres.nix
ElrohirGT: Your right! Done.

src/modules/services/postgres.nix
```diff
@@ -85,6 +85,8 @@ let
 
   configFile = pkgs.writeText "postgresql.conf" (lib.concatStringsSep "\n"
     (lib.mapAttrsToList (n: v: "${n} = ${toStr v}") cfg.settings));
+  setupPgHbaFileScript = if cfg.hbaConf != null then
+    ''cp ${cfg.hbaConf} "$PGDATA/pg_hba.conf"'' else "";
```
domenkozar, line 89: You'll need to use `pkgs.writeText here to get it into Nix store.

src/modules/services/postgres.nix
```diff
@@ -276,6 +281,19 @@ in
...
       '';
     };
+
+    hbaConf = lib.mkOption {
+      type = types.nullOr types.str;
+      default = null;
+      description = ''
+        The path to a custom pg_hba.conf file to copy into the postgres installation. This
+        allows for custom connection rules that you want to establish on the server. If
+        you're using flakes remember to add the file to version control!
```
domenkozar, line 291: The flake remark no longer applies.

src/modules/services/postgres.nix
```diff
@@ -276,6 +281,19 @@ in
...
+      '';
+      example = lib.literalExpression ''
+        "${./my-custom/directory/to/pg_hba.conf}"
```
domenkozar, line 294: Needs to use `builtins.readFile`

domenkozar (changes requested):

src/modules/services/postgres.nix
ElrohirGT: Isn't just putting the file path enough to copy it into the Nix store? From what I understand it copies the file into a path in the nix store and then just references it.

Source: https://stackoverflow.com/questions/43850371/when-does-a-nix-path-type-make-it-into-the-nix-store-when-converted-to-a-string

domenkozar: Yeah but it expect it a string now, so you don't want the contents of pg_hba.conf to be a nix store path, but the contents of it.

ElrohirGT: Gotcha, ok! I'll do the changes in a minute

domenkozar:
Need to run `pre-commit run -a`.

domenkozar:
That's strange. You cab just run nixpkgs-fmt on the file you change as a workaround

ElrohirGT:
Gotcha! [There's two nixfmt](https://search.nixos.org/packages?channel=unstable&from=0&size=50&sort=relevance&type=packages&query=nixfmt) on the store, I tried both. Decided on the one with the less amount of changes to the rest of the file.

domenkozar:
We're using nixpkgs-fmt

ElrohirGT:
> We're using nixpkgs-fmt

haha gotcha, I'm dumb, I thought you were refering to nixfmt. Fixed it! Now it should be on the correct format.

ElrohirGT:
Hmmm... weird. I don't know why the tests fail on Mac ARM-64 but I can confirm that using this code I can run it successfully on my project.

domenkozar:
I believe that's an error on our side.

Issue #1367, opened 2024-08-08
Being able to build smaller containers


blackheaven:
I'm trying to define some simple containers:

```nix
  containers.blue = {
    name = "blue";
    copyToRoot = pkgs.buildEnv {
      name = "image-root";
      paths = [ blue ];
      pathsToLink = [ "/bin" ];
    };
    entrypoint = [ "/env/bin/blue" ];
  };
```

`blue` being a haskell.nix derivation

It creates a first layer with my derivation (~420MiB) and adds a layer with many things (coreutils-full, bashInteractive, su, etc.) which weight 15 GiB.

We should be able to passe the final derivation, or to disable extra layers.

domenkozar:
Fully agreed, someone needs to look into container.nix and expose a knob to disable the layering.

therealpxc:
> It creates a first layer with my derivation (~420MiB) and adds a layer with many things (coreutils-full, bashInteractive, su, etc.) which weight 15 GiB.

> 15 GiB

I don't think [those packages](https://github.com/cachix/devenv/blob/25e48e98c19e9424d517a144db0d31a89536bf94/src/modules/containers.nix#L105) you name can be the culprit. They have really small closures! Together they're probably only like 100 MiB. Something weirder is going on.

domenkozar:
https://github.com/cachix/devenv/pull/1375

ppenguin:
I was having the same issue when trying to build a production container which includes a simple binary produced by `buildGoModule`.  (`15GB` still seems like there's something else wrong too though, maybe you can get "closer to reasonable" when doing like below with the `isBuilding` logic)

My use case:
1. existing `Makefile` powered build of a `go` binary including private go module dependencies
2. desire to use `devenv` as a "one stop shop" to replace [my own hacks](https://github.com/ppenguin/nix-dev-imgs) and avoid implementing extending these hacks with "production image" functionality.

Immediate problems not (yet) solved with `devenv`:
1. Can't add binary artifacts due to [a bug](https://github.com/cachix/devenv/issues/1334) and the fact that you'd have to stage them (a la `flake`) for them to be found, which I want to avoid (no binary artifacts in git!)
2. Adding the artifact as a nix package is unattractive:
   1. forces me to make a `nix` package for my artifact
   2. which is non-trivial if you want to use `buildGoModule` (due to private dependency/vendoring/sandboxing nightmares => the only way I could make it work is to use `go mod vendor` but for this I have to add `vendor` to git as well)
[...]

PR #1375, opened 2024-08-10
feat: add flag to disable containers tooling (#1367)

blackheaven:


domenkozar:
Nice! Any chance to separate styling changes from the behavior change?

blackheaven:
Actually it's enforced by the pre-commit hook.

I can try to make a first commit with only formatting, would it be okay?

domenkozar:
That would be perfect, thank you!

blackheaven:
done :)

sandydoo:
> Actually it's enforced by the pre-commit hook.

Are you sure it's not your editor's tooling doing the formatting? This doesn't look like `nixpkgs-fmt` to me. If it really is, I'll have to investigate why our hooks are broken.

blackheaven:
> @blackheaven, are you sure it's not your editor's tooling doing the formatting? This doesn't look like `nixpkgs-fmt` to me. If it really is, I'll have to investigate why our hooks are broken.

I think my language server formatted twice, anyway, it's fixed now.

domenkozar:
The main reason why I haven't merged this yet is because I'm not convinced what the API should look like.

Maybe we should add something like `containers.<name>.contents = [];` that would imply to build the container from scratch?

src/modules/containers.nix
```diff
@@ -135,17 +135,19 @@ let
...
+      }
+      (if cfg.isDev then {
+        Env = lib.mapAttrsToList (name: value: "${name}=${toString value}")
+          config.env ++ [ "HOME=${homeDir}" "USER=${user}" ];
+        Entrypoint = cfg.entrypoint;
+        Cmd = [ cfg.startupCommand ];
+      } else
+        { })
+    ];
   };
```
Mic92, line 151: Nitpick: A bit easier to read because it uses less control flow and we don't have conflicting keys here.

```suggestion
    config = {
      User = "${user}";
      WorkingDir = "${homeDir}";
    } // lib.optionalAttrs cfg.isDev {
      Env = lib.mapAttrsToList (name: value: "${name}=${toString value}")
          config.env ++ [ "HOME=${homeDir}" "USER=${user}" ];
      Entrypoint = cfg.entrypoint;
      Cmd = [ cfg.startupCommand ];
   };
```

blackheaven:
@Mic92 Thank you so much!

blackheaven:
> Maybe we should add something like `containers.<name>.contents = [];` that would imply to build the container from scratch?

Should we check that for other parameters? (i.e. `perms`, etc.)

domenkozar:
@sandydoo can you check this one? :eyes:

domenkozar:
Closing as outdated.

PR #1378, opened 2024-08-11
fix: redis should listen on unix socket when port is 0

### LINKED ISSUE #1068: Redis service should listen on unix socket when `port = 0`
Current workaround

```
  services.redis.port = 0;
  services.redis.extraConfig = ''
    unixsocket redis.sock
    unixsocketperm 700
  '';
  env.REDISSOCKET = "${config.env.REDISDATA}/redis.sock";
```

frectonz:
Fixes #1068 

This PR configures the redis service to listen on a unix socket when the specified port is 0. The socket file will be stored in `DEVENV_RUNTIME`. A new environment variable `REDIS_UNIX_SOCKET` will also be injected into the environment. It points to the socket file, so that users can easily connect to to the service like this.

```sh
redis-cli -s $REDIS_UNIX_SOCKET
```

--------

_`devenv` is the main tool that got me introduced to the nix world, it is a really useful tool, thanks for making it._


domenkozar:
Looks great! Can you move the test into tests?

frectonz:
I have moved it and `devenv-run-tests --only redis-socket tests` succeeds on my machine.

src/modules/services/redis.nix
```diff
@@ -67,14 +74,15 @@ in
       cfg.package
     ];
 
+    env.REDIS_UNIX_SOCKET = REDIS_UNIX_SOCKET;
```
domenkozar, line 77: I wonder if we should set this on non-zero port?

frectonz: Yeah, you're right. It is only needed when port is zero.

I don't know what the right way to conditionally set an environment variable would be, would this work.

```nix
env.REDIS_UNIX_SOCKET = if cfg.port == 0 then REDIS_UNIX_SOCKET else null;
```

frectonz: I went with `optionalAttrs` instead of setting it to `null`.

frectonz:
I don't think the failing tests are related to my changes. Am i wrong?

PR #1379, opened 2024-08-12
feat: add a `--quiet` flag to disable logging

### LINKED ISSUE #435: Silent/Quiet option when opening a new terminal
Hello 👋, very excited about this project!

Could we add an ENV variable or an option to suppress the `Building shell ...` and other progress reports when opening a new terminal? Ideally, I'd like to only see errors reported.

The related code seems here: https://github.com/cachix/devenv/blob/1aa3dbbf745f50e77aadc016d124fed3ca2dc9be/src/devenv.nix#L59-L65

frectonz:
Fixes #435 

This is useful for people who don't want to clutter their terminal with logs. This will also be useful in CI environments for people who want a cleaner log. For example a user can enter a shell without seeing any logs using the following commad.
```sh
devenv -q shell
```

### DIFF
--- a/devenv/src/cli.rs
+++ b/devenv/src/cli.rs
@@ -23,7 +23,13 @@ pub struct GlobalOptions {
     #[arg(short, long, global = true, help = "Enable debug log level.")]
     pub verbose: bool,
 
-    #[arg(short, long, global = true, help = "Disable all logs")]
+    #[arg(
+        short,
+        long,
+        global = true,
+        conflicts_with = "verbose",
+        help = "Disable all logs"
+    )]
     pub quiet: bool,
 
     #[arg(short = 'j', long,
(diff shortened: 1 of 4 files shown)


devenv/src/cli.rs
```diff
@@ -23,6 +23,9 @@ pub struct GlobalOptions {
     #[arg(short, long, global = true, help = "Enable debug log level.")]
     pub verbose: bool,
 
+    #[arg(short, long, global = true, help = "Disable all logs")]
```
sandydoo, line 26: Let's tell the parser that this conflicts with `--verbose`:
```suggestion
    #[arg(short, long, global = true, conflicts_with = "verbose", help = "Disable all logs")]
```

sandydoo (changes requested):
Thanks!

devenv/src/cli.rs
frectonz: Added

PR #1388, opened 2024-08-16
Fix maven failing to build with `function 'anonymous lambda' called with unexpected argument 'jdk'`

MoritzR:
On nixpkgs, the `jdk` argument has been renamed to `jdk_headless`, currently causing the devenv to fail to build maven with:
```
error: function 'anonymous lambda' called with unexpected argument 'jdk'

at /nix/store/pwwns4fq736wqpmf5xfsxds4kh4f580w-source/pkgs/by-name/ma/maven/package.nix:1:1:
```

See [this commit on nixpkgs](https://github.com/NixOS/nixpkgs/commit/ea0bc3224593ddf7ac6c702c7acb6c89cf188f0f#diff-b3af06f6694c5996094cf6cd4db45398f256ff6e2d56d9a7b1b275092badd7d9)

### DIFF
--- a/docs/reference/options.md
+++ b/docs/reference/options.md
@@ -2649,7 +2649,7 @@ package
 
 
 *Default:*
-` "pkgs.maven.override { jdk = cfg.jdk.package; }" `
+` "pkgs.maven.override { jdk_headless = cfg.jdk.package; }" `
 
 *Declared by:*
  - [https://github.com/cachix/devenv/blob/main/src/modules/languages/java.nix](https://github.com/cachix/devenv/blob/main/src/modules/languages/java.nix)
--- a/docs/supported-languages/java.md
+++ b/docs/supported-languages/java.md
@@ -128,4 +128,4 @@ package
 
 
 *Default:*
-` "pkgs.maven.override { jdk = cfg.jdk.package; }" `
+` "pkgs.maven.override { jdk_headless = cfg.jdk.package; }" `
--- a/src/modules/languages/java.nix
+++ b/src/modules/languages/java.nix
@@ -21,7 +21,7 @@ in
       enable = mkEnableOption "maven";
       package = mkOption {
         type = types.package;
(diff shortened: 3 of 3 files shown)


domenkozar:
We still to support the old `jdk`, so we'll need to use `builtins.functionArgs` to decide which one to use.

MoritzR:
> We still to support the old jdk, so we'll need to use builtins.functionArgs to decide which one to use.

Can you help me work out how this is used? @domenkozar 
In [my last commit](https://github.com/cachix/devenv/pull/1388/commits/156ec19244f9d7a731a36ec3d0b98c0a9a420f20) I tried to push that change, but it fails for me with `error: 'functionArgs' requires a function`.
On [this page](https://ryantm.github.io/nixpkgs/using/overrides/) I read that `The function override is usually available`.
So I would assume that something like `builtins.functionArgs pkgs.maven.override;` works.

domenkozar:
You'll need to import the functrion yourself from the path where it's defined in nixpkgs

sandydoo:
`pkgs.maven.override.__functionArgs` gets you the same output as `builtins.functionArgs`. I think it should be fine to use in this case.

MoritzR:
> `pkgs.maven.override.__functionArgs` gets you the same output as `builtins.functionArgs`. I think it should be fine to use in this case.

I am not sure what the difference is, but this worked for me nicely. I pushed an update.

sandydoo:
> I am not sure what the difference is, but this worked for me nicely. I pushed an update.

Nix voodoo. `override` isn't exactly a function, but rather a set `{ __functionArgs = {...}; __functor = <lambda ...>; }`. The functor is a whole story in its own, but in this case the point is to attach metadata (`__functionArgs`) to the function `__functor` to preserve reflection, allowing us to inspect the arguments of the function after it's been curried and that information has been lost. The native `builtin.functionArgs` doesn't understand this.

Turns out there's a library function that works in both cases: [`lib.functionArgs`](https://noogle.dev/f/lib/functionArgs). I've updated the code to use it.

PR #1390, opened 2024-08-16
feat: Expose devenv options for nixd

k3yss:
This PR exposes options for modules which is required if we want to support nixd lsp natively. We would need this before https://github.com/nix-community/nixd/pull/584 is merged.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -104,6 +104,7 @@
         {
           default = self.packages.${system}.devenv;
           devenv = mkPackage pkgs;
+          devenv-module-options = evaluatedModules;
           devenv-docs-options = options.optionsCommonMark;
           devenv-docs-options-json = options.optionsJSON;
           devenv-generate-individual-docs =


domenkozar:
We need to expose the options from .devenv.flake.nix, so that local options are picked up.

This way if users define local options they are part of auto-complete as well.

domenkozar:
Going another route

sebb3:
Is something like this being implemented? would be really helpful with some autocompletion!

k3yss:
@sebb3 There is ongoing work in https://github.com/cachix/devenv/tree/work/keys/lsp. I got pretty busy with my university the last few weeks, going to pick it up again.

oscar-schwarz:
I stumbled upon this closed PR in hope to find an easy way to get devenv option hints for nixd. Well it wasn't easy (I spend a whole day on this to procrastinate my university work, @k3yss has much better discipline there). But I found some solutions which will help anyone having the same problem as me as long as the official solution isn't done yet.

If we just want the basic devenv options without any additions we can use `lib.evalModules` (how its also done in the official flake.nix file of devenv)
```nix
let 
  devenvFlake = (builtins.getFlake "/path/to/your/flake").inputs.devenv;
  # if you don't want to add it to your own flake as an input:
  # devenvFlake = builtins.getFlake "github:cachix/devenv?rev=3febc91939aea65bdff8850f026443afb6b6b22f";
  pkgs = import devenvFlake.inputs.nixpkgs {};
in
# This is copied and adapted from: 
# https://github.com/cachix/devenv/blob/3febc91939aea65bdff8850f026443afb6b6b22f/flake.nix#L95
(pkgs.lib.evalModules {
  modules = [
    (devenvFlake.outPath + "/src/modules/top-level.nix")
  ];
  specialArgs = {
    inherit pkgs;
    inputs = devenvFlake.inputs;
  };
}).options
```

Now that expression is quite heavy (but the lightest one compared to the others!) you don't want have this inline defined in your code editors nixd config. For that I recommend putting this nix expression somewhere and then using an import statement as the actual statement given to nixd.
```js
//...
"devenv": {
  "expr": "import /path/to/expression.nix"
 }
//...
```

Of course we can do better than that! Just like @domenkozar said we need to incorporate `.devenv.flake.nix` to include our own option definitions. So I've been cooking up this monster:

```nix
let
  pkgs = import (builtins.getFlake "/path/to/your/flake").inputs.nixpkgs {};
  # if you don't want to add it to your own flake as an input:
  # pkgs = builtins.getFlake "github:/nixos/nixpkgs?rev=b3582c75c7f21ce0b429898980eddbbf05c68e55";

  inherit (builtins) getEnv readFile replaceStrings toFile typeOf attrNames match head concatStringsSep;
  inherit (pkgs) runCommand copyPathToStore;
  inherit (pkgs.lib) pipe;

  # Helper function to convert a deep set with only strings to a valid string expression
  # (Why does something like this not exist in nixpkgs?)
  stringSetToString = set: pipe set [
    # Get attribute names
    attrNames
    # collapse to string
    (map (name: "${name}=" + (
        if (typeOf set.${name} == "set") then
          stringSetToString set.${name}
        else
          "\"${set.${name}}\""
      ) + ";"
    ))
    # concat
    (concatStringsSep "")
    # add braces
    (str: "{${str}}")
  ];
in
pipe (
  (getEnv "PWD") + "/.devenv.flake.nix"
) [
  # The .devenv.flake.nix file has a special format. It allows complex logic inside the inputs block
  # which normal flakes do not allow. So we evaluate the inputs and then convert it to a string now with 
  # baked in let values.
  # The outputs expression must stay the same. So the part where the outputs begin is just kept.
  (file: '' 
    { inputs = ${stringSetToString (import file).inputs};
    ${pipe file [readFile (match ".*(outputs.*$)") head]}
  '')

  # A little hack that includes the `project` variable that contains the options set.
  # the variable is defined in a let..in block in the outputs section.
  (replaceStrings ["devShell ="] ["inherit project;devShell ="])

  # put this file together in a store directory with devenv.nix and devenv.lock (as flake.lock)
  (expr: runCommand "devenv-flake" {} ''
    mkdir -p $out
    cp ${copyPathToStore ((getEnv "PWD") + "/devenv.nix")} $out/devenv.nix
    cp ${copyPathToStore ((getEnv "PWD") + "/devenv.lock")} $out/flake.lock
    cp ${toFile "flake.nix" expr} $out/flake.nix;
    echo "builtins.getFlake \"$out\"" > $out/default.nix # needed for the import statement
  '')

  # one last time, import the final flake, this time actually as a flake
  import

  # lastly, we can extract the options from the outputs
  (finalFlake: finalFlake.outputs.project.options)
]
```

I tried my best to explain what happens there with all these comments. But in a nutshell the `.devenv.flake.nix` is manipulated in way that it contains the options in the outputs and can be evaluated by `builtins.getFlake`. Then it is copied to a directory in the store with `devenv.nix` and `devenv.lock`. From there `builtins.getFlake` is called and the options are returned.

This will obviously fail when the current working directory doesn't contain devenv stuff. It's totally fine to have two entries, one for general devenv and one for project specific devenv options in your nixd configuration. But if we want to profit from nix optimizations we need to merge it into one expression. So here is it (with some tweaks, thats the version I have in my nix config!):

```nix
flake:
let
  pkgs = import flake.inputs.nixpkgs {};

  inherit (builtins) getEnv hasAttr readFile replaceStrings toFile pathExists typeOf attrNames match head concatStringsSep;
  inherit (pkgs) runCommand copyPathToStore;
  inherit (pkgs.lib) pipe;

  # Helper function to convert a deep set with only strings to a valid string expression
  # (Why does something like this not exist in nixpkgs?)
  stringSetToString = set: pipe set [
    # Get attribute names
    attrNames
    # collapse to string
    (map (name: "${name}=" + (
        if (typeOf set.${name} == "set") then
          stringSetToString set.${name}
        else
          "\"${set.${name}}\""
      ) + ";"
    ))
    # concat
    (concatStringsSep "")
    # add braces
    (str: "{${str}}")
  ];
  
  # all options provided by devenv 
  devenvOptions = let 
    devenvFlake = flake.inputs.devenv;
    pkgs = import devenvFlake.inputs.nixpkgs {};
  in
  # This is copied and adapted from: 
  # https://github.com/cachix/devenv/blob/3febc91939aea65bdff8850f026443afb6b6b22f/flake.nix#L95
  (pkgs.lib.evalModules {
    modules = [
      (devenvFlake.outPath + "/src/modules/top-level.nix")
    ];
    specialArgs = {
      inherit pkgs;
      inputs = devenvFlake.inputs;
    };
  }).options;

  # The options specific for the currently openend devenv project
  projectOptions = pipe (
    (getEnv "PWD") + "/.devenv.flake.nix"
  ) [
    # The .devenv.flake.nix file has a special format. It allows complex logic inside the inputs block
    # which normal flakes do not allow. So we evaluate the inputs and then convert it to a string now with 
    # baked in let values.
    # The outputs expression must stay the same. So the part where the outputs begin is just kept.
    (file: '' 
      { inputs = ${stringSetToString (import file).inputs};
      ${pipe file [readFile (match ".*(outputs.*$)") head]}
    '')

    # A little hack that includes the `project` variable that contains the options set.
    # the variable is defined in a let..in block in the outputs section.
    (replaceStrings ["devShell ="] ["inherit project;devShell ="])

    # put this file together with devenv.nix and devenv.lock (as flake.lock)
    (expr: runCommand "devenv-flake" {} ''
      mkdir -p $out
      cp ${copyPathToStore ((getEnv "PWD") + "/devenv.nix")} $out/devenv.nix
      cp ${copyPathToStore ((getEnv "PWD") + "/devenv.lock")} $out/flake.lock
      cp ${toFile "flake.nix" expr} $out/flake.nix;
      echo "builtins.getFlake \"$out\"" > $out/default.nix # needed for the import statement
    '')

    # lastly, we can extract the options from the outputs
    import

    # lastly, we can extract the options by apply the inputs to the outputs function
    (finalFlake: finalFlake.outputs.project.options)
  ];

in
  # only include general devenv options when devenv is in inputs of provided flake
  (if hasAttr "devenv" flake.inputs then devenvOptions else {})
  //
  # only include project options if we actually are in a devenv project
  (if pathExists ((getEnv "PWD") + "/.devenv.flake.nix") then projectOptions else {})
```

(as a function that takes a flake containing nixpkgs and devenv as an input)

PR #1395, opened 2024-08-19
fix: let nix not cache bash generated file

Atry:
Originally, bash generates `/dev/fd/63` from `"file+file://"<(printf %s "$PWD")` and nix caches the file content with TTL 3600 seconds, as a result, if two flake-parts based devenv projects are created under different paths, they would interfere each other's cache for `/dev/fd/63`, and result in wrong `DEVENV_ROOT`.

This PR disables the nix file cache to avoid the problem.

### DIFF
--- a/templates/flake-parts/.envrc
+++ b/templates/flake-parts/.envrc
@@ -5,7 +5,9 @@ fi
 nix_direnv_watch_file flake.nix
 nix_direnv_watch_file flake.lock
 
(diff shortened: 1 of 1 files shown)


Atry:
Another solution is to create a flag like `--override-input devenv-root https://postman-echo.com/get?devenv-root="<(printf %s "$PWD")` 

postman-echo.com is a service to print the request information, for example, given the URL https://postman-echo.com/get?devenv-root=/home/myuser/myproject , we can use `builtins.fromJSON` to get the query string of `devenv-root`.

PR #1404, opened 2024-08-27
process-compose: fix running processes with `is_elevated`

### LINKED ISSUE #1391: process-compose `sudo: exec: command not found`
**Describe the bug**

Using process-compose with its new `is_elevated` option leads to the bug described in https://github.com/F1bonacc1/process-compose/issues/223

This happens because `exec` is not available when used in combination with `sudo`: https://github.com/cachix/devenv/blob/b285601679c7686f623791ad93a8e0debc322633/src/modules/process-managers/process-compose.nix#L60

cc @stephank: You added this change in https://github.com/cachix/devenv/commit/78766e0df9ba0ff1923511b288d0489ba2d3a077, do you maybe have any more insights to share or an idea how to solve this?

**To reproduce**
Please see https://github.com/F1bonacc1/process-compose/issues/223

**Version**
devenv 1.0.8 (x86_64-linux)

sandydoo:
Fixes #1391.

### DIFF
--- a/src/modules/process-managers/process-compose.nix
+++ b/src/modules/process-managers/process-compose.nix
@@ -57,7 +57,16 @@ in
           (name: value: "${name}=${toString value}")
           config.env;
         processes = lib.mapAttrs
-          (name: value: { command = "exec ${pkgs.writeShellScript name value.exec}"; } // value.process-compose)
+          (name: value:
+            let
+              scriptPath = pkgs.writeShellScript name value.exec;
+              command =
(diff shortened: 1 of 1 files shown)


RafaelKr:
@sandydoo is this `or false` actually required?

RafaelKr:
> Nix's `or` isn't the usual boolean operator `||`, in case that's your concern.

@sandydoo Thanks for the clarification!
And for the fix :tada:

PR #1406, opened 2024-08-30
Add support for Go delve workaround

andrzejressel:


src/modules/top-level.nix
```diff
@@ -250,6 +261,8 @@ in
       pkgs.pkg-config
     ];
 
+    hardeningDisable = [];
```
domenkozar, line 264: This is already handled with the default

domenkozar (changes requested):

src/modules/top-level.nix
andrzejressel: Fixed

PR #1408, opened 2024-09-01
Improve services.postgres db init

ppenguin:
Add user/pass capability to `services.postgres.initialDatabases` because we usually like to test our services with specific DB users. The options are optional, functionality remains the same if not given.

If we now do this:

```nix
  services.postgres = {
    enable = true;
    listen_addresses = "localhost";
    port = 15432;
    initialDatabases = [
      {
        name = "pdns";
        user = "pdns";
        pass = "mypdnssecretpass1234";
        schema = ./_testdata; # *.sql in version order
      }
    ];
  };
```

the following will happen:

1. `postgres` DB is created as usual, owned by `$USER` (who is SUPERUSER) for this instance)
2. A role `pdns` with password `${pass}` is created
3. Database `${name}` is created
4. All privileges on `pdns` are granted to user `pdns`.
5. The schema seeding scripts in `${schema}` are executed as user `pdns`.

### DIFF
--- a/docs/reference/options.md
+++ b/docs/reference/options.md
@@ -34784,6 +34784,8 @@ list of (submodule)
 [
   {
     name = "foodatabase";
+    user = "ufoo";
+    pass = "barpaz";
     schema = ./foodatabase.sql;
   }
   { name = "bardatabase"; }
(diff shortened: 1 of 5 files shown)


domenkozar:
Since postgresql is critical bit of the infra, can you add a test for it to tests/?

PR #1415, opened 2024-09-05
Add `containers.<name>.layers` and `containers.<name>.enableLayerDeduplication`

Atry:
These options allow for layer deduplication using the approach described at https://blog.eigenvalue.net/2023-nix2container-everything-once/

### DIFF
--- a/src/modules/containers.nix
+++ b/src/modules/containers.nix
@@ -116,12 +116,7 @@ let
 
     maxLayers = cfg.maxLayers;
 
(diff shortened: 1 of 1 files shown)


domenkozar:
Is there any reason not to use `enableLayerDeduplication`?

Atry:
When `enableLayerDeduplication` is `false`, layers are independent, therefore layers would not be invalidated due to changes in previous layers.

PR #1418, opened 2024-09-05
feat: add `devenv.flakeArgs` to avoid `--no-pure-eval`

Atry:
This PR avoids `--no-pure-eval` in `devenv-devShell.nix`, in favor of `"--override-input" "devenv-root" "path:.devenv/state/pwd"`.

### DIFF
--- a/src/devenv-devShell.nix
+++ b/src/devenv-devShell.nix
@@ -19,7 +19,7 @@ pkgs.writeScriptBin "devenv" ''
 
   case $command in
     up)
-      procfilescript=$(nix build '.#${shellPrefix (config._module.args.name or "default")}devenv-up' --no-link --print-out-paths --no-pure-eval)
+      procfilescript=$(nix build '.#${shellPrefix (config._module.args.name or "default")}devenv-up' --no-link --print-out-paths ${lib.escapeShellArgs config.devenv.flakeArgs})
       if [ "$(cat $procfilescript|tail -n +2)" = "" ]; then
         echo "No 'processes' option defined: https://devenv.sh/processes/"
         exit 1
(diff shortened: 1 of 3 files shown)


Atry:
@domenkozar What do you think?

Nebucatnetzer:
I like the idea of making devenv more pure but I would avoid putting necessary steps into the `.envrc` file.
This way direnv becomes a dependency, currently it's just a nice tool to have.
In CI I would then either need to have direnv present or make the state directory manually.
The the moment it its enough to just run `nix develop --no-pure-eval --command some-test-command`

Atry:
> I like the idea of making devenv more pure but I would avoid putting necessary steps into the `.envrc` file. This way direnv becomes a dependency, currently it's just a nice tool to have. In CI I would then either need to have direnv present or make the state directory manually. The the moment it its enough to just run `nix develop --no-pure-eval --command some-test-command`

This PR did not affect `nix develop --no-pure-eval  --command some-test-command` behavior in general.  It just makes flake projects' `devenv up` pure. It does not affect `devenv.yaml` projects because they have a diiferent implementation of `devenv`.

Since it changed `devenv up`, I updated the PR in 56f801a to still support `nix develop --no-pure-eval --command devenv up` for flake projects that do not include an `.envrc`.



> BTW. isn't `devenv-root` at `.devenv` and `devenv-state` would be `.devenv/state`?

Yes

Atry:
@domenkozar Potentially `devenv.yaml` based projects could also utilize the `.devenv/state/pwd` file to avoid `--no-pure-eval`, but it is out of the scope of this PR.

Nebucatnetzer:
> This PR did not affect nix develop --no-pure-eval --command some-test-command behavior in general.

We have some PHP tests that write session files to `.devenv/state/php-fpm` so they wouldn't have worked anymore.
Thank you for updating it :).

PR #1422, opened 2024-09-06
Allways creating home directory even if nothing in `copyToRoot`

chitter99:
Fixes missing home directory when no `copyToRoot`.

```
{
  containers.example = {
    name = "example";
    startupCommand = "bash";
    copyToRoot = [ ];
  };
}
```

Previous error message

```
✦ ❯ docker run -it example
mkdir: cannot create directory '/env/.devenv': Permission denied
ln: failed to create symbolic link '/env/.devenv/profile': No such file or directory
ln: failed to create symbolic link '/env/.devenv/run': No such file or directory
```

### DIFF
--- a/src/modules/containers.nix
+++ b/src/modules/containers.nix
@@ -40,8 +40,11 @@ let
   gid = "1000";
   homeDir = "/env";
 
-  mkHome = path: (pkgs.runCommand "devenv-container-home" { } ''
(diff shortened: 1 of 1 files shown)


domenkozar:
I've reverted this in https://github.com/cachix/devenv/pull/1433 as it breaks container tests

PR #1428, opened 2024-09-07
Update templates .envrc files

miranbastaja:
Update templates .envrc files to use watch_file
instead of the deprecated nix_direnv_watch_file.


templates/flake-parts/.envrc
```diff
@@ -2,8 +2,8 @@ if ! has nix_direnv_version || ! nix_direnv_version 2.2.1; then
   source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/2.2.1/direnvrc" "sha256-zelF0vLbEl5uaqrfIzbgNzJWGmLzCmYAkInj/LNxvKs="
```
thenonameguy, line 2: With this current `direnvrc` direnv extension, this break the `nix_watches` extra functionality, as `devenv` doesn't use https://github.com/nix-community/nix-direnv directly.

miranbastaja: Okay, I bumped the version to 2.5 as that's the version that `nix_direnv_watch_file` was originally deprecated in. Sorry for the inconvenience, I'm not familiar with the more advanced usage of direnv.

PR #1421, opened 2024-09-06
Fix `container build` command failing due to insufficient config option specification for `perms`

ento:
Trying to build a shell container against a minimal devenv.nix fails right now:

```sh
$ devenv container build shell
[...]
error: The option `containers.shell.layers."[definition 1-entry 1]".perms."[definition 1-entry 1]".gid' does not exist. Definition values:
- In `/nix/store/virtual0000000000000000000000007-source/src/modules/containers.nix': 1000
```

devenv version: devenv 1.0.8 (x86_64-linux)

devenv.nix:

```nix
{ pkgs, lib, config, inputs, ... }:

{}
```

<details><summary>devenv.yaml</summary>
<p>

```yaml
inputs:
  devenv:
    # fails
    url: github:cachix/devenv/58f2ff7b7cbb6f2325f5ddae6858a6e0c2d66cbd?dir=src/modules
    # ok
    #url: github:cachix/devenv/d1e690595ed39a15426cfa503bb44383cfa03ae7?dir=src/modules
  mk-shell-bin:
    url: github:rrbutani/nix-mk-shell-bin
  nixpkgs:
    url: github:cachix/devenv-nixpkgs/rolling
  nix2container:
    url: github:nlewo/nix2container
    inputs:
      nixpkgs:
        follows: nixpkgs
```

</p>
</details> 

<details><summary>devenv.lock</summary>
<p>

```json
{
  "nodes": {
    "devenv": {
      "locked": {
        "dir": "src/modules",
        "lastModified": 1725552963,
        "owner": "cachix",
        "repo": "devenv",
        "rev": "58f2ff7b7cbb6f2325f5ddae6858a6e0c2d66cbd",
        "treeHash": "1cd0d97307e87b086e3c47b723d3f0037dc4d865",
        "type": "github"
      },
      "original": {
        "dir": "src/modules",
        "owner": "cachix",
        "repo": "devenv",
        "rev": "58f2ff7b7cbb6f2325f5ddae6858a6e0c2d66cbd",
        "type": "github"
      }
    },
    "flake-compat": {
      "flake": false,
      "locked": {
        "lastModified": 1696426674,
        "owner": "edolstra",
        "repo": "flake-compat",
        "rev": "0f9255e01c2351cc7d116c072cb317785dd33b33",
        "treeHash": "2addb7b71a20a25ea74feeaf5c2f6a6b30898ecb",
        "type": "github"
      },
      "original": {
        "owner": "edolstra",
        "repo": "flake-compat",
[...]

PR #1432, opened 2024-09-09
`cli` test should use local devenv module

ento:
One change from #1421 that wasn't part of #1427. (Both PRs addressed the same bug in `devenv container build shell`, and the latter was merged.)

While working on #1421, I noticed `tests/cli`, which runs `devenv container build shell` as part of its test script, wouldn't fail as it should've when the bug was present. I believe it's because `tests/cli/devenv.yaml` doesn't override the `devenv` input like `devenv-run-tests` or `tests/clean/devenv.yaml` does.

https://github.com/cachix/devenv/blob/9ff4999b3555936942ee84f6c11fbb103ca52b3e/devenv-run-tests/src/main.rs#L80-L85

https://github.com/cachix/devenv/blob/9ff4999b3555936942ee84f6c11fbb103ca52b3e/tests/clean/devenv.yaml#L1-L3

### DIFF
--- a/tests/cli/.test.sh
+++ b/tests/cli/.test.sh
@@ -1,6 +1,7 @@
 set -xe
 
 rm devenv.yaml || true
+devenv inputs add devenv "path:../../?dir=src/modules"
 devenv build languages.python.package
 devenv shell ls -- -la | grep ".test.sh"
 devenv shell ls ../ | grep "cli"


domenkozar:
Thanks for catching this one! I wonder if there's a better design to make sure `devenv` cli always uses the override in all commands.

domenkozar:
I'm going to let @sandydoo chip in next week if we could expose `devenv` on `$PATH` with correct `devenv` input override.

domenkozar:
https://github.com/cachix/devenv/commit/10ab16635312bc59aac940e9faef5936cae0bc6f

Issue #1364, opened 2024-08-07
Devenv hangs forever


artemlive:
**Describe the bug**
Hello! 
I have a problem with Devenv on MacOS (Sonoma 14.6). 
When I try to init devenv it hangs at the devenv init step:
```
devenv init                                           
• Creating devenv.nix
• Creating devenv.yaml
• Creating .envrc
• Creating .gitignore
direnv: loading ~/ops-stuff/repos/github-management/.envrc
direnv: loading https://raw.githubusercontent.com/cachix/devenv/95f329d49a8a5289d31e0982652f7058a189bfca/direnvrc (sha256-d+8cBpDfDBj41inrADaJt+bDWhOktwslgoP5YiGJ1v0=)
direnv: using devenv
direnv: .envrc changed, reloading
direnv: ([direnv export zsh]) is taking a while to execute. Use CTRL-C to give up.||
```

It hangs forever if you don't press Ctrl-C. 
I checked what it runs underneath the hood via ps and ran the same command with an extra verbose. 
```
/nix/store/z1zrijdcxzx5a8yf4prggamg6k41hcmi-nix-2.21-devenv/bin/nix --show-trace --extra-experimental-features nix-command --extra-experimental-features flakes --option warn-dirty false --keep-going --max-jobs 4 --option eval-cache false eval .#devenv.cachix --json -vvvvvvvv
evaluating file '<nix/derivation-internal.nix>'
using cache entry '{"_what":"gitLastModified","rev":"0a41d25e85caeb81211a5742b47a9a986edf35d2"}' -> '{"lastModified":1722604605}'
evaluating file '/.devenv.flake.nix'
```
And it stuck on the evaluating file /.devenv.flake.nix. 
The same thing happens if I'm trying to run any command. 
For example if I run `devenv info`: 
```
devenv info -v
• Running command: /nix/store/z1zrijdcxzx5a8yf4prggamg6k41hcmi-nix-2.21-devenv/bin/nix --show-trace --extra-experimental-features nix-command --extra-experimental-features flakes --option warn-dirty false --keep-going --max-jobs 4 --option eval-cache false flake metadata
• Running command: /nix/store/z1zrijdcxzx5a8yf4prggamg6k41hcmi-nix-2.21-devenv/bin/nix --show-trace --extra-experimental-features nix-command --extra-experimental-features flakes --option warn-dirty false --keep-going --max-jobs 4 --option eval-cache false eval --raw .#info
```
And it hangs forever.
[...]

Issue #1114, opened 2024-04-09
Failed to copy container. Digest did not match


kvz:
**Describe the bug**
I'm trying to copy my container to AWS ECR from my Linux X64 CI server, but hitting the following issue. `Failed to copy container. Digest did not match`.

**To reproduce**

```bash
  # also tried with `--system x86_64-linux`, does not make a difference
  devenv container build processes

  # ecr login
  aws ecr get-login-password --region us-east-1 | docker login --username AWS --password-stdin 123123123123.dkr.ecr.us-east-1.amazonaws.com
  # ecr create repository
  if ! aws ecr describe-repositories --repository-names processes --region us-east-1; then
    aws ecr create-repository --repository-name processes --region us-east-1
  fi
  # ecr push. also tried with `--copy-args="--preserve-digests"`. does not make a difference
  devenv container \
    --registry docker://123123123123.dkr.ecr.us-east-1.amazonaws.com/ \
  copy processes
```

Output:

```bash
• Building processes container ...
• Using Cachix: devenv

/nix/store/31g4rq1p6spmihc95qm3qc2iy8w7hm5b-image-processes.json
✔ Building processes container in 2.2s.
• Copying processes container ...• Running /nix/store/g96s72rwnk3vdkqnvd5wmkmnrmk0z9sk-copy-container /nix/store/31g4rq1p6spmihc95qm3qc2iy8w7hm5b-image-processes.json docker://123123123123.dkr.ecr.us-east-1.amazonaws.com/ --preserve-digests

Copying container /nix/store/31g4rq1p6spmihc95qm3qc2iy8w7hm5b-image-processes.json to docker://123123123123.dkr.ecr.us-east-1.amazonaws.com/processes:latest

Getting image source signatures
Copying blob sha256:67eb02cf19ceef2d35b6a7373359a5f7ede6f478e59ef55ef0688be70b03[97](https://github.com/transloadit/api2/actions/runs/8615325344/job/23610670898#step:6:98)d2
Copying blob sha256:2ac708344cd9024282848c53dbd92614463d4ce574312949cd43f09ff92e8e85
[...]

PR #1433, opened 2024-09-10
Nix 2.24

domenkozar:
Fixes https://github.com/cachix/devenv/issues/1364
Fixes https://github.com/NixOS/nix/issues/10647
Fixes https://github.com/cachix/devenv/issues/1114


lf-:
I'm not sure the actual version here but the .version suggests it might be 2.24.0, which doesn't have this morning's security fix. Worth at least checking.

domenkozar:
Hey, our versions don't really match upstream versions. I'm happy to make that clearer, open to ideas!

lf-:
Yeah, I realize. I'm not sure how the git history was done so this may not actually be an issue but the "compare" button on GitHub seems to imply the fix patches may be missing from the branch; it's unclear. That's why I'm suggesting to check to make sure :)

domenkozar:
https://github.com/domenkozar/nix/commits/devenv-2.24/

PR #1436, opened 2024-09-11
feat: add environment variables for process-compose

Atry:
These environment variables enable `process-compose` commands like `process-compose process list`


domenkozar:
Can you make a PR against https://github.com/cachix/devenv/pull/1161 since we'll merge that soon? Thanks!

domenkozar:
Needs a rebase

Atry:
Rebased

sandydoo (approved):
I've pushed the following fixes:
- Filtered out `PC_CONFIG_FILES` from the final config to avoid infinite recursion.
- Updated things to use the new renamed process compose options.

I've also dropped `PC_DISABLE_TUI`. I'm not quite sure why we'd want to set this by default. End users may want to set this in their own config to temporarily override `tui.enable`. And we also have that devenv-specific `PC_TUI_ENABLED` var that we need to support.

PR #1465, opened 2024-09-23
Add `ncurses` if process-compose's TUI is enabled

Atry:
process-compose's TUI depends on ncurses

### DIFF
--- a/src/modules/process-managers/process-compose.nix
+++ b/src/modules/process-managers/process-compose.nix
@@ -44,7 +44,7 @@ in
         -U up "$@" &
     '';
 
-    packages = [ cfg.package ];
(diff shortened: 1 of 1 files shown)


src/modules/process-managers/process-compose.nix
```diff
@@ -44,7 +44,7 @@ in
         -U up "$@" &
     '';
 
-    packages = [ cfg.package ];
+    packages = [ cfg.package ] ++ lib.optionals config.process.process-compose.tui pkgs.ncurses;
```
sandydoo, line 47:
```suggestion
    packages = [ cfg.package ] ++ lib.optionals cfg.tui.enable pkgs.ncurses;
```

sandydoo (changes requested):

sandydoo (approved):

PR #1477, opened 2024-09-26
devenv: escape shell vars in task exports

sandydoo:
Fixes sourcing exported variables with spaces in them. Hello, macOS.

### DIFF
--- a/devenv/src/tasks.rs
+++ b/devenv/src/tasks.rs
@@ -178,7 +178,11 @@ impl TaskState {
                     for (env_key, env_value) in env_obj {
                         if let Some(env_str) = env_value.as_str() {
                             command.env(env_key, env_str);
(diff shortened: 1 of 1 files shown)


devenv/src/tasks.rs
```diff
@@ -836,6 +840,21 @@ impl TasksUi {
     }
 }
 
+/// Escape a shell variable by wrapping it in single quotes.
+/// Any single quotes within the variable are escaped.
+fn shell_escape(s: &str) -> String {
```
domenkozar, line 845: I wonder if we should instead use some library like https://docs.rs/shell-quote/latest/shell_quote/ not to maintain this code which is tricky to get right

sandydoo: Yeah, maybe. The above is what `nix develop` does, which felt like a safe choice without getting into the weeds.

PR #1478, opened 2024-09-26
docs: lock requirements

sandydoo:


sandydoo:
And that seems to have fixed our CF pages build!

PR #1489, opened 2024-10-01
Bump nixpkgs

domenkozar:


src/modules/process-managers/process-compose.nix
```diff
@@ -76,6 +76,7 @@ in
   config = lib.mkIf cfg.enable {
     process.manager.args = {
       "config" = cfg.configFile;
+      "disable-dotenv" = true;
```
shyim, line 79: we need to check if the version of process compose is high enough to add that parameter right?

domenkozar: Since we're bumping devenv-nixpkgs this should be good.

If someone pins nixpkgs, we currently have no way to check if all supported versions work.

shyim: we avoid devenv-nixpkgs, as the cache is working really bad for our asia colleagues. a `devenv update` should fix it, so it's fine i guess

PR #1499, opened 2024-10-05
chore: reduce tasks binary build time

sandydoo:
Speeds up the tasks binary build by x3 and reduces the dependency list.


domenkozar:
Nice!

PR #1505, opened 2024-10-07
Remove PATH from devenv:python:uv task exports.

### LINKED ISSUE #1502: python.uv.sync.enable and python.venv.enable have conflicting PATH exports
**Describe the bug**
In order to have a python venv that is managed through uv, one has to set both `languages.python.uv.sync.enable` and `languages.python.venv.enable` to true. Since the tasks update, both define PATH as exports - however, the initUvScript does not activate the virtual environment and therefore its path does not include the venv at the front. This randomly leads to a `.devenv/load-exports` file in which PATH from "devenv:python:virtualenv" gets overwritten by the value from "devenv:python:uv" and therefore not correctly activating the venv.

I tried to fix it by setting:
```nix
{
   tasks."devenv:python:uv".before = ["devenv:python:virtualenv"];
}
```
But the order of the tasks does not seem to reflect the order of variable exports in `.devenv/load-exports`.

Removing "PATH" from the task exports of "devenv:python:uv" seems to work and I have not noticed any problems:
```nix
{
   tasks."devenv:python:uv".exports = lib.mkForce ["VIRTUAL_ENV"];
}
```

**To reproduce**
Minimal devenv.nix to reproduce:
```nix
{
  languages.python = {
    enable = true;
    venv.enable = true;
    uv.enable = true;
    uv.sync.enable = true;
  };
}
```

**Version**
 devenv 1.3.0 (x86_64-linux)

mauricege:
Removes PATH from the exports of the `devenv:python:uv` task. PATH is not modified in initUvScript and will conflict with `devenv:python:virtualenv` PATH export which leads to virtual environments randomly not activating on enterShell

Fixes #1502

### DIFF
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -470,7 +470,7 @@ in
       "devenv:python:uv" = lib.mkIf cfg.uv.sync.enable {
         description = "Initialize uv sync";
         exec = initUvScript;
-        exports = [ "PATH" "VIRTUAL_ENV" ];
+        exports = [ "VIRTUAL_ENV" ];
       };
 
       "devenv:enterShell".after = lib.optional cfg.venv.enable "devenv:python:virtualenv"


domenkozar:
See https://github.com/cachix/devenv/pull/1504/files

PR #1506, opened 2024-10-07
devenv: fix warning about `--update-input` when running `devenv update`

sandydoo:


devenv/src/cnix.rs
```diff
@@ -217,20 +217,14 @@ impl<'a> Nix<'a> {
...
-                .await?;
-            }
-            None => {
-                self.run_nix("nix", &["flake", "update"], &self.options)
-                    .await?;
-            }
+        let mut args = vec!["flake", "update"];
+
+        if let Some(input_name) = input_name {
+            args.push(input_name);
```
domenkozar, line 223: I don't see that `nix flake update` supports args for inputs?

sandydoo: What version are you on? 😄 Our patched nix throws an deprecation warning about this.

We could even take a list of input names to update.

sandydoo:
```
➜ devenv update devenv
• Updating devenv.lock with input devenv ...
warning: '--update-input' is a deprecated alias for 'flake update' and will be removed in a future version.
✔ Updating devenv.lock with input devenv in 0.2s.
```

PR #1510, opened 2024-10-08
fix: output non-zero code when `devenv test` fails

sandydoo:
Make sure we don't exit with code `0` if `devenv test` fails.

This requires both a module and a CLI update. On older versions of devenv, the updated tasks runner will cut of a bit of output at the end ("Tests failed :("), as it bails out of the `nix develop` command early.

I would've preferred something like `cmd.run().await?.bail_on_error()`, where we can make the decision to bail on each invocation, but that's a bit too invasive at this time.

Reported on Discord.

### DIFF
--- a/devenv-eval-cache/src/command.rs
+++ b/devenv-eval-cache/src/command.rs
@@ -202,6 +202,7 @@ pub fn supports_eval_caching(cmd: &Command) -> bool {
     cmd.get_program().to_string_lossy().ends_with("nix")
 }
 
+#[derive(Debug)]
 pub struct Output {
     /// The status code of the command.
     pub status: process::ExitStatus,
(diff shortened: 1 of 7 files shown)


sandydoo:
Hmmm can't get it to fail locally 🤔

sandydoo:
```
✔ Building shell in 0.0s.
/nix/store/s5mr3j8hj6ph3ppx03q46r6jjdldbkfg-init-npm.sh: line 18: /nix/store/230hxd4jaz9rdk20dh66na4bx26c10fc-nodejs-slim-20.11.1/bin/npm: No such file or directory
/nix/store/s5mr3j8hj6ph3ppx03q46r6jjdldbkfg-init-npm.sh: line 18: /nix/store/230hxd4jaz9rdk20dh66na4bx26c10fc-nodejs-slim-20.11.1/bin/npm: No such file or directory
```

sandydoo:
Depends on #1511.

Issue #1527, opened 2024-10-16
Cachix substituter not being picked up by devenv


samchouse:
**Describe the bug**
Devenv complains about its cachix instance not being added to trusted substituters even though it is added.

**To reproduce**
Since I've added devenv as a trusted substituter I'd expect cachix to work. https://gist.github.com/samchouse/0cba1c6787af2d26a0c7a93a681cc292

**Version**
devenv 1.3.0 (x86_64-linux) via nixpkgs (home.packages)

rvveber:
Same issue here!
In addition the recommendation on how to add them is incorrect:
```
b) Add binary caches to /etc/nix/nix.conf yourself by editing configuration.nix:
{
    nix.extraOptions = ''
        extra-substituters = https://devenv.cachix.org;
        extra-trusted-public-keys = devenv.cachix.org-1:w1cLUi8dv3hnoSPGAuibQv+f9TZLr6cv/Hm9XgU50cw= nixpkgs-python.cachix.org-1:hxjI7pFxTyuTHn2NkvWCrAUcNZLNS3ZAvfYNuYifcEU=;
    '';
}
```

Should be without the `;`'s at the end.

sandydoo:
devenv doesn't check your configured substituters, so the only way to disable this atm is to add the following to your `devenv.nix`:

```nix
cachix.enable = false;
```

I'll try to get this fixed.

The error message should've at least mentioned this, but it somehow got lost over time.

samchouse:
Does this completely disable cachix meaning I have to build everything myself?

sandydoo:
You have both the nixos and devenv caches set up in `/etc/nix/nix.conf`, so you're all set up. This tells devenv not to try to set them up for you.

samchouse:
Is this normal with `cache.enable = false;`?

```
warning: ignoring the client-specified setting 'trusted-public-keys', because it is a restricted setting and you are not a trusted user
```

PR #1532, opened 2024-10-16
devenv: fixup nixos-specific cache fallback docs

sandydoo:
Removes semicolons from the `nix.conf` suggestion and adds back the missing instructions to set `cachix.enable = false` when using os-wide caches on nixos.


sandydoo:
Part of https://github.com/cachix/devenv/issues/1527

Issue #1111, opened 2024-04-08
Mixing system packages with python results in shared library errors


bobvanderlinden:
**Describe the bug**

When using system packages from within a python application `symbol lookup error` may happen.
This happens when the system package (like `git`) is linked against a different version of `libc` compared to the python version (from devenv) is linked against.

A specific case where this happens more often occurs when using a remote git repository as a requirement for your devenv-based python project. devenv calls `pip install -r requirements.txt`, which will run `git`. `git` is usually not defined inside devenv, thus the required `libc` is likely to be different from what is used in devenv. The `python` wrapper within devenv runs python with `LD_LIBRARY_PATH=$DEVENV_PROFILE/lib`, which will force ld to use libraries within `$DEVENV_PROFILE/lib`. The `libc` that resides in `$DEVENV_PROFILE/lib` is a different version than what `git` assumes to be ld-ed against.

This results in errors like the following:

```
Collecting git+https://github.com/pypa/sampleproject.git (from -r /nix/store/g41hqm03k2179v5pkyv889gyrgb370g0-requirements.txt (line 1))
  Cloning https://github.com/pypa/sampleproject.git to /tmp/nix-shell.07Ag0j/nix-shell.tKP4lT/pip-req-build-jidkerj5
  error: subprocess-exited-with-error
  
  × git version did not run successfully.
  │ exit code: 127
  ╰─> [2 lines of output]
      git: error while loading shared libraries: __vdso_gettimeofday: invalid mode for dlopen(): Invalid argument
      [end of output]
```

**Potential solutions**

* Adding `git` to `packages` in devenv will resolve the problem with git, but on my system I got the next binary that resulted in the a similar error (`gh`). Adding all personal tools is one way to get over this problem.
* Avoid using other binaries within the python application by not using git.
* Resetting `LD_LIBRARY_PATH` upon calling other binaries from within the wrapped devenv python.
* Make sure Python libraries are always linked against specific (working) versions of native libraries. This can be done using `patchelf`/`autopatchelf`. In addition, `LD_LIBRARY_PATH` should not be used so that `ld` won't be forced to use different libraries. See PR https://github.com/cachix/devenv/pull/840 as one possibility.

**To reproduce**

Follow the instructions from the readme:

https://gist.github.com/bobvanderlinden/f31299079b67ddcc1b2c6029cf6569f6

**Version**
 
```
devenv 1.0.2 (x86_64-linux)
```

shoskensMagics:
[...]

PR #1542, opened 2024-10-19
introduce languages.python.finalPackage

bobvanderlinden:
Currently it isn't possible to use `languages.python` without the `LD_LIBRARY_PATH` wrapper. Since this wrapper resulted into some unwanted side-effects I'd like to be able to override the wrapper as well.

`languages.python.package` is the unwrapped package and `languages.python.finalPackage` wraps `languages.python.package`, which is in turn used like it was previously.

Since the `package = cfg.finalPackage;` was a bit confusing, I opted to replace this with `inherit (cfg.finalPackage) interpreter sitePackages;`.

### DIFF
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -10,21 +10,7 @@ let
   );
 
   readlink = "${pkgs.coreutils}/bin/readlink -f ";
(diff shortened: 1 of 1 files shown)


bobvanderlinden:
This should also help to implement solutions like https://github.com/cachix/devenv/pull/840 to workaround issues related to https://github.com/cachix/devenv/issues/1111.

PR #1565, opened 2024-10-31
feat: Added kafka service

### LINKED ISSUE #1563: Apache Kafka service
Hello, I think it would be nice to have a service for apache kafka.
I wanted to try to set the service up myself, but I am not sure how to add default configuration or how it could be changed by the user.

Alexnortung:
Here is a rough implementation of a kafka service for devenv.

I think there are definetely place for configuration options. but I am not sure how we can implement this properly. So feedback would be appreciated :)

closes #1563

### DIFF
--- a/tests/kafka/.test.sh
+++ b/tests/kafka/.test.sh
@@ -1,0 +1,3 @@
+set -e
+
+kafka-topics.sh --list --bootstrap-server localhost:9092
--- a/src/modules/services/kafka-connect.nix
+++ b/src/modules/services/kafka-connect.nix
@@ -98,6 +98,7 @@ in
 
           "plugin.path" = lib.mkOption {
             type = types.nullOr (types.listOf (types.either types.str types.path));
+            default = null;
             description = ''
               The list should consist of top level directories that include any combination of:
               a) directories immediately containing jars with plugins and their dependencies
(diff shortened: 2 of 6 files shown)


Alexnortung:
Okay, I figured it out, you could just pass the connectors to the start command

sandydoo:
Take a look at https://github.com/NixOS/nixpkgs/blob/nixos-unstable/nixos/modules/services/misc/apache-kafka.nix. You should be able to reuse large parts of this, particularly the config/settings handling, and ideally keep a similar interface.

Alexnortung:
Hey @sandydoo thanks for the feedback!
I have now used many of the options provided in the NixOS module. However I have tweaked some defaults, so that it just works without having the user configure anything but `enable = true`.

sandydoo:
> However I have tweaked some defaults, so that it just works without having the user configure anything but enable = true.

That's the right approach.

@Alexnortung, is there a basic test we could add for this to make sure it launches? I'm not familiar with Kafka, but maybe a ping/healthcheck/queue-unqueue type of thing. See the `examples` and `tests` directories.

Alexnortung:
Hi @sandydoo I have Added settings to kafka connect and added a test for kafka.
I also wanted to add readiness probe to kafka, but it seems like something weird is going on when you add readiness probe to the service. Do you have any suggestions?

tests/kafka/.test.sh
```diff
@@ -0,0 +1,3 @@
+set -e
+
+kafka-topics.sh --list --bootstrap-server localhost:9092
```
domenkozar, line 3: Where does `kafka-topics.sh` come from?

Alexnortung: If you prefer we could probably also do a `curl` since the script is probably just calling an endpoint in kafka

Alexnortung:
Am I missing something since the github actions are failing?

tests/kafka/.test.sh
domenkozar: yeah let's do just curl

domenkozar:
> Am I missing something since the github actions are failing?

some of them fail, but you'll see kafka being run in the test action

Alexnortung:
I have now also added readiness probes for kafka and kafka connect. I have also added kafka as a dependency to kafka connect, so that kafka connect doesn't fail to start if kafka is not started correctly or is using more time than usual.

PR #1571, opened 2024-11-03
pre-commit: rename and alias to `git-hooks`

sandydoo:
Renames every mention of `pre-commit-hooks` to `git-hooks`.

##### For backwards-compatibility

- `pre-commit` is aliased to `git-hooks`, without throwing a deprecation warning to reduce churn.
- `pre-commit-hooks` input follows the `git-hooks` input.
- Probably unnecessary, but the `git-hooks` module will try both inputs before failing.

##### Removed

I removed `pre-commit-hooks` from `specialArgs`. It was only used by our module and I doubt anyone relied on it. We use `inputs.<name>` instead.


sandydoo:
`git-hooks.hooks` is a little annoying to read, but we can't really do `git.hooks` because of all the other top-level `git-hooks` options, like `settings`, `excludes`, etc.

PR #1573, opened 2024-11-04
tasks: pickup outgoing task dependencies

### LINKED ISSUE #1557: Cannot run task after `poetry install`
**Describe the bug**
I'm attempting to run a small script after `poetry install` completes in my repo. `poetry install` downloads some binaries that are dynamically linked against `ld`, and thus do not work on NixOS. I'm trying to automatically run `patchelf` after `poetry install` completes in order to patch `ruff`, thus allowing NixOS developers to not have to worry about doing it themselves or manually running a command after each `poetry install`.

Here is my devenv module:

```nix
{
  # Configure packages to install.
  # Search for package names at https://search.nixos.org/packages?channel=unstable
  packages = with pkgs; [
    openssl

    patchelf
    autoPatchelfHook
  ];

  # Install Python and manage a virtualenv with Poetry.
  languages.python.poetry.enable = true;
  # Automatically activate the poetry virtualenv upon entering the shell.
  languages.python.poetry.activate.enable = true;
  languages.python.poetry.install.installRootPackage = true;
  languages.python.poetry.install.enable = true;
  
  tasks = {
    "sygnal:patchpythonbinaries" = {
      description = "Patch binaries dynamically linked to ld, which will not work on NixOS";
      exec = ''
        patchelf --remove-rpath .venv/bin/ruff
        autoPatchelf .venv/bin/ruff
      '';
      after = [ "devenv:enterShell" ];
    };
  };
}
```

I first tried to put the above `patchelf; autoPatchelf` invocation in `enterShell`, but that failed as `poetry install` runs *after* whatever a devenv module defines in `enterShell`. Meaning `patchelf` would be called, and *then* `ruff` would be downloaded by `poetry install`. 

Interestingly, `poetry install` is defined as a task that apparently runs *after* `enterShell`:

https://github.com/cachix/devenv/blob/fa53082e82e90bd378cf850ca3d2b6892eebc77e/src/modules/languages/python.nix#L457-L486

I attempted all sorts of combinations of trying to get my task to run after both `"devenv:enterShell"` and `"devenv:python:poetry"`, but to no avail. If I do:

```
      before = [ "devenv:enterShell" ];
```

then my task runs, but does so before the `poetry install`, so fails as `ruff` does not exist:

```
direnv: loading ~/code/sygnal/.envrc
direnv: using flake path:///home/work/code/nix-flakes/project-flakes/sygnal --impure
[0/1 built, 0.0 MiB DL] querying devenv-shell-env on https://devenv.cachix.orgdirenv: ([/nix/store/3mydh7746lji25ry2aygsy5i4s0i23x2-direnv-2.35.0/bin/direnv export zsh]) is taking a while to execute. Use CTRL-C to give up.
direnv: nix-direnv: Renewed cache
Running tasks     devenv:enterShell
Failed            sygnal:patchpythonbinaries 3ms
Succeeded         devenv:python:poetry       2582ms
Dependency failed devenv:enterShell          
1 Succeeded, 1 Failed, 1 Dependency Failed   2.59s

--- sygnal:patchpythonbinaries failed with error: Task exited with status: exit status: 1
--- sygnal:patchpythonbinaries stdout:
--- sygnal:patchpythonbinaries stderr:
0002.59: patchelf: getting info about '.venv/bin/ruff': No such file or directory
```

however, if I do the logical thing and:

```
      after = [ "devenv:enterShell" ];
```

my task doesn't run at all:

```
direnv: loading ~/code/sygnal/.envrc
direnv: using flake path:///home/work/code/nix-flakes/project-flakes/sygnal --impure
direnv: ([/nix/store/3mydh7746lji25ry2aygsy5i4s0i23x2-direnv-2.35.0/bin/direnv export zsh]) is taking a while to execute. Use CTRL-C to give up.
direnv: nix-direnv: Renewed cache
Running tasks     devenv:enterShell
Succeeded         devenv:python:poetry       2675ms
Succeeded         devenv:enterShell          7ms
2 Succeeded                                  2.68s
```

So I'm at a loss. This is using a flake, not `devenv.nix/yaml/lock`.

**To reproduce**

My devenv module can be found here, along with all other relevant files: https://github.com/element-hq/nix-flakes/blob/anoa/devenv_repro/project-flakes/sygnal/module.nix

**Version**
 
Using flakes, devenv v1.3.1.

sandydoo:
~I think we're running the tasks in reverse order after the toposort. The following tasks are speced to run _after_ `enterShell`, but they instead run before.~

This PR lets the graph traversal visit both incoming _and_ outgoing edges.
Traversing all the edges of a node allows picking up tasks that are only specified to run _after_ the current node. For example:

```nix
tasks."my-app:after-venv" = {
  exec = "echo hello";
  after = [ "devenv:python:virtualenv" ];
};
```

Previously, the only way to do this was to also use `before` to add this dependency to the graph, usually via `devenv:enterShell`. This change also allows running tasks after `devenv:enterShell`.

Fixes #1557.

##### TODO

While this change makes setting up tasks more predictable, we'll need to make this behavior configurable to support one-off tasks. You might want to run _just_ the specified task, or all the tasks _up_ to the specified task.

This also opens up the question of different types of constraints, e.g. _ordering_ vs _dependency_. systemd has `After` and `Requires`.

- [x] Update existing tests

### DIFF
--- a/devenv-tasks/src/lib.rs
+++ b/devenv-tasks/src/lib.rs
@@ -1321,6 +1321,51 @@ mod test {
         Ok(())
     }
 
+    // Ensure that tasks before and after a root are run in the correct order.
+    #[tokio::test]
+    async fn test_non_root_before_and_after() -> Result<(), Error> {
+        let script1 = create_basic_script("1")?;
+        let script2 = create_basic_script("2")?;
+        let script3 = create_basic_script("3")?;
+
+        let tasks = Tasks::new(
+            Config::try_from(json!({
+                "roots": ["myapp:task_2"],
+                "tasks": [
+                    {
+                        "name": "myapp:task_1",
+                        "command": script1.to_str().unwrap(),
+                        "before": [ "myapp:task_2"]
+                    },
+                    {
+                        "name": "myapp:task_2",
+                        "command": script2.to_str().unwrap()
+                    },
+                    {
+                        "name": "myapp:task_3",
+                        "after": ["myapp:task_2"],
+                        "command": script3.to_str().unwrap()
+                    },
+                ]
+            }))
+            .unwrap(),
+        )
+        .await?;
+        tasks.run().await;
+
+        let task_statuses = inspect_tasks(&tasks).await;
+        let task_statuses = task_statuses.as_slice();
+        assert_matches!(
+            task_statuses,
+            [
+                (name1, TaskStatus::Completed(TaskCompleted::Success(_, _))),
+                (name2, TaskStatus::Completed(TaskCompleted::Success(_, _))),
+                (name3, TaskStatus::Completed(TaskCompleted::Success(_, _)))
+            ] if name1 == "myapp:task_1" && name2 == "myapp:task_2" && name3 == "myapp:task_3"
+        );
+        Ok(())
+    }
+
     #[tokio::test]
     async fn test_dependency_failure() -> Result<(), Error> {
         let failing_script = create_script("#!/bin/sh\necho 'Failing task' && exit 1")?;
--- a/devenv-tasks/src/lib.rs
+++ b/devenv-tasks/src/lib.rs
@@ -434,7 +434,7 @@ impl Tasks {
 
             for dep_name in &task_state.task.after {
                 if let Some(dep_idx) = task_indices.get(dep_name) {
-                    edges_to_add.push((*dep_idx, index));
+                    edges_to_add.push((index, *dep_idx));
                 } else {
                     unresolved.insert((task_state.task.name.clone(), dep_name.clone()));
                 }
@@ -442,7 +442,7 @@ impl Tasks {
 
             for before_name in &task_state.task.before {
                 if let Some(before_idx) = task_indices.get(before_name) {
-                    edges_to_add.push((index, *before_idx));
+                    edges_to_add.push((*before_idx, index));
                 } else {
                     unresolved.insert((task_state.task.name.clone(), before_name.clone()));
                 }
@@ -450,7 +450,7 @@ impl Tasks {
         }
 
         for (dep_idx, idx) in edges_to_add {
-            self.graph.add_edge(dep_idx, idx, ());
+            self.graph.add_edge(idx, dep_idx, ());
         }
 
         if unresolved.is_empty() {
@@ -493,7 +493,7 @@ impl Tasks {
             for edge in self.graph.edges(old_node) {
                 let target = edge.target();
                 if let Some(&new_target) = node_map.get(&target) {
-                    subgraph.add_edge(new_node, new_target, ());
+                    subgraph.add_edge(new_target, new_node, ());
                 }
             }
         }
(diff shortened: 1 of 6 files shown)


devenv-tasks/src/lib.rs
```diff
@@ -1196,6 +1193,57 @@ mod test {
...
+            Config::try_from(json!({
+                "roots": ["myapp:task_1"],
+                "tasks": [
+                    {
+                        "name": "myapp:task_1",
+                        "command": script1.to_str().unwrap(),
+                    },
+                    {
+                        "name": "myapp:task_3",
+                         "after": ["myapp:task_1"],
```
domenkozar, line 1218:
```suggestion
                        "after": ["myapp:task_1"],
```

src/modules/languages/python.nix
```diff
@@ -459,25 +459,24 @@ in
         description = "Initialize Python virtual environment";
         exec = initVenvScript;
         exports = [ "PATH" "VIRTUAL_ENV" ];
+        after = [ "devenv:enterShell" ];
```
domenkozar, line 462: Shouldn't this be before?

sandydoo: True!

src/modules/integrations/git-hooks.nix
```diff
@@ -34,10 +34,14 @@ in
     packages = lib.mkAfter ([ config.git-hooks.package ] ++ (config.git-hooks.enabledPackages or [ ]));
     tasks = {
       # TODO: split installation script into status + exec
-      "devenv:git-hooks:install".exec = config.git-hooks.installationScript;
-      "devenv:git-hooks:run".exec = "pre-commit run -a";
-      "devenv:enterShell".after = [ "devenv:git-hooks:install" ];
-      "devenv:enterTest".after = [ "devenv:git-hooks:run" ];
+      "devenv:git-hooks:install" = {
+        exec = config.git-hooks.installationScript;
+        before = [ "devenv:git-hooks:install" ];
```
domenkozar, line 39: devenv:enterShell

src/modules/integrations/git-hooks.nix
```diff
@@ -34,10 +34,14 @@ in
...
+      };
+      "devenv:git-hooks:run" = {
+        exec = "pre-commit run -a";
+        before = [ "devenv:git-hooks:run" ];
```
domenkozar, line 43: devenv:enterShell

sandydoo:
> While this change makes setting up tasks more predictable, we'll need to make this behavior configurable to support one-off tasks. You might want to run just the specified task, or all the tasks up to the specified task.

This also opens up the question of different types of constraints, e.g. ordering vs dependency. systemd has After and Requires.

We need to brainstorm on this one. 

Only doing incoming (outgoing) edges means that only `before` (`after`) works for adding a task to the graph.

Using undirected lets us traverse the entire graph, at the cost of not letting us run subgraphs of tasks. We can't tell the difference between ordering hints and requirements.

In this example, if we run `devenv:enterShell` in the context of launching a shell, we want to run every task reachable from it. This includes the tasks specified using `after`: `app:afterVenv` and `app:afterShell`.

```mermaid
graph TD;
    devenv:python:virtualenv-->app:afterVenv;
    devenv:python:virtualenv-->devenv:enterShell;
    devenv:enterShell-->app:afterShell;
    devenv:git-hooks:install-->devenv:enterShell;
```

But if we run just `app:afterVenv`, we may want to only run that and `devenv:python:virtualenv`.

domenkozar:
@sandydoo maybe hooks should do the former while `devenv tasks run` do the latter?

domenkozar:
@sandydoo shall we open an issue and merge this?

sandydoo:
> @sandydoo maybe hooks should do the former while `devenv tasks run` do the latter?

Yeah, we need a CLI flag to modify the "traversal strategy"

PR #1590, opened 2024-11-14
Files

bobvanderlinden:
Follow-up on https://github.com/cachix/devenv/pull/142

This adds `files`, which allows creating (=linking) files into the devenv directory. This adds the following options:

```
files.<name>.ini
files.<name>.json
files.<name>.yaml
files.<name>.toml
files.<name>.text
```

Only one of these options may be set for each file.

`name` indicates the name/path of the file. For example:

```
files."some/file.txt".text = ''
  this is some file
'';
```

Will create a file `$DEVENV_ROOT/some/file.txt`.

The test `tests/files/devenv.nix` shows more basic examples of each supported format.

I'm on the fence whether the internal options are useful. The intention was to still be able to override the format ini that has different options, even though this isn't supported by the docs.

I did drop the support for `files.<name>.executable`, it didn't make much sense for most of the formats. It could've been for `text`, but then I'd rather see something like `script` that uses an `interpreter` option to be used in the shebang. That said, it's still possible to create such a script with the internal options.

### DIFF
--- a/docs/reference/options.md
+++ b/docs/reference/options.md
@@ -1989,6 +1989,153 @@ lazy attribute set of anything
 
 
 
+## files
+
+
+
+A set of files that will be linked into devenv root.
+
+
+
+*Type:*
+attribute set of (submodule)
+
+
+
+*Default:*
+` { } `
+
+*Declared by:*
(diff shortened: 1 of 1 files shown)


domenkozar:
Thank you :heart_eyes:

PR #1599, opened 2024-11-19
devenv-run-tests: switch directories in between tests

sandydoo:


sandydoo:
I've figured out what's going on with this test. `git-hooks` installs the hooks in the `git` root dir, so the test/example hooks end up running against the whole devenv repo.
Either we add some tweakable option to `git-hooks`, or we could `cp` the whole  directory with the test/example to `tmp`, run `git init`, and do our stuff without worry.

domenkozar:
Maybe we just need to `git init`?

sandydoo:
> Maybe we just need to `git init`?

Yes, but it would be best to do so in a temporary directory.

sandydoo:
This doesn't fully fix the wasm test, but it's a good cleanup to merge.

PR #1600, opened 2024-11-20
Added a uv pyproject example

whowatchlist:
Instead of using `uv pip`, this example uses uv's project setup


domenkozar:
Seems like we hit 256 jobs limit with this example. Not sure yet what to do :)

PR #1612, opened 2024-11-24
R: install packages from DESCRIPTION file

pchabros:
PR works but I need some help:

1. I've created `r/` subdirectory to split code into separate files, but it broke generated docs - can I fix it somehow or I should not create subdirectory?
2. I wanted `descriptionFile.path` to be relative path (`./DESCRIPTION`) or just a string (`"./DESCRIPTION"`) but after concatenating it with `config.devenv.root` I got `error: access to absolute path '/home' is forbidden in pure evaluation mode` - how can I move absolute path to nix-store? Or somehow evaluate relative path in context of `devenv.nix` instead of `modules/languages/r.nix`?
3. I've created [utils/strings.nix](https://github.com/cachix/devenv/compare/main...pchabros:r-add-description-packages?expand=1#diff-22bcaf02944d43b61868f4729b08b34380402fced78c0062374eaf16312cf500) with some generic functions. Should I put it somewhere else?


b-rodrigues:
Sorry, I’m very new to devenv, but am involved quite a bit with the R ecosystem in Nix. Could you explain what this would achieve? My understanding was that devenv provides the right build dependencies to build any R package, so wouldn’t trying to install an R packages using devenv result in its dependencies also be pulled from CRAN, as they’re listed in "depends"?

pchabros:
No, it doesn't pull anything from CRAN. It just automatically adds all packages listed in `DESCRIPTION` file to `packages`. So, let's assume that you have the following [`DESCRIPTION`](https://github.com/pchabros/devenv/blob/r-add-description-packages/examples/r/DESCRIPTION) file, than instead of copying and pasting dependencies to `devenv.nix`:
```nix
{
  languages.r.enable = true;
  packages = with pkgs.rPackages; [
    readr
    stringr
    data_table
    yaml
    testthat
  ];
}
```
you can just enable this option:
```nix
{
  languages.r = {
    enable = true;
    descriptionFile = {
      path = "${./DESCRIPTION}";
      installPackages.enable = true;
    };
  };
}
```

src/modules/languages/r/default.nix
```diff
@@ -21,9 +25,23 @@ in
         description = "The radian package to use.";
       };
     };
+    descriptionFile = {
```
domenkozar, line 28: What do you think if we go with `description.enable` that would also install the packages? Not sure it makes a lot of sense to enable it without doing anything.

src/modules/languages/r/descriptionFile.nix
```diff
@@ -0,0 +1,42 @@
...
+  inherit (strings)
+    matches
+    extractWithRegex
+    splitWithRegex
+    ;
+in
+rec {
+  getSection =
+    section: descriptionFile:
+    extractWithRegex ".*${section}:\n +([a-zA-Z0-9., (>=)\n]*)\n[A-Z].*" descriptionFile;
```
Turrentianos, line 19: I was testing out your changes and came across some DESCRIPTION files that define sections (e.g. Suggested, Import, Depends) and their packages on a single line instead of putting every package on a new line see for example [MASS](https://github.com/cran/MASS/blob/master/DESCRIPTION#L6-L8). I would change the pattern:

```suggestion
    extractWithRegex ".*${section}:\n? ?+([a-zA-Z0-9., (>=)\n]*)\n[A-Z].*" descriptionFile;
```

PR #1626, opened 2024-12-05
Add option to auto-activate uv

leiserfg:


domenkozar:
Shouldn't that be on by default?

leiserfg:
I just mimicked the options of `poetry`, `uv` environment does not need to be active, one can always use it via `uv run python something` is just not very ergonomic.

leiserfg:
Now I'm double-checking the code, and maybe the problem happens because the task to sync uv, happens before the one for activating venv? that would explain why uv is not using the correct python (it is lacking the libraries defined in `libraries`).

domenkozar:
Can you provide a way to reproduce the issue?

wgordon17:
FWIW, I get around this with using `venv` to create the virtual environment, and then `uv` just reuses that environment

```
  languages.python = {
    enable = true;
    package = pkgs.python312Full;
    venv.enable = true;
    uv = {
      enable = true;
      sync.enable = true;
    };
  };
```

abelREK:
@wgordon17 This does not seem to work properly with devenv shell. On a fresh system without using direnv to load in first, uv will not install the packages.

domenkozar:
> The following works:
> 
> ```nix
> { pkgs, ... }:
> {
>   languages.python = {
>     enable = true;
>     uv.enable = true;
>     uv.sync.enable = true;
>   };
>   enterShell = ''
>     source $DEVENV_STATE/venv/bin/activate
>   '';
> }
> ```
> 
> The issue is that the `devenv:python:virtualenv` task, that runs after `devenv:python:uv` recreates the virtualenv with `uv venv $DEVENV_STATE/venv`.

should we instead be just sourcing?

LorenzBischof:
As far as I understand `uv` can be used either with a `requirements.txt` or with a `uv.lock` file. I haven't tested this, but I would probably always use `source` when `uv.sync = true` and then also assert that `venv.enable = false`. If my hunch is correct this should allow both workflows.

leiserfg:
> The following works:
> 
> ```nix
> { pkgs, ... }:
> {
>   languages.python = {
>     enable = true;
>     uv.enable = true;
>     uv.sync.enable = true;
>   };
>   enterShell = ''
>     source $DEVENV_STATE/venv/bin/activate
>   '';
> }
> ```
> 
> The issue is that the `devenv:python:virtualenv` task, that runs after `devenv:python:uv` recreates the virtualenv with `uv venv $DEVENV_STATE/venv`.

That's what I'm doing locally, but I think it's better to have it as an option, that's the reason for the pr.

LorenzBischof:
Yes, I know. I only looked at your changes after posting my comment. It was in reply to @wgordon17, because that was the workaround I was previously using.

Is there anything that speaks against automatically sourcing if `uv.sync` is activated?

leiserfg:
Maybe not wanting to "activate" the venv, as some may like to use `uv run <script>`. Also, to have a similar api as `languages.python.poetry.activate.enable = true;`

LorenzBischof:
How about `languages.python.uv.sync.activate = true`? Or maybe updating `venv.enable` to only `source` and not use `uv venv` if `uv.sync` is enabled?

LorenzBischof:
I am not quite sure why the following test passes: https://github.com/cachix/devenv/blob/f8be0ed0141abab89cefb20d3d375740671fbee1/tests/python-uv-sync/devenv.nix#L9

If I run `devenv-run-tests --only python-uv-sync` it passes, however if I enter a shell with `cd tests/python-uv-sync && ../../result/bin/devenv shell` it correctly fails when importing:  
```
$ python -c 'import requests'
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'requests'
```

LorenzBischof:
Can we close this? `venv.enable` should work now.

PR #1629, opened 2024-12-06
ollama: new service

i-am-logger:
- had an issue with nixpkgs-fmt in devenv in order to be able to commit. (i'll recreate the commit if i need to use a different version)
- having an issue with running the test ,it keeps failing, even though ollama service is running ok. will appriciate a look.


i-am-logger:
fixed and now the tests are passing

image

examples/ollama/devenv.nix
```diff
@@ -0,0 +1,13 @@
+{ ... }:
```
domenkozar, line 1: We have hit the limit of examples as github jobs, so we have to remove it for now.

tests/ollama/devenv.nix
```diff
@@ -0,0 +1,11 @@
+{
+  services.ollama = {
+    enable = true;
+    address = "0.0.0.0";
+    port = 11434;
+    loadModels = [
+      "llama3.2:3b"
+    ];
+    # acceleration = "";
```
domenkozar, line 9:
```suggestion
```

domenkozar (changes requested):

examples/ollama/devenv.nix
i-am-logger: removed, thank you

domenkozar: This file still exists but shouldn't

Issue #1628, opened 2024-12-06
NIXPKGS_ALLOW_UNFREE=1 test fails for n8n


i-am-logger:
basically
```bash
devenv-run-tests --only n8n
```
and 

![image](https://github.com/user-attachments/assets/d318220d-8eb8-405c-b4e3-1fbfb2ec0f9f)

i-am-logger:
tried also 
- devenv.yaml: `allowUnfree: true`
- .envrc `NIXPKGS_ALLOW_UNFREE=1 use devenv`

still same error

sandydoo:
> devenv.yaml: allowUnfree: true

This option should work. The `devenv.yaml` should be placed in the same directory as the `devenv.nix` for the test/example.

i-am-logger:
i'll give it a shot

i-am-logger:
that worked! thanks

PR #1630, opened 2024-12-06
n8n: new service

i-am-logger:
having an issue #1628 will appriciate the help.


i-am-logger:
working, would appriciate a review

src/modules/services/n8n.nix
```diff
@@ -0,0 +1,86 @@
...
+      address = lib.mkOption {
+        type = types.str;
+        description = "Listen address";
+        default = "127.0.0.1";
+        example = "127.0.0.1";
+      };
+
+      port = lib.mkOption {
+        type = types.port;
+        default = 5432;
```
sandydoo, line 29: The default in their docs is `5678`. `5432` is the default for postgres, so we should avoid using that.

```suggestion
        default = 5678;
```

src/modules/services/n8n.nix
```diff
@@ -0,0 +1,86 @@
...
+      assertion = cfg.enable;
+      message = ''
+        To use n8n, you have to enable it. (services.n8n.enable = true;)
+      '';
+    }];
+    env = {
+      N8N_PORT = cfg.port;
+      N8N_LISTEN_ADDRESS = cfg.address;
+      WEBHOOK_URL = "${cfg.webhookUrl}";
+      N8N_CONFIG_FILES = "${configFile}";
```
sandydoo, line 66: This should work, no?

```suggestion
      WEBHOOK_URL = cfg.webhookUrl;
      N8N_CONFIG_FILES = configFile;
```

src/modules/services/n8n.nix
```diff
@@ -0,0 +1,86 @@
...
+    };
+
+    processes.n8n = {
+      exec = "${pkgs.n8n}/bin/n8n";
```
sandydoo, line 70: It's nice to have a `package` option in the module to let people override the package used.

tests/n8n/devenv.nix
```diff
@@ -0,0 +1,7 @@
+{
```
sandydoo, line 1: To test that it launches, you can add a `.test.sh` script in this folder and `curl` a health endpoint of some kind. There are lots of examples of this in the repo.

examples/n8n/devenv.nix
```diff
@@ -0,0 +1,2 @@
+{ ... }:
```
sandydoo, line 1: The `n8n` can be removed in favor of the test.

src/modules/services/n8n.nix
```diff
@@ -0,0 +1,86 @@
...
+      port = lib.mkOption {
+        type = types.port;
+        default = 5432;
+        description = ''
+          The TCP port to accept connections.
+        '';
+      };
+
+      webhookUrl = lib.mkOption {
+        type = lib.types.str;
```
sandydoo, line 36: If `null`, then the `WEBOOK_URL` env var won't be set at all, as long as you remove the quotes when setting the env var later.

```suggestion
        type = types.nullOr types.str;
```

src/modules/services/n8n.nix
```diff
@@ -0,0 +1,86 @@
...
+        default = "";
```
sandydoo, line 37:
```suggestion
        default = null;
```

sandydoo (changes requested):
Thanks, @i-am-logger. I've requested a bunch of minor changes, but otherwise looks good overall.

sandydoo:
@i-am-logger, I think quite a few of the requested changes are also applicable to your previous PR. I'll be happy to review it once updated.

PR #1633, opened 2024-12-07
postgres: Add setupSchemaScript option for Schema Setup

i-am-logger:
Path to a script that will set up or update the PostgreSQL database schema.

This script must be idempotent, meaning it can be run multiple times without causing unintended side effects. If you change your schema dynamically, ensure that this script handles such cases gracefully to maintain database integrity.

this is how i use it:
```nix
let
  db-name = "database";
in
{
  services.postgres.setupSchemaScript = ''
      ${builtins.readFile ./setup-local-postgres.sh} // create db and users (idempotent)
      cd infra/sqitch
      sqitch deploy --db-name ${db-name}
    '';
}
```


domenkozar:
This makes me think that `initialScript` should be named `initialSchema` and this should be `initialScript`.

i-am-logger:
> This makes me think that `initialScript` should be named `initialSchema` and this should be `initialScript`.

I Agree. 

I much rather have an idempotent script that runs every time Postgres starts in devenv than a script that runs one time... this is why I moved away from using `InitialScript` as its quite limited and can cause some headaches around updating the DB. 

Maybe I should rename it from `InitialSchema` to `UpdateScript` ?

PR #1634, opened 2024-12-07
feat(ocaml): add extra packages for opam

anntnzrb:
This introduces a `extraPackages` option for OCaml module. This addition allows users to specify additional OCaml packages they want to include in their environment. The primary use case for this feature is to enable users to easily add extra OCaml-specific packages, such as `base` or others.

## Feedback

I would like feedback on the following points:
+ Naming convention: Should the option name `extraPackages` be more specific, such as `extraOcamlPackages`, to clarify its purpose? If the latter is the case, should we force `pkgs.ocamlPackages` for less verbosity?
+ General usability: Does this approach align with user expectations and other modules?

## Use

This is how I am using this new option:
```nix
{ ... }:
{
  imports = [ inputs.devenv.flakeModule ];
  perSystem =
    { ... }:
    {
      devenv.shells.default = {
        languages = {
          nix.enable = true;
          ocaml = {
            enable = true;
            extraPackages = with pkgs.ocamlPackages; [ base ];
          };
        };

        packages = with pkgs; [ ... ];
      };
    };
}
```

### DIFF
--- a/src/modules/languages/ocaml.nix
+++ b/src/modules/languages/ocaml.nix
@@ -14,6 +14,13 @@ in
         default = pkgs.ocaml-ng.ocamlPackages;
         defaultText = lib.literalExpression "pkgs.ocaml-ng.ocamlPackages_4_12";
       };
+
+    extraPackages = lib.mkOption
+      {
+        type = lib.types.listOf lib.types.package;
+        description = "Extra OCaml packages to use";
+        default = [ ];
+      };
   };
 
   config = lib.mkIf cfg.enable {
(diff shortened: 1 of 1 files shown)


domenkozar:
I'm a bit confused, how is this different to `packages = [ ... ];`?

anntnzrb:
> I'm a bit confused, how is this different to `packages = [ ... ];`?

I may be the one confused. I thought this was an option to change the origin of the OCaml package set... I may have overlooked this because the type of this option is not `types.listOf types.package`.

Being now the case that I've been mistaken, this may be closed. However, I still don't quite get how would I add some other libraries using this (syntax?), for example: `pkgs.ocamlPackages.base`?

domenkozar:
`packages = [ pkgs.ocamlPackages.base ];`

PR #1657, opened 2025-01-02
docs: Use defaultText

roberth:
This is a bit more useful and passes the stricter rules I apply in flake-parts docs.

Would you be interested in having such a check here in devenv?


domenkozar:
A check would be nice!

PR #1660, opened 2025-01-06
cachix: use `cachix.enable` to disable cache setup logic

sandydoo:
The previous logic assumed that cache setup should be disabled if `cachix.pull` is empty, but that `devenv` is always a default cache. We now read in `cachix.enable` and use that to enable/disable setting up caches.

### DIFF
--- a/src/modules/cachix.nix
+++ b/src/modules/cachix.nix
@@ -13,7 +13,8 @@ in
     pull = lib.mkOption {
       type = lib.types.listOf lib.types.str;
       description = "Which Cachix caches to pull from.";
-      default = [ "devenv" ];
+      default = [ ];
+      defaultText = lib.literalExpression ''[ "devenv" ]'';
     };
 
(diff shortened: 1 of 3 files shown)


src/modules/cachix.nix
```diff
@@ -12,13 +12,13 @@ in
 
     pull = lib.mkOption {
       type = lib.types.listOf lib.types.str;
-      description = "What caches to pull from.";
-      default = [ ];
+      description = "Which Cachix caches to pull from.";
+      default = [ "devenv" ];
```
domenkozar, line 16: That means if someone sets a cache, devenv won't be used anymore, which is surprising.

domenkozar (changes requested):

src/modules/cachix.nix
sandydoo: Ah, yeah, I thought they'd get merged for some reason.

PR #1700, opened 2025-02-04
`devenv generate`

domenkozar:


NotAFile:
This feature flagrantly violates GDPR in every aspect:
- There is no legal basis for sending all of your local files to a server. In particular there is no reasonable expectation that running a local command would result in a third party processing your data in this way. This is neither clarified to the user, nor is any consent gathered.
- None of the information that under articles 12, 13, 14, *must* be provided to data subjects are present. Particularly egregious is the lack of the identity of the data controller, the types of data collected and the storage duration and legal based thereof.
- While some types of minimal telemetry may be considered necessary in order to ensure performance of the service, this is limited to only that purpose. In particular, *if* you are storing any information for any other purpose, which we don't know because you are (illegally) not providing that information, say to train AI models, this requires explicit opt-in, not opt-out.
- As a reminder, consent and compliance is not retroactive. If data is gathered without necessary legal basis, this data *must be deleted* and can not be further retained or used. This also applies to processing data later for a purpose for which it was not originally gathered.

Note that the cachix "privacy policy" too appears to lack any of the legally required information in the legally required specificity. I highly urge you to correct this too.

NotAFile:
On a broader note, I should note that gdpr requires that you have processes in place to ensure compliance that are in proportion with the sensitivity of the data you are processing. The fact that you are doing something as invasive as uploading people's private files apparently without even a cursory review of the regulatory requirements and privacy considerations, ones which are law in the country you are based no less, is alone a serious deficiency.

Issue #1525, opened 2024-10-15
`devenv up` stuck


MarceColl:
**Describe the bug**
When I run devenv up in a brand new arch linux machine it gets stuck and never finishes at the `Building shell` stage. The same devenv.nix works in a separate arch linux machine, also up-to-date.

**To reproduce**
Devenv files: https://gist.github.com/MarceColl/bf5f924cb23b16f2efab99ac5775f451

```
[seikaku@seikaku seikaku]$ uname -a
Linux seikaku 6.6.52-1-lts #1 SMP PREEMPT_DYNAMIC Wed, 18 Sep 2024 19:02:04 +0000 x86_64 GNU/Linux
```

I also need `--max-jobs=4` in this case for some reason.

```
[seikaku@seikaku seikaku]$ devenv up --max-jobs=4
• Building processes ...
• Using Cachix: devenv
these 3 derivations will be built:
  /nix/store/q37ckd0zi65qfzkzvzs38akxhbijl3ki-devenv-profile.drv
  /nix/store/8l7v5p81gsfb3zlf27r4sadnabr19nw9-process-compose.yaml.drv
  /nix/store/7zd2ysl9i2fgxrk9gk2drs2v159kxh21-devenv-up.drv
building '/nix/store/q37ckd0zi65qfzkzvzs38akxhbijl3ki-devenv-profile.drv'
created 833 symlinks in user environment
building '/nix/store/8l7v5p81gsfb3zlf27r4sadnabr19nw9-process-compose.yaml.drv'
building '/nix/store/7zd2ysl9i2fgxrk9gk2drs2v159kxh21-devenv-up.drv'
✔ Building processes in 6.1s.
• Starting processes ...
• Building shell ...
building '/nix/store/vjzm8xxzjg7x0y6xkfykndyiqlgxywmy-devenv-shell-env.drv'
```

User seikaku is part of `nixbld`.

```
[seikaku@seikaku seikaku]$ groups
wheel nixbld seikaku
```

**Version**
 
```
devenv 1.3.0 (x86_64-linux)
```

sandydoo:
Does `devenv shell` work? Also, run it with the `--verbose` flag to see the nix command it gets stuck on.

MarceColl:
Sorry, I'm super sleep deprived, I wanted to send this part as well, it gets stuck on 
```
devenv readFile: '«github:cachix/devenv/ef61728d91ad5eb91f86cdbcc16070602e7afa16»/src/modules/latest-version'
[...]

PR #1714, opened 2025-02-11
eval-cache: prevent deadlock reading pipes on 1-cpu machines

### LINKED ISSUE #1715: Building the shell hangs on machines with 1 cpu
> I can confirm that this is indeed related to either CPU or RAM as @dunderrrrrr said.
> 
> After switching to a 2 vcpu 32gb memory machine, everything ran smoothly.
> 
> As I already confirmed it got stuck on a 1 vcpu 16gb memory machine and 16gb is a pretty high number for the use case I was testing, my guess is that for some reason devenv is not working properly on 1 vcpu machines. 

 _Originally posted by @dsalaza4 in [#1525](https://github.com/cachix/devenv/issues/1525#issuecomment-2632423122)_

sandydoo:
Fixes #1715.
Related to https://github.com/cachix/devenv/issues/1525.

Makes sure we don't mix tokio threads with non-tokio processes.

### DIFF
--- a/devenv-eval-cache/src/command.rs
+++ b/devenv-eval-cache/src/command.rs
@@ -1,7 +1,7 @@
 use futures::future::join_all;
 use miette::Diagnostic;
 use sqlx::SqlitePool;
-use std::io::{self, BufRead, BufReader, Read};
+use std::io::{self, BufRead, BufReader};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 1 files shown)


dsalaza4:
Thank you so much @sandydoo @domenkozar 🫶 

Looking forward to a new release so we can benefit from this on the container!

Issue #67, opened 2022-11-17
Support defining devenvs outside the repository tree


domenkozar:
It's not always possible to add devenv.* files to the repository, so in those cases we can recommend the best workflow.

domenkozar:
I'm thinking something like `devenv -p myproject <command>` where it will load the file from `~/.devenv/projects/myproject`.

madjar:
As a starting point, here is a setup I have:
- Ran `devenv init` in a folder, added everything devenv-related (both config files and cache files) to `.git/info/exclude`. This got some complains from nix about ignored files on the first run, and then no issues afterward. This works well, but nothing is committed, and it's vulnerable to an aggressive `git clean`.
- A variant of this is to have the same excluded files, but instead of having the actual files in the folder, we have links to another folder (and that other folder can have git if you want). The links can be symlinks, or some devenv-specific import.

I think you'll always want something along these lines, because you always want at least the custom `.envrc`, so you need _some_ files in the folder.

Something like `devenv -p myproject` has the benefit that you need nothing in the current folder, but the drawback that it changes the normal workflow, as well as preventing usage of direnv.

bcardiff:
Something I would like is to be able to track/share the devenv configuration with others until things mature enough to transition them to the project.

I would be more than satisfied if the workflow involves adding a symlink to an .envrc from the project working copy to the direnv config repo, and add that single file to the project ignore list. Currently the .envrc watch file locally and loads the environment from the local with print-dev-env from the current directory. Is there a way to tweak that? Maybe `devenv -f /path/to/config/directory`. It could also mean that we can scope devenv files somewhere else in the repo if wanted.

workflow:
> Something I would like is to be able to track/share the devenv configuration with others until things mature enough to transition them to the project.
> 
> I would be more than satisfied if the workflow involves adding a symlink to an .envrc from the project working copy to the direnv config repo, and add that single file to the project ignore list. Currently the .envrc watch file locally and loads the environment from the local with print-dev-env from the current directory. Is there a way to tweak that? Maybe `devenv -f /path/to/config/directory`. It could also mean that we can scope devenv files somewhere else in the repo if wanted.

Something like `devenv -f /path/to/config/directory` would also enable the use case where you have a shared devenv referenced as a basis in many similar repos that don't need any overrides as a default, like the following structure:

```
  organization/
    .envrc -> loads the shared base devenv defined in organization/dev-env/
    dev-env/
       devenv.nix
       ...
[...]

PR #1721, opened 2025-02-12
cli: init with template

### LINKED ISSUE #1718: Devenv init template support
## Problem

I'm developing [drupal-devenv](https://gitlab.com/woolwichweb/drupal-devenv), which builds on devenv to provide Drupal development environments.

I feel the Getting Started process is less than ideal, users have to run `devenv init`, then edit our repo into their project's inputs and add the corresponding import. If you're new to devenv, this is all a bit opaque and gives a first impression that it requires a lot of boilerplate.

Meanwhile, the competition is [ddev](https://ddev.com), where getting started is `ddev config` followed by `ddev start`. While the config step does ask some questions, the whole process is short and streamlined.

## Solution

If we could provide users with a template, with pre-configured inputs and such, they could get started just as easily as with ddev. For example, the usage might look like:

```bash
devenv init --template gitlab:woolwichweb/drupal-devenv
```

Then users just have to run `devenv up` to get a fully working Drupal development environment.

Related: #1705

LorenzBischof:
Closes #1718

### DIFF
--- a/docs/guides/.pages
+++ b/docs/guides/.pages
@@ -1,3 +1,4 @@
 nav:
    - Using with Flakes: using-with-flakes.md
    - Using with flake.parts: using-with-flake-parts.md
+   - Creating templates: creating-templates.md
--- a/docs/guides/creating-templates.md
+++ b/docs/guides/creating-templates.md
@@ -1,0 +1,92 @@
+The following guide describes how to extend devenv and provide templates for other users. It is possible to bootstrap `devenv.nix` with custom content, change defaults or even define custom devenv options.
+
+## Templates
+
+To provide a devenv template, add the following to your `devenv.nix` file:
+```nix title="devenv.nix"
+  templates.rust = {
+    path = ./rust;
+    description = "A simple Rust/Cargo project";
+    welcomeText = ''
+      # Simple Rust/Cargo Template
+      ## Intended usage
+      The intended usage is...
+
+      ## More info
+      - [Rust language](https://www.rust-lang.org/)
+      - [Rust on the NixOS Wiki](https://wiki.nixos.org/wiki/Rust)
+      - ...
+    '';
+  };
+
+  templates.default = config.templates.rust;
+}
+```
+
+A template has the following attributes:
+- `description`: A one-line description of the template, in CommonMark syntax.
+- `path`: The path of the directory to be copied. This directory can be created with `devenv init rust`.
+- `welcomeText`: A block of markdown text to display when a user initializes a new project based on this template.
+
+Template consumers can then use the following command to use the `default` template:
+```
+devenv init --template github:owner/repo
+```
+
+The reference to the Github repository is called a Flake reference. Some more examples:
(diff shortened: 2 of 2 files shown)


liammcdermott:
Thanks for this! I tested this PR using:

```bash
devenv init --template github:LorenzBischof/devenv-issue-1718
```

And it worked as expected.

sandydoo (approved):

LorenzBischof:
If this feature is being considered, we should probably also merge the `.gitignore` file. 

I also thought about maybe forcing a template prefix? We could default to `devenv-default` and if that does not exist list all the templates that start with `devenv-`? This would ensure that there can be both flake and devenv templates at the same time.

We could also think about adding a `templates` option to the `devenv.nix` file, but that would require committing the generated `flake.nix` or reimplementing the whole template downloading logic ourselves.

sandydoo:
@LorenzBischof, this is fine as it is. I don't particularly like coupling ourselves to the flake API, but this feels like a useful feature for some and we can re-implement the behaviuor separately if needed down the line.

Could you please rebase this on main for the builds to pass and I'll merge it.

domenkozar:
I don't think we should be using flakes for templates, we should add `templates` top-level attribute to devenv if we want to support it

LorenzBischof:
Not exactly sure why this is not working. I know you are working on Snix, but I am enjoying the challenge :smile: 

```
❯ result/bin/devenv init --template github:LorenzBischof/devenv-issue-1718/templates
Cloning into '/tmp/devenv-clone'...
error:
       error: '/tmp/devenv-clone/.devenv.flake.nix' does not exist
Error:   × Command `/nix/store/ni24hs509kjhi1ckm6wxhhj8h333437l-nix-devenv-2.24.0pre20250312_f3f44b2/bin/nix --show-trace --extra-
  │ experimental-features nix-command --extra-experimental-features flakes --option warn-dirty false --keep-going --max-jobs 4 --option
  │ eval-cache false flake init -t /tmp/devenv-clone` failed with with exit code 1

❯ ls /tmp/devenv-clone -A1
.devenv
.devenv.flake.nix
devenv.lock
devenv.nix
devenv.yaml
example
.git
.gitignore
LICENSE
README.md
vscode-integration.nix
```

LorenzBischof:
The templates can now be specified in the `devenv.nix` file under the top-level `templates` option. The basic functionality is implemented, but still needs some cleanup. 

Since I have no idea how far along the Snix implementation is, I do not know if it makes sense to continue. Would you consider merging this before the rewrite?

domenkozar:
> The templates can now be specified in the `devenv.nix` file under the top-level `templates` option. The basic functionality is implemented, but still needs some cleanup.
> 
> Since I have no idea how far along the Snix implementation is, I do not know if it makes sense to continue. Would you consider merging this before the rewrite?

Sorry for taking so long, I'm going to fix https://github.com/cachix/devenv/issues/67 soon and then we can bring this one in! I don't think it it will require much changes from what you've done so far.

PR #1734, opened 2025-02-15
Reverted the addition of `generate` command due to security concerns

trm109:
**See [my original issue](https://github.com/cachix/devenv/issues/1733) for full reasoning.**

Confirmed that the problem code in `devenv/src/devenv.rs` was removed, along with documentation and CLI command.

Performed minimal testing via `devenv init` and `devenv update` per https://devenv.sh/community/contributing/#creating-development-project and saw no issues.

Will continue to test in actual workloads throughout the week. If I find any other issues I will report back.

I'll append this to the issue as well, but if you want to use the 1.4.0 with generate removed and you are using NixOS, here you go:

```nix
...
# Sorry for the gross indentation.
  environment.systemPackages = with pkgs; [
    (devenv.override 
      (old: 
        {
          rustPlatform = old.rustPlatform // {
            buildRustPackage = args: old.rustPlatform.buildRustPackage (
              args // {
                verison = "1.4.0";
                src = old.fetchFromGitHub {
                  owner = "trm109";
                  repo = "devenv-no-ai";
                  rev = "5a7d859293c24e930942b0a777bddb9551476975";
                  hash = "sha256-BFdIoDLJnyrw/ba42Gd5cIJNagc+Rkofh5xALbhVzLg=";
                };
                cargoHash = "sha256-kJ5N34SoIUtDxHDiZKo4+eQiUf11ibvzA0/4H5olQdY=";
              }
            );
          };
        }
      )
    )
  ];
...
```
For those of you who aren't on NixOS, I apologize but I don't know how to make my patch available via the command line. If someone could reply to the PR with a patch, I would be very appreciative.

**Please update this PR if you run into any issues with my regression.**

_Thank you to lilyball on discourse.nixos.org for the [template](https://discourse.nixos.org/t/is-it-possible-to-override-cargosha256-in-buildrustpackage/4393/10) for patching buildRustPackage sources!_

PR #1740, opened 2025-02-18
WIP: Move devenv generate into a separate binary

WeetHet:
This makes `devenv-generate`, so that `devenv generate` just delegates to it

`devenv-generate`still needs to be packaged but it is just a `--package=devenv-generate` addition, so it would probably be easy to do

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -612,6 +612,28 @@ dependencies = [
 ]
 
 [[package]]
+name = "devenv-generate"
+version = "1.4.1"
+dependencies = [
+ "binaryornot",
+ "clap",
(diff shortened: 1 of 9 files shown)


domenkozar:
Nice, I'll work on top of this :)

WeetHet:
> Nice. I think shipping it as a separate package makes sense, that removes a lot of the issues brought up by the feedback in #1734

This wasn't finished, like I haven't added a way to build devenv-generate with nix and some tests failed in CI. Nevertheless, I'm unsure if I would've gotten around to it, so I would rather prefer someone else does that

Issue #1667, opened 2025-01-10
new meilisearch version (1.11.3) fails to start through `devenv up`


realugi:
**Describe the bug**
Once I upgraded my meilisearch version from 1.7.6 to 1.11.3 (through upgrading nixpkgs), I encountered an error when starting it via `devenv up`. The error says that the value `1` is invalid for the `--no-analytics`  command line parameter, which in devenv is set via the environment variable `MEILI_NO_ANALYTICS`.

**To reproduce**
1. create the following `flake.nix` file:
<details>
<summary>see flake.nix</summary>

```nix
{
    inputs = {
        nixpkgs.url = "github:NixOS/nixpkgs/nixos-24.11";
        devenv.url = "github:cachix/devenv";
        devenv.inputs.nixpkgs.follows = "nixpkgs";
    };

    nixConfig = {
        extra-trusted-public-keys = "devenv.cachix.org-1:w1cLUi8dv3hnoSPGAuibQv+f9TZLr6cv/Hm9XgU50cw=";
        extra-substituters = "https://devenv.cachix.org";
    };

    outputs = inputs@{ self, nixpkgs, devenv, ...}:
        let
            system = "x86_64-linux";
            pkgs = nixpkgs.legacyPackages."${system}";
        in
        {
            devShells.${system}.default = devenv.lib.mkShell {
                inherit inputs pkgs;
                modules = [
                    ({ pkgs, config, ... }: {
                        services.meilisearch.enable = true;
                    })
                ];
            };
            devenv-up = self.devShells.${system}.default.config.procfileScript;
        };
}
```
</details>

2. create the following `flake.lock` file (for reproducability):
<details>
<summary>see flake.lock</summary>

```json
{
  "nodes": {
    "cachix": {
      "inputs": {
        "devenv": [
          "devenv"
        ],
        "flake-compat": [
          "devenv"
        ],
        "git-hooks": [
          "devenv"
        ],
        "nixpkgs": "nixpkgs"
      },
      "locked": {
        "lastModified": 1728672398,
        "narHash": "sha256-KxuGSoVUFnQLB2ZcYODW7AVPAh9JqRlD5BrfsC/Q4qs=",
        "owner": "cachix",
        "repo": "cachix",
        "rev": "aac51f698309fd0f381149214b7eee213c66ef0a",
        "type": "github"
      },
      "original": {
        "owner": "cachix",
        "ref": "latest",
        "repo": "cachix",
        "type": "github"
      }
    },
[...]

PR #1741, opened 2025-02-19
Fix/meilisearch 'error: invalid value '1' for '--no-analytics'

linusfri:
This pull request is related to https://github.com/cachix/devenv/issues/1667

### DIFF
--- a/src/modules/services/meilisearch.nix
+++ b/src/modules/services/meilisearch.nix
@@ -76,7 +76,7 @@ in
 
     env.MEILI_DB_PATH = config.env.DEVENV_STATE + "/meilisearch";
     env.MEILI_HTTP_ADDR = "${cfg.listenAddress}:${toString cfg.listenPort}";
-    env.MEILI_NO_ANALYTICS = toString cfg.noAnalytics;
+    env.MEILI_NO_ANALYTICS = if cfg.noAnalytics == true then "true" else "false";
     env.MEILI_ENV = cfg.environment;
     env.MEILI_DUMP_DIR = config.env.MEILI_DB_PATH + "/dumps";
     env.MEILI_LOG_LEVEL = cfg.logLevel;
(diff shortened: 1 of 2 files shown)


src/modules/services/meilisearch.nix
```diff
@@ -76,7 +76,7 @@ in
 
     env.MEILI_DB_PATH = config.env.DEVENV_STATE + "/meilisearch";
     env.MEILI_HTTP_ADDR = "${cfg.listenAddress}:${toString cfg.listenPort}";
-    env.MEILI_NO_ANALYTICS = toString cfg.noAnalytics;
+    env.MEILI_NO_ANALYTICS = if cfg.noAnalytics == true then "true" else "false";
```
sandydoo, line 79:
```suggestion
    env.MEILI_NO_ANALYTICS = lib.boolToString cfg.noAnalytics;
```

tests/meilisearch/devenv.nix
```diff
@@ -0,0 +1,36 @@
...
+      if [[ -z "$RUNNING" ]]; then
+        exit 1
+      else
+        exit 0
+      fi
+    '';
+
+    enterTest = ''
+      set -e
+
```
sandydoo, line 27: devenv sets this option by default.
```suggestion
```

tests/meilisearch/devenv.nix
```diff
@@ -0,0 +1,36 @@
...
+      wait_for_port ${toString port}
+
+      # Give meilisearch time to initialize
+      sleep 5
```
sandydoo, line 31: We usually use `timeout` + `until` to start fetching the endpoint immediately, instead of relying on a fixed delay. See the `.test.sh` files in `tests/` for examples. You can add it to the healthcheck script.

sandydoo (changes requested):

tests/meilisearch/devenv.nix
linusfri: I updated the test script with timeout + until so it should be okay now.

PR #1743, opened 2025-02-20
ci: use self-hosted x86_64-darwin runners

sandydoo:
Let's see if this speeds things up a bit or not.

### DIFF
--- a/.github/workflows/buildtest.yml
+++ b/.github/workflows/buildtest.yml
@@ -19,10 +19,10 @@ jobs:
       matrix:
         os:
           [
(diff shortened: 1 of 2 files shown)


domenkozar:
@sandydoo looks good?

sandydoo:
Yeah, individual jobs are 6x faster that the hosted `macos-13` runners. But we're running out of disk space now, which triggers a lot of GC runs.

sandydoo:
Ah, let's give it a whirl and revert if it sucks.

PR #1745, opened 2025-02-22
tasks: use `lib.getExe` to fetch the package binary by default

sandydoo:
`tasks.<name>.binary` is now an optional override, in case `lib.getExe` (i.e. `meta.mainProgram`) returns something that doesn't work.

This fixes an upcoming issue in nixpkgs/master where the pname for `bash` is changing to `bash-interactive` by default. Using the `pname` as the name of the binary will break tasks.

### DIFF
--- a/docs/reference/options.md
+++ b/docs/reference/options.md
@@ -83325,17 +83325,17 @@ list of string
 
 
 
-Override the binary name if it doesn’t match package name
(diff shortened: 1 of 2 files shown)


sandydoo:
`lib.getExe` is a better option, but now we need to rethink the assertion.

PR #1766, opened 2025-03-10
devenv: refresh the env if the gc root is invalid

sandydoo:


devenv/src/cnix.rs
```diff
@@ -144,19 +151,25 @@ impl Nix {
...
         let now_ns = get_now_with_nanoseconds();
         let target = format!("{}-shell", now_ns);
-        symlink_force(
-            &fs::canonicalize(gc_root).expect("to resolve gc_root"),
-            &self.devenv_home_gc.join(target),
-        );
+        if let Ok(resolved_gc_root) = fs::canonicalize(gc_root) {
+            symlink_force(&resolved_gc_root, &self.devenv_home_gc.join(target));
+        } else {
+            warn!(
```
domenkozar, line 168: Is this just a safety check in case `refresh_cached_output` doesn't take care of it?

Issue #1769, opened 2025-03-14
`devenv init target` creates `.devenv` at wrong location


LorenzBischof:
**Describe the bug**
`devenv init target` incorrectly creates  `.devenv/` inside the current working directory, instead of inside the target directory.

**To reproduce**
```
devenv init target
ls -al
```

**Version**

devenv 1.4.1 (x86_64-linux)

PR #1773, opened 2025-03-15
tests: add cli test for init with target path

LorenzBischof:
Tests for the issue described in https://github.com/cachix/devenv/issues/1769
Fixed by https://github.com/cachix/devenv/pull/1771

I had to remove the first test, because an interactive confirmation was added. The tests still worked in CI, because the command fails to allocate an interactive terminal.


domenkozar:
I'm not sure I understand why the CI fails

PR #1776, opened 2025-03-17
Warn before doing anything AI related

domenkozar:


devenv-generate/src/main.rs
```diff
@@ -144,7 +165,21 @@ async fn main() -> Result<()> {
         }
     };
 
-    info!("Generating devenv.nix and devenv.yaml, this should take about a minute ...");
+    if body.is_some() {
+        info!("Uploading project source code using `git ls-files` to https://devenv.new to analyze the environment ...");
```
jtojnar, line 169: Uploading → Will upload?

Maybe also add a link to privacy policy.

domenkozar:
@jtojnar how does this look to you? :)

devenv-generate/src/main.rs
```diff
@@ -144,7 +165,21 @@ async fn main() -> Result<()> {
         }
     };
 
-    info!("Generating devenv.nix and devenv.yaml, this should take about a minute ...");
+    if body.is_some() {
+        info!("Going to upload source code using `git ls-files` to https://devenv.new to analyze the environment using AI and generate devenv.nix...");
```
jtojnar, line 169: “upload source code using `git ls-files`“  sounds a bit like git will upload the files. Maybe something like “upload source code files from the current Git repository“, and have the details on the website in the privacy policy.

Presumably, this has different processing than the analytics whose privacy policy is on the website so it would need a separate document. Even something simple like the following would probably be fine:

- how the files are collected: `devenv generate` will collect all non-binary files in the project as listed by `git ls-files` and upload them to the server
- who collects it: "Devenv corporation"
- reason for processing: building a regression test suite for AI model prompt
- how long will the data be persisted: indefinitely
- basis for processing: explicit consent – though here it might be necessary to allow people to use the service with an option to not persist their data, for the consent to be considered freely given (see [Art. 7, section 4 of GDPR](https://eur-lex.europa.eu/legal-content/EN/TXT/?uri=CELEX:02016R0679-20160504#art_7)).

Of course, I am not a lawyer.

devenv-generate/src/main.rs
```diff
@@ -90,6 +91,26 @@ async fn main() -> Result<()> {
 
     log::init_tracing(level, cli.log_format);
 
+    // Show one-time telemetry warning if not disabled and not shown before
+    if !cli.disable_telemetry && !has_shown_telemetry_warning() {
+        println!("\nGenerating collects anonymous usage data sent to https://devenv.new (see privacy policy) to improve generation output.");
```
jtojnar, line 96: From the Privacy policy linked in the footer of the website, it is not really clear to me what data will be collected.

devenv-generate/src/main.rs
```diff
@@ -201,6 +236,19 @@ async fn main() -> Result<()> {
     Ok(())
 }
 
+/// Checks if the telemetry warning has been shown for this project
+fn has_shown_telemetry_warning() -> bool {
+    let dotfile = Path::new(".devenv/telemetry-warning-shown");
```
jtojnar, line 241: Maybe include `generate` in the name to make it clear it is specific to that feature.

devenv-generate/src/main.rs
domenkozar: I'm also not a lawyer, but I'm getting lawyers involved.

domenkozar:
We're going to remove `devenv generate` command and expose ways like `devenv mcp` so you can work with existing AI tooling of your choice. For generating devenv.nix files for now you can use https://devenv.new

PR #1791, opened 2025-03-24
Add Typst fontPaths option

victorbnl:
In order to make some fonts available to Typst, the latter supports the `TYPST_FONT_PATHS` environment variable. Sure, fonts can be made available by simply adding them to the `packages`. However, doing so results in Typst seeing them as system fonts, so you can't really know if it mistakenly chooses a font that is installed on your computer instead. Setting them via `TYPST_FONT_PATHS`, in combination with running Typst with the `--ignore-system-fonts` flag, allows for a fully reproducible build.

### DIFF
--- a/src/modules/languages/typst.nix
+++ b/src/modules/languages/typst.nix
@@ -19,7 +19,7 @@ in
       description = "Directories to be searched for fonts.";
       default = [ ];
       defaultText = lib.literalExpression "[]";
-      example = lib.literalExpression ''[ "\${pkgs.roboto}/share/fonts/truetype "] ]'';
+      example = lib.literalExpression ''[ "''${pkgs.roboto}/share/fonts/truetype "] ]'';
     };
   };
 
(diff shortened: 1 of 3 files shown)


src/modules/languages/typst.nix
```diff
@@ -13,6 +13,14 @@ in
       default = pkgs.typst;
       defaultText = lib.literalExpression "pkgs.typst";
     };
+
+    fontPaths = lib.mkOption {
+      type = lib.types.listOf lib.types.str;
+      description = "Directories to be searched for fonts.";
+      default = [ ];
+      defaultText = lib.literalExpression "[]";
+      example = [ "\${pkgs.roboto}/share/fonts/truetype" ];
```
domenkozar, line 22:
```suggestion
      example = lib.literalExpression [ "\${pkgs.roboto}/share/fonts/truetype" ];
```

victorbnl: I don't understand what lib.literalExpression does so I'll trust you on this.

victorbnl:
Now it puts the actual package path into the example, which is unexpected

```md
*Example:*
` [ "\/nix/store/pbcb7j89fggfmjafzplcpylvksay4v9q-roboto-2.138/share/fonts/truetype "] ] `
```

domenkozar:
Oh, the dollar sign should be escaped with `''`

PR #1794, opened 2025-03-25
devenv: make overlay changes backwards compatible with older modules

sandydoo:
Since overlays require CLI-level changes to module eval, all of the changes should be shipped with the CLI. Othewise, newer CLI versions will not work with older modules.

Continuation of #1768.

### DIFF
--- a/devenv/src/flake.tmpl.nix
+++ b/devenv/src/flake.tmpl.nix
@@ -60,8 +60,11 @@
               then devenvdefaultpath
               else throw (devenvdefaultpath + " file does not exist for input ${name}.");
           project = pkgs.lib.evalModules {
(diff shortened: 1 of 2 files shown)


devenv/src/flake.tmpl.nix
```diff
@@ -63,6 +63,9 @@
             specialArgs = inputs // { inherit inputs; bootstrapPkgs = pkgs; };
```
domenkozar, line 63: Not sure we still need to pass `bootstrapPkgs` here

devenv/src/flake.tmpl.nix
```diff
@@ -63,6 +63,9 @@
...
             modules = [
               (inputs.devenv.modules + /top-level.nix)
```
domenkozar, line 65: We should put the pkgs override first, just in case.

PR #1809, opened 2025-04-04
postgres: improve how PGHOST is computed

sandydoo:
- Use one of the IP addresses from `listen_addresses` if configured.
- Handles special listen addresses, like `*`, `0.0.0.0`, and `::`.
- Only set `PGPORT` if using IP addresses.
- Expand the docs for `listen_addresses` using the official postgres docs as reference.

### DIFF
--- a/docs/reference/options.md
+++ b/docs/reference/options.md
@@ -84429,7 +84429,19 @@ CREATE ROLE bar;
 
 
 
-Listen address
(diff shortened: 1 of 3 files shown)


schonfinkel:
@sandydoo thank you for these changes, I had once resorted to overriding the `PGHOST` set by devenv, to follow whatever value I had put in my listen_address.

Issue #1820, opened 2025-04-14
When using uv, better be sure it's not pulling python from the internet


leiserfg:
If for a mistake one sets a version of python in `devenv.nix`, that does not mach the one in `pyproject.toml`, uv will fetch a python from internet which can cause issues with nix. I  think it's better to set `UV_PYTHON_DOWNLOADS=never` and see uv crash than having unexpected behaviors.

domenkozar:
Yeah! Can you make a PR?

PR #1822, opened 2025-04-14
Force uv not to download

leiserfg:
Forces uv not to download binaries in case of mismatch
Implements #1820

I also considered adding an assert for checking if the version of the python is in the set specified by `project.requires-python` but I'm not sure if that makes sense in devenv context.


sandydoo:
Sounds good to me. @leiserfg, could you please remove the css changes?

leiserfg:
Sorry, no Idea how that happened, I'm far from home will do that tomorrow morning.

sandydoo:
Thanks!

PR #1835, opened 2025-04-18
direnv: fix excessive reloads

### LINKED ISSUE #1824: Devenv + devcontainer + direnv-vscode = infinite process spawn
**Describe the bug**
When using devenv inside a devcontainer with the direnv-vscode extention enabled, some devenv command (like info or up, likely any command that use nix in some way) make the direnv-vscode extension go nut and spawn and infinite nix processes, which cause the host to crash due to a lack of ram.
This bug is only present when I do a devenv command (the devcontainer boot fine and the  direnv-vscode work normally if no devenv command are done).
direnv without the vscode extension works also fine

**To reproduce**
https://gist.github.com/cfouche3005/8c9b5c3fe962b031c3a2e7898ab3743f
make devcontainer with devenv
install the direnv-vscode extension and let it do the scan and restart
run devenv info

**Version**
 devenv 1.5.0 (x86_64-linux)

cannot be reproduced in 1.4.1 with the same setup (extension and other)
affect multiple different machine setup (work vm and private laptop pc)

sandydoo:
When watching generated files, we need to take care not to touch the files unless they change. direnv only compares the modification time, not the file contents, so it's triggered every time we write these files. Even worse, the vscode extension is happy to re-evaluate the shell without waiting for the previous one to finish loading.

Fixes #1824.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -652,6 +652,7 @@ dependencies = [
  "devenv-tasks",
  "dialoguer",
  "dotlock",
+ "fd-lock",
  "hex",
  "http-client-tls",
  "include_dir",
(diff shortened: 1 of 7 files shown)


TrevorWinstral:
I just updated to 1.6.1 and this issue came back, I had no issues on 1.6.0 but was forced to update for the fixes.

My issue is that when I update devenv.nix while I have VSCode open (with the direnv extension) it spawns a ton of processes and maxes out my CPU, RAM, and swap. If VSCode is not open, and just modify devenv.nix in a terminal, there is no issue at all. In the screenshot below my swap hadn't been maxed out yet, so I took the screenshot before my computer crashed.
Screenshot From 2024-12-13 14-57-50

PR #1837, opened 2025-04-18
devenv: propagate the exit code when executing a shell command

### LINKED ISSUE #1836: devenv shell swallows exit codes
**Describe the bug**
When using devenv shell to execute commands non-interactively, devenv shell does not correctly pass on the exit code. 

One such simple use-case are Github CI/CD Pipelines as documented in https://devenv.sh/integrations/github-actions/#run-multiple-commands

If I can believe my git bisect run, this issue started with 1efab3e6e3b75d3d319f812d5a970438a3c4d930 / https://github.com/cachix/devenv/pull/1789

**To reproduce**

### Simple test script for bisect:

```
#!/usr/bin/env bash

cargo build
cargo sqlx database reset -y

cat <<EOF > test_script.sh
ls /doesnt_exist
EOF

chmod a+rx ./test_script.sh

target/debug/devenv shell bash -- -e ./test_script.sh
SHELL_EXIT_CODE=$?

git reset --hard
if [[ $SHELL_EXIT_CODE -eq 0 ]]; then
    exit 1
else
    exit 0
fi;
```

### Bisect run:

```
$ git bisect start
$ git bisect good v1.3.1
$ git bisect bad main
$ git bisect run ./run-fail-test.sh
```

The problem seems to be in https://github.com/cachix/devenv/blob/main/devenv/src/main.rs#L95
This is because this receives an Ok(Output) at this level, so the unwrap does not automatically propagate execution failures.

One (quckfix) way that makes this work for me again is like so:

```
---
 devenv/src/main.rs | 17 +++++++++++++----
 1 file changed, 13 insertions(+), 4 deletions(-)

diff --git a/devenv/src/main.rs b/devenv/src/main.rs
index c4ae13d9..ca07e9c2 100644
--- a/devenv/src/main.rs
+++ b/devenv/src/main.rs
@@ -91,10 +91,19 @@ async fn main() -> Result<()> {

     match command {
         Commands::Shell { cmd, ref args } => match cmd {
-            Some(cmd) => {
-                devenv.exec_in_shell(cmd, args).await?;
-                Ok(())
-            }
+            Some(cmd) => match devenv.exec_in_shell(cmd, args).await {
+                Ok(shell_res) => match shell_res.status.code() {
+                    Some(0) => Ok(()),
+                    Some(exit_code) => {
+                        miette::bail!("Shell execution failed: exit code {exit_code}")
+                    }
+                    None => miette::bail!("Failed retrieving exit code from shell execution"),
+                },
+                Err(err) => {
+                    println!("Failed executing command: {err}");
+                    miette::bail!("Shell execution failed: exit code {err}")
+                }
+            },
             None => devenv.shell().await,
         },
         Commands::Test { .. } => devenv.test().await,
--
2.47.2
```

**Version**

`>=1.5.0` including current main, not using flakes, but likely is a problem in both scenarios.

sandydoo:
Fixes #1836.


d-goldin:
Thanks for the swift reaction!

Briefly tried it out and can confirm that this now correctly passes on the error code.
Because this is imho a very important and critical piece, I think we should add some type of test for this. Since I'm quite new to this code-base I'm not sure about the usual testing strategy, but maybe we could add some step to the builtest.yml pipeline to "cement" this interface?

PR #1838, opened 2025-04-18
Add --option CLI flag for overriding configuration options

domenkozar:
It's common to be able to build matrix on CI jobs, so let's allow passing options via CLI:

```
devenv --option services.redis.enable:bool true test
```

cc @staticdev

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -12,7 +12,7 @@
 ## Code Style Guidelines
 - **Imports**: Group by category (std lib first, then external crates, then internal)
 - **Naming**: Use `snake_case` for functions/variables, `CamelCase` for types/traits
-- **Error Handling**: Use `thiserror` crate with custom error types and `?` operator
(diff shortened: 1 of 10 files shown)


vetsin:
Given the task model, why not natively support the concept of matrix/parallel *within* devenv? E.g. in the interest of reproducibility I would want to run a 'matrix task' via devenv not require some sort of shell wrapper to reproduce locally

domenkozar:
> Given the task model, why not natively support the concept of matrix/parallel _within_ devenv? E.g. in the interest of reproducibility I would want to run a 'matrix task' via devenv not require some sort of shell wrapper to reproduce locally

I'd like to support matrix within devenv and has always been the plan, but this is more of a quick way to be able to write scripts using devenv and for testing quick changes.

domenkozar:
I've changed the design to always require specifying the type:

`devenv --option env.FOO:bool true`

PR #1841, opened 2025-04-22
docs: document how to use profiles

### LINKED ISSUE #1426: [ENHANCEMENT] Multiple devenv(-up) profiles/variants
Concrete use case:
I have a devenv.nix in the root of our repo to run everything together. Now depending on different development focus, we would love to either pass variables for devenv to evaluate (think `--no-frontend`) or to have multiple devenv profiles where we could e.g. say `devenv up --backend-only`. Similar to a flake's `nix develop '.#profile'.

I know that this should be possible with flakes, but since the rest of the team is not "nix-native" and we really enjoy the abstraction devenv brings, this would be a killer feature.

What would be necessary to build this up?

domenkozar:
Sometimes it's useful to be able to only run a subset of developer environment and choose.

[guide preview](https://docs-profiles.devenv.pages.dev/guides/profiles/)

Fixes #1426

### DIFF
--- a/docs/blog/posts/devenv-v1.6-ad-hoc-developer-environments.md
+++ b/docs/blog/posts/devenv-v1.6-ad-hoc-developer-environments.md
@@ -90,6 +90,16 @@ This feature is particularly useful for:
 
 When used with an existing `devenv.nix` file, `--option` values will override the configuration in the file, making it easy to temporarily modify your environment.
 
+## Switching Between Environment Profiles
+
+Ad-hoc options are perfect for switching between predefined profiles in your development environment:
+
+```shell-session
+$ devenv --option profile:string backend up
+```
+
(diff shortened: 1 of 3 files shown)


euphemism:
Thank you. This is helpful and illuminating.

The linked issue mentioned the flake syntax for selecting an output. That seems more ergonomic than `devenv —option profile:string backend up`. i.e.  `devenv up .#backend`, but I assume that means extra work on top of just utilizing the new option functionality to achieve the goal.

domenkozar:
> Thank you. This is helpful and illuminating.
> 
> The linked issue mentioned the flake syntax for selecting an output. That seems more ergonomic than `devenv —option profile:string backend up`. i.e. `devenv up .#backend`, but I assume that means extra work on top of just utilizing the new option functionality to achieve the goal.

You can select profile in `devenv.loca.nix` using `{ profile = "full"; }`, is that what you need? It helps to know what kind of workflow you'd like to follow to come up with an example :)

PR #1853, opened 2025-04-29
tasks: implement execIfModified

### LINKED ISSUE #1616: tasks: status_if_modified
https://github.com/cachix/devenv/commit/a520f05c40ebecaf5e17064b27e28ba8e70c49fb

### LINKED ISSUE #1682: Tasks should accept rerunIfModified only run if a file was modified since task execution
There's a common pattern with language-specific package managers only needing to re-run if I changed something like `Cargo.toml`, or if I checked out another revision and my `Cargo.lock` was updated. It's already possible to implement this with the `status` command, but I feel like this should come built-in with devenv.

```nix
tasks.cargo = {
  exec = "cargo fetch";
  rerunIfModified = [ "Cargo.toml" "Cargo.lock" ];
};
```

domenkozar:
Fixes #1682 
Fixes #1616 

For example:

```nix
  tasks = {
    "devenv:compile-requirements" = {
      before = [ "devenv:python:virtualenv" ];
      exec = "uv pip compile requirements.in -o requirements.txt";
      execIfModified = [
        "requirements.in"
        "requirements.txt"
      ];
    };
  };
```

cc @rawkode


devenv-tasks/src/task_cache.rs
```diff
@@ -0,0 +1,436 @@
...
+            })?;
+        }
+
+        // Create database if it doesn't exist
+        if !Sqlite::database_exists(&db_url).await.unwrap_or(false) {
+            debug!("Creating task cache database at {}", db_path);
+            Sqlite::create_database(&db_url).await.map_err(|e| {
+                TaskCacheError::Initialization(format!("Failed to create SQLite database: {}", e))
+            })?;
+        }
```
sandydoo, line 86: The `create_if_missing(true)` option below does the same thing

devenv-tasks/src/task_cache.rs
```diff
@@ -0,0 +1,436 @@
...
+            Ok(0) => break,
+            Ok(n) => {
+                hasher.update(&buffer[..n]);
+            }
+            Err(e) => return Err(e.into()),
+        }
+    }
+
+    let hash = hasher.finalize();
+    Ok(hash.to_hex().to_string())
```
sandydoo, line 357:
```suggestion
    let mut file = std::fs::File::open(path)?;
    let mut hasher = Hasher::new();
    std::io::copy(&mut file, &mut hasher)?;
    Ok(hasher.finalize().to_hex().to_string())
```

sandydoo (commented):
This is a subset of the eval cache implementation with a few task-specific naming differences. I'm wondering whether it would make sense to extract the database and file checking parts of eval-cache and build both eval-cache and task-cache on top of a shared base.

domenkozar:
> This is a subset of the eval cache implementation with a few task-specific naming differences. I'm wondering whether it would make sense to extract the database and file checking parts of eval-cache and build both eval-cache and task-cache on top of a shared base.

I was thinking the same, ideally we'd be using the same database for all the caching.

domenkozar:
@sandydoo I've created `devenv-cache-core` that shared the sqlite logic and file operations and reused it in both crates :)

PR #1882, opened 2025-05-11
git-hooks: deprecate pre-commit in favour of git-hooks

sandydoo:
This has been aliased for a while now. We'll now show a deprecation notice.

While investigating the performance of the reference page, I realized that `mkAliasOptionModule` actually keeps the aliased option in the docs. 
This explains why the rendering performance of that page drastically worsened in the recent months.


sandydoo:
-36k 🤣

Issue #36, opened 2022-11-13
zsh support


kubukoz:
Especially when used with direnv - it'd be nice to keep a similar UX to what plain Nix with flakes does.

devenv:

<img width="392" alt="image" src="https://user-images.githubusercontent.com/894884/201531361-50a0316a-f959-4170-a3f7-976383b89557.png">

direnv with flakes:

<img width="484" alt="image" src="https://user-images.githubusercontent.com/894884/201531376-d6a707df-4ac8-4884-9696-d044bc7251b6.png">

domenkozar:
Here's where we set `$PS1`: https://github.com/cachix/devenv/blob/main/src/modules/top-level.nix#L81

How should it be adapted for zsh?

kubukoz:
hmm, I think I got confused here. I thought since the prompt says `bash`, we're actually in bash - but this would suggest otherwise:

<img width="575" alt="image" src="https://user-images.githubusercontent.com/894884/201536558-8f225e9d-0c79-4406-80c6-7fefd202ec57.png">

shyim:
I guess with devenv your actual shell is bash and not your configured one. I guess the `SHELL` env gets not overwritten and is from the parent shell

edio:
@domenkozar , I think the author wanted devenv to launch zsh rather than bash. It's not about `PS1`.

We can't really compare to direnv, it's a different beast, a more fair would be a comparison to nix-shell. It too launches `bash` no matter what.

There's a 7-year old issue about supporting shells other than bash in nix-shell. It's still open :see_no_evil: 
https://github.com/NixOS/nix/pull/545

@kubukoz , one way to check your shell from within is
```
echo $0
```

manveru:
Yeah, this is the main thing keeping me from using direnv+devenv over direnv+mkShell... forcing people into using bash isn't an option for me.

domenkozar:
@manveru I'd really like to fix this but I'm not sure how to reproduce the issue.

manveru:
I have this in `.envrc`:

```
watch_file devenv.nix
watch_file devenv.yaml
watch_file devenv.lock
if [[ ! "$DIRENV_ACTIVE" =~ (^|:)"$PWD"(:|$) ]]; then
  devenv shell
fi
```

and it starts a new bash instead of exporting the environment to zsh.

domenkozar:
And this otherwise works with `nix develop`? If I understand correctly, nix always launches bash.

manveru:
[...]

PR #1885, opened 2025-05-14
Add DEVENV_CMDLINE envvar

mcdonc:
FWIW, I've now tested each devenv command using my exec-zsh-within-enterShell thing and at https://github.com/devenv-zsh and I think the only problem I cannot work around without modifying devenv source is this one:

- At the moment "devenv up" / "devenv processes up" must be run from within an existing devenv shell, because if you invoke it from without, enterShell will exec zsh before the process manager script runs.

This is because there is no reliable way to detect whether enterShell is being invoked by a call to "devenv shell" or from a call to "devenv processes up".  This PR sets an environment variable that distinguishes these two situations from each other.

See also https://github.com/cachix/devenv/issues/36


mcdonc:
Ugh, not sure why all those stray commits came along for the ride.  I'm hoping you can squash them out if you merge it, because the diff is correct.

PR #1886, opened 2025-05-14
Add services.postgres.initialDatabases.initialSQL config option

### LINKED ISSUE #1673: initialScript isn't run against initialDatabases
**Describe the bug**
The [documentation for Services](https://devenv.sh/services/) has an example of running timescaledb that uses `initialScript` to create the extension. However, `initialScript` is [only run against the default database](https://github.com/cachix/devenv/blob/1c384bc4be3ee571511fbbc6fdc94fe47d60f6cf/src/modules/services/postgres.nix#L97) named `postgres`, meaning that the timescaledb extensions are not available to the other databases ("mydb" in the Services documentation example). 

**To reproduce**
https://gist.github.com/CalebJohn/6bd1cfc8c4daff6a849ddf7e17d1ecf8

Run `devenv up`.

It can be seen that the timescaledb extension is not actually loaded for the mydb database (`initialDatabases[0]`), but it is available under the databased named postgres.
```
$ psql --dbname mydb --command="SELECT installed_version FROM pg_available_extensions WHERE name = 'timescaledb';"
 installed_version
-------------------

(1 row)


$ psql --dbname postgres --command="SELECT installed_version FROM pg_available_extensions WHERE name = 'timescaledb';"
 installed_version
-------------------
 2.17.2
(1 row)
```

**Version**
 
`devenv 1.3.1 (x86_64-linux)`

I think this is a matter of the documentation being lacking. There is nothing in the documentation to suggest which database `initialScript` is being applied against. After digging into the source I think my expected behaviour was wrong, but I wanted to flag up my troubles here as I think I'm not the only one who has run into the same misunderstanding.

thesoulless:
This PR fixes #1673. If you agree with the approach to resolve this issue, please let me know so I can update the documentation as well.

### DIFF
--- a/src/modules/services/postgres.nix
+++ b/src/modules/services/postgres.nix
@@ -67,10 +67,10 @@ let
             if [ 1 -ne "$dbAlreadyExists" ]; then
               echo "Creating database: ${database.name}"
               echo 'CREATE DATABASE "${database.name}";' | psql --dbname postgres
-              if [ ${q database.initialScript} != null ]
+              if [ ${q database.initialSQL} != null ]
               then
                 echo "Running initial script on database ${database.name}"
-                echo ${q database.initialScript} | psql --dbname ${database.name}
+                echo ${q database.initialSQL} | psql --dbname ${database.name}
               fi
               ${lib.optionalString (database.user != null && database.pass != null) ''
               echo "Creating role ${database.user}..."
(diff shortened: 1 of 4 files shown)


src/modules/services/postgres.nix
```diff
@@ -332,6 +337,14 @@ in
               Password of owner of the database (only takes effect if `user` is not `null`).
             '';
           };
+          initialScript = lib.mkOption {
```
domenkozar, line 340: Let's this time get it right and name it `initialSQL`

domenkozar:
Yeah, let's also update docs to clarify which one to use and why :)

thesoulless:
@domenkozar Changed the option name to `initialSQL`, and updated the docs.
Let me know if I've missed something.

PR #1890, opened 2025-05-17
Avoid setting TYPST_FONT_PATHS if not explicitly set

leiserfg:


leiserfg:
Without this, some typst commands fail if font's are not set.

src/modules/languages/typst.nix
```diff
@@ -30,6 +35,6 @@ in
       pkgs.typstyle # formatter
     ];
 
-    env.TYPST_FONT_PATHS = lib.concatStringsSep ":" cfg.fontPaths;
+    env.TYPST_FONT_PATHS = if cfg.fontPaths then (lib.concatStringsSep ":" cfg.fontPaths) else null;
```
domenkozar, line 38: I think this needs to check for empty list

leiserfg: You are right, forgot that nix does not autocast to bool.

PR #1892, opened 2025-05-19
Allow configuration of the Nixpkgs instance on a per-system basis

ndam-hexagon:
# Description

This PR aims to allow setting some of the attribute of the `config` attribute set that is normally passed as argument when importing Nixpkgs.

An example that is currently not supported (but would be after this PR):
```nix
{
  pkgs = import nixpkgs {
    inherit system;
    overlays = [ ... ];
    config = {
      allowUnfree = true;
      allowBroken = true;
      cudaSupport = true;
      cudaCapabilities = [
        "7.5"
        "8.6"
        "8.9"
      ];
    };
  };
}
```

## Context

We have a project where we work with CUDA and override various packages (e.g. OpenCV) to have CUDA support enabled. Currently, we are using `devenv` with the flake integration, but would like to now migrate to using the `devenv` command and `devenv.yaml`/`devenv.nix` to benefit from faster load times.

We also have the added complexity of supporting 2 architectures: x86 and aarch64 with incompatible sets of CUDA capabilities. We therefore needed to patch `devenv` in order to allow us to configure those settings on a per-system basis.

## Changes in a nutshell

This PR modifies the config structure of the `devenv.yaml` file in the following ways:

- Add a top-level element to enable/disable CUDA support for `nixpkgs` globally: `cudaSupport`
- Add a new top-level element: `config` that can be used to configure on a per-system basis how `devenv` later configures its Nixpkgs instance

Here's an example of this new format for the config file:
```yaml
# ... usual config

config:
  x86_64-linux:
    allowUnfree: true
    allowBroken: true
    cudaSupport: true
    cudaCapabilities:
      - "7.5"
      - "8.6"
      - "8.9"
  aarch64-linux:
    allowUnfree: true
    allowBroken: true
    cudaSupport: true
    cudaCapabilities:
      - "8.7"
```

The `devenv/src/flake.tmpl.nix` was modified in such a way that the `config.<system>.<keys>` take precedence over `<keys>` if present, but should not otherwise have an impact on existing configurations.

### DIFF
--- a/devenv/src/config.rs
+++ b/devenv/src/config.rs
@@ -108,6 +108,20 @@ pub struct Clean {
 #[derive(schematic::Config, Clone, Serialize, Debug, JsonSchema)]
 #[config(rename_all = "camelCase", allow_unknown_fields)]
 #[serde(rename_all = "camelCase")]
+pub struct Nixpkgs {
+    #[serde(flatten)]
+    pub config_: NixpkgsConfig,
+    #[serde(
+        rename = "per-platform",
+        skip_serializing_if = "BTreeMap::is_empty",
+        default
+    )]
+    pub per_platform: BTreeMap<String, NixpkgsConfig>,
+}
+
+#[derive(schematic::Config, Clone, Serialize, Debug, JsonSchema)]
+#[config(rename_all = "camelCase", allow_unknown_fields)]
+#[serde(rename_all = "camelCase")]
 pub struct Config {
     #[serde(skip_serializing_if = "BTreeMap::is_empty", default)]
     #[setting(nested)]
@@ -120,9 +134,9 @@ pub struct Config {
     pub cuda_support: bool,
     #[serde(skip_serializing_if = "Vec::is_empty", default)]
     pub cuda_capabilities: Vec<String>,
-    #[serde(skip_serializing_if = "BTreeMap::is_empty", default)]
+    #[serde(skip_serializing_if = "Option::is_none", default)]
     #[setting(nested)]
-    pub config: BTreeMap<String, NixpkgsConfig>,
+    pub nixpkgs: Option<Nixpkgs>,
     #[serde(skip_serializing_if = "Vec::is_empty", default)]
     pub imports: Vec<String>,
     #[serde(skip_serializing_if = "Vec::is_empty", default)]
--- a/devenv/src/config.rs
+++ b/devenv/src/config.rs
@@ -130,10 +130,6 @@ pub struct Config {
     pub allow_unfree: bool,
     #[serde(skip_serializing_if = "is_false", default = "false_default")]
     pub allow_broken: bool,
-    #[serde(skip_serializing_if = "is_false", default = "false_default")]
-    pub cuda_support: bool,
-    #[serde(skip_serializing_if = "Vec::is_empty", default)]
-    pub cuda_capabilities: Vec<String>,
     #[serde(skip_serializing_if = "Option::is_none", default)]
     #[setting(nested)]
     pub nixpkgs: Option<Nixpkgs>,
--- a/devenv/src/config.rs
+++ b/devenv/src/config.rs
@@ -100,6 +100,8 @@ pub struct Config {
     pub allow_unfree: bool,
     #[serde(skip_serializing_if = "is_false", default = "false_default")]
     pub allow_broken: bool,
+    #[serde(skip_serializing_if = "is_false", default = "false_default")]
+    pub cuda_support: bool,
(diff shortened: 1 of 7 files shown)


devenv/src/config.rs
```diff
@@ -6,6 +6,20 @@ use std::{collections::BTreeMap, fmt, path::Path};
 
 const YAML_CONFIG: &str = "devenv.yaml";
 
+#[derive(schematic::Config, Clone, Debug, PartialEq, Eq, Serialize, Deserialize, JsonSchema)]
```
ndam-hexagon, line 9: I'm new to rust, so not sure that this is correct. Let me know if I should change this pragma looking line

domenkozar:
Hey! I'd change a bit the way the config is structured:

```yaml
nixpkgs:
  allowUnfree: true
  allowBroken: false
  
  per-platform:
    x86_64-linux:
      allowUnfree: true
      allowBroken: true
    aarch64-darwin:
      allowUnfree: false
```

ndam-hexagon:
> Hey! I'd change a bit the way the config is structured:
> 
> ```yaml
> nixpkgs:
>   allowUnfree: true
>   allowBroken: false
>   
>   per-platform:
>     x86_64-linux:
>       allowUnfree: true
>       allowBroken: true
>     aarch64-darwin:
>       allowUnfree: false
> ```

Sure. I'll make those changes and push a new commit. Are you otherwise ok with how I structured the changes?

domenkozar:
> > Hey! I'd change a bit the way the config is structured:
> > ```yaml
> > nixpkgs:
> >   allowUnfree: true
> >   allowBroken: false
> >   
> >   per-platform:
> >     x86_64-linux:
> >       allowUnfree: true
> >       allowBroken: true
> >     aarch64-darwin:
> >       allowUnfree: false
> > ```
> 
> Sure. I'll make those changes and push a new commit. Are you otherwise ok with how I structured the changes?

Yeah, really good stuff :)

ndam-hexagon:
Given the changes you are proposing, should the old syntax still be accepted?

```yaml
allowUnfree: true
```

Or should the only way to configure be:

```yaml
nixpkgs:
    allowUnfree: true
```

domenkozar:
Yeah, we should still support the old way. And change docs for the `nixpkgs` :)

docs/reference/yaml-options.md
```diff
@@ -1,19 +1,32 @@
...
+| nixpkgs.cudaSupport                                   | Enable CUDA support for nixpkgs. Defaults to `false`.                         |
+| nixpkgs.permittedInsecurePackages                     | A list of insecure permitted packages. Defaults to `[]`                       |
+|                                                       |                                                                               |
+| nixpkgs.&lt;per-platfom&gt;.allowBroken               | (per-platform) Allow packages marked as broken. Defaults to `false`.          |
+| nixpkgs.&lt;per-platfom&gt;.allowUnfree               | (per-platform) Allow unfree packages. Defaults to `false`.                    |
+| nixpkgs.&lt;per-platfom&gt;.cudaCapabilities          | (per-platform) Select CUDA capabilities for nixpkgs. Defaults to `[]`         |
+| nixpkgs.&lt;per-platfom&gt;.cudaSupport               | (per-platform) Enable CUDA support for nixpkgs. Defaults to `false`.          |
+| nixpkgs.&lt;per-platfom&gt;.permittedInsecurePackages | (per-platform) Select CUDA capabilities for nixpkgs. Defaults to `[]`         |
 
 !!! note "Added in 1.0"
```
domenkozar, line 34: Here we need to add above 1.0, that `nixpkgs` was added in 1.7

ndam-hexagon: Was not sure where you wanted that. Will add it there then.

ndam-hexagon: If you want it below instead of above, let me know ;-)

ndam-hexagon:
Now that think about it, the followings would be introduced with this PR at the top-level:
- `cudaSupport`
- `cudaCapabilities`

Do you want those to only be allowed within `nixpkgs`?

domenkozar:
Yes, thank you!

domenkozar:
Note that jsonschema will get automatically generated, see https://github.com/cachix/devenv/actions/runs/15467562476/job/43544279281#step:15:23

ndam-hexagon:
> Note that jsonschema will get automatically generated, see https://github.com/cachix/devenv/actions/runs/15467562476/job/43544279281#step:15:23

Ah neat. I will remove those changes then.

PR #1908, opened 2025-06-03
outputs: fixes `devenv build` when building all outputs

### LINKED ISSUE #1902: `devenv build` causes crashes: error: attribute 'name' missing
**Describe the bug**
`devenv build` causes crashes

```
[...]
       error: at <snip>/.devenv.flake.nix:139:22:
          138|                 if builtins.hasAttr "type" option then
          139|                   if option.type.name == "output" || option.type.name == "outputOf" then {
             |                      ^
          140|                     ${name} = config.${name};

       error: attribute 'name' missing
```

The code in question:

```nix
           build = options: config:
              lib.concatMapAttrs
                (name: option:
                  if builtins.hasAttr "type" option then
                    if option.type.name == "output" || option.type.name == "outputOf" then {
                      ${name} = config.${name};
                    } else { }
                  else
                    let v = build option config.${name};
                    in if v != { } then {
                      ${name} = v;
                    } else { }
                )
                options;
```

Not sure how to debug it as the module isn't exposed (or I don't know where it is) in `direnv repl`. Looks like some options might not have `name` attribute, but I can't tell which one. I was trying to use trace but devenv overrides it.

**Version**
 
devenv 1.6.1 (aarch64-darwin)

sandydoo:
Improve how we inspect options and their types by using the module helper functions.

This fixes #1902, where `services.keycloak.database` was being incorrectly classified as an option because it has a nested option named `type`.

### DIFF
--- a/src/modules/tasks.nix
+++ b/src/modules/tasks.nix
@@ -123,7 +123,7 @@ in
       description = "The generated tasks.json file.";
     };
     task.package = lib.mkOption {
-      type = config.lib.types.output;
+      type = types.package;
       internal = true;
       default = lib.getBin devenv-tasks;
     };
--- a/devenv/src/flake.tmpl.nix
+++ b/devenv/src/flake.tmpl.nix
@@ -112,11 +112,13 @@
               );
           };
 
+          # Recursively search for outputs in the config.
+          # This is used when not building a specific output by attrpath.
           build = options: config:
             lib.concatMapAttrs
               (name: option:
-                if builtins.hasAttr "type" option then
-                  if option.type.name == "output" || option.type.name == "outputOf" then {
+                if lib.isOption option then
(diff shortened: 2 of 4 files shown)


src/modules/tasks.nix
```diff
@@ -123,7 +123,7 @@ in
       description = "The generated tasks.json file.";
     };
     task.package = lib.mkOption {
-      type = config.lib.types.output;
+      type = types.package;
```
sandydoo, line 126: @domenkozar is this a typo? this gets picked up by `devenv build`, which is kinda weird for end-users to see.
I guess a related question is whether we should be recursively looking for outputs in the first place?

domenkozar: I was thinking at the time, that `devenv build` should build all packages used inside the environment. Maybe that's where the line should be drawn, most users would expect it to build only their own.

PR #1922, opened 2025-06-10
zig: override option zls

Daholli:
Included option to override zls e.g using a flake input


Daholli:
Thanks for merging <3

PR #1934, opened 2025-06-15
Add opentofu example

jreslock:
I noticed that there is not an example for opentofu. I just grabbed/copied the terraform example and modified it to create one for opentofu. Looking at the nixos packages repository I _think_ that this is correct. This is my first PR to this repository and I  may have missed something. Happy to fix or add anything that was missed in the initial commit.

### DIFF
--- a/examples/opentofu/.test.sh
+++ b/examples/opentofu/.test.sh
@@ -1,0 +1,3 @@
+#!/usr/bin/env bash
+set -ex
+tofu --version | grep "1.9.1"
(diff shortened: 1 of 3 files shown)


jreslock:
Tests fail locally so I am setting this to a draft while I figure out where to pull the actual binary from.

domenkozar:
Can you move this to tests? We're limited by github actions 250 jobs sadly so we can't have more examples :(

domenkozar:
:(

PR #1940, opened 2025-06-18
fix up args to join with space

tzemanovic:
hi, there's a regression in 1.6 that joins the `devenv up` args without a space so it doesn't work properly with multiple args


domenkozar:
Thanks :heart:

PR #1947, opened 2025-06-23
feat: add enable/disable toggles for realm export

gabyx:
- Enable and disable realm export with
  ```nix
  scripts.exportRealm = false;
  processes.exportRealms = false;
  ```
  
- Only make the process `keycloak-realm-export-all` show up if at least some `realm.XXX.export = true` are set.

### DIFF
--- a/docs/reference/options.md
+++ b/docs/reference/options.md
@@ -19442,6 +19442,28 @@ list of absolute path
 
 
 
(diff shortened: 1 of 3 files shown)


tboerger (approved):
Haven't tested it but LGTM

PR #1952, opened 2025-06-25
process-compose: inherit the environment from the shell

### LINKED ISSUE #1951: devenv up sets RUSTFLAGS= even if not specified in the config
**Describe the bug**

When running devenv up, it always starts process-compose with RUSTFLAGS set as empty even when on shell it is not set.
The issue of that is that cargo won't read .cargo/config.toml anymore and builds break.

**To reproduce**
Clone devenv itself and apply this patch

```diff
diff --git a/devenv.nix b/devenv.nix
index de392900..408f46dd 100644
--- a/devenv.nix
+++ b/devenv.nix
@@ -248,6 +248,9 @@
     };
   };
 
+
+  processes.test.exec = "env|grep RUST";
+
   git-hooks.hooks = {
     nixpkgs-fmt.enable = true;
     rustfmt.enable = true;

```
Run devenv up and you will see the variable.

**Version**

devenv 1.6.1 (aarch64-darwin)

sandydoo:
Fixes #1951.

We can inherit all the env vars from the shell and avoid more complex filtering logic.
`config.env` isn't an exhaustive attrset of environment variables anyways, so it's not like the resulting process-compose config is a standalone file.

### DIFF
--- a/src/modules/process-managers/process-compose.nix
+++ b/src/modules/process-managers/process-compose.nix
@@ -120,12 +120,6 @@ in
         version = lib.mkDefault "0.5";
         is_strict = lib.mkDefault true;
         log_location = lib.mkDefault "${config.env.DEVENV_STATE}/process-compose/process-compose.log";
-        # Filter out the recursive PC_CONFIG_FILES env.
-        # Otherwise, we would get a loop:
-        #   PC_CONFIG_FILES -> configFile -> settings -> PC_CONFIG_FILES -> ...
-        environment = lib.mapAttrsToList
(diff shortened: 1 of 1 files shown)


sandydoo:
> Question: How is this recursive loop now being prevented? The PR description doesn't address this specific issue that the original code was designed to solve.

Not an issue with the custom environment removed.

> Processes will now inherit ALL environment variables from the shell, including potentially sensitive ones (AWS keys, tokens, etc.)

`environment` is additive AFAIKT.

domenkozar:
fc9b40d8f

PR #1959, opened 2025-06-26
fix: add `UV_PYTHON` to force uv to use the correct python

gabyx:
In this module case:

```
{
          languages.python = {
            enable = true;
            package = pkgsPinned.python;
            uv.enable = true;
          };
}
```
when I run `uv venv ...` it will take up my python version from a virtual environment (activated on shell entry by `.zshrc` etc , of course this I should avoid). But here is a simple fix which IMO was missed. 

- `uv` must always use the correct python version by forcing it to use it with `UV_PYTHON`.
- No fix for poetry so far, as poetry does not nicely play with this and is also deprecated in terms of modernity.

### DIFF
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -445,6 +445,8 @@ in
       UV_PROJECT_ENVIRONMENT = "${config.env.DEVENV_STATE}/venv";
       # Force uv not to download a Python binary when the version in pyproject.toml does not match the one installed by devenv
       UV_PYTHON_DOWNLOADS = "never";
+      # Force uv to always use the correct python interpreter.
+      UV_PYTHON = "${cfg.package.interpreter}";
     }) // (lib.optionalAttrs cfg.poetry.enable {
       # Make poetry use DEVENV_ROOT/.venv
       POETRY_VIRTUALENVS_IN_PROJECT = "true";


domenkozar:
@claude can you highlight the lines that changes, ignoring formatting changes and comment on the changes.

d-goldin:
For us this actually breaks some functionality of UV for us. For instance when running `uv pip freeze` it will now fail to find the venv in devenv;

```
$ echo $UV_PYTHON
$ uv -v pip freeze
DEBUG uv 0.6.17
DEBUG Checking for Python interpreter at path `/nix/store/g9cnd5v4a4zyv010pg22fx88kzaqljjj-python3-3.12.7-env/bin/python3.12`
Using Python 3.12.7 environment at: /nix/store/g9cnd5v4a4zyv010pg22fx88kzaqljjj-python3-3.12.7-env

$ UV_PYTHON= uv -v pip freeze|head -5
DEBUG uv 0.6.17
DEBUG Searching for default Python interpreter in virtual environments or search path
DEBUG Found `cpython-3.12.7-linux-x86_64-gnu` at `/home/.../.devenv/state/venv/bin/python3` (active virtual environment)
Using Python 3.12.7 environment at: .devenv/state/venv
adbc-driver-manager==1.2.0
adbc-driver-postgresql==1.2.0
agate==1.9.1
aiobotocore==2.17.0
aiohappyeyeballs==2.4.3
```

sandydoo:
@d-goldin, could you please open an issue with a minimal repro?

I'd like to add more tests for the various (and way too many) ways to set up a python environment so that we can track regressions like this.

gabyx:
@d-goldin : This `UV_PYTHON` variable is probably a bit problematic to set: A few observations:
Why maybe we should revert it:

- If its set: certain commands like `uv pip install` etc will forcefully use it and try to use the `/nix/store` version to install packages which will not work. When a `uv venv` activated environment ( `source ./devenv/state/venv/activate`) is used: `uv` will **anyway** use `UV_PYTHON` which might seem correct but for certain commands such as `uv pip install` will not work as its similar to using `--python`  which forcefully sets the interpreter which does not work. Since `source .../bin/activate` does not set `UV_PYTHON` (as its only for the `uv` CLI which is correcty behaviour) there is no way any `uv pip install` will work except resetting `UV_PYTHON=` which is clearly not what we want users to really do.

I think setting `UV_PYTHON` (my PR) is a mistake as I thought UV will use that only to drive initial installation of `uv venv` etc. but thats not the case. So I strongly suggest to quickly revert my PR as the whole setup is more wrong now.
Sorry for this mistake its not good.

domenkozar:
Sounds good to me, can you make a PR?

gabyx:
See #2024

gabyx:
P.S. Actually its sad that `uv` does not have a way to control that behavior. I created an issue https://github.com/astral-sh/uv/issues/14748, maybe that will improve int the future.

frederiksemmel:
Hi @gabyx, thanks for looking into this! I also stumbled on this. What do you think about setting `UV_PYTHON_PREFERENCE=system`.

[uv docs on python-preference](https://docs.astral.sh/uv/reference/settings/#python-preference)

We set that in the `env` section of `devenv.nix` to make uv environments more consistent across NixOS and non NixOS systems. Currently only this is set: `UV_PYTHON_DOWNLOADS = "never";`. That is not enough in the case that a compatible python was already downloaded on non NixOS systems, then uv might still use that.

I still think a proper fix upstream like you suggested is the best option, but this might be worth a PR as a temporary solution. I will open it, if you agree this is helpful.

Potential issue with this:
Some purely non NixOS devenv projects might prefer using the uv provided pythons, because things like torch + cuda sometimes work without any hacks (like #1973). But I think that these projects already have to hack around the `UV_PYTHON_DOWNLOADS=never` status quo, so it shouldn't be a widespread issue.

gabyx:
@frederiksemmel: Thanks for the pointer: Jeah the env variable as explained in the link makes sense to set IMO. I guess there are some orthogonal usecases here:

1. I think all we want is that you should prohibit in any way the usage of another python than the one provided from Nix.
      So always picking up the python from Nix and at the same time making a `.venv` folder in the repository (set by `VIRTUAL_ENV=.devenv/state/venv` which is good.
      The `UV_PYTHON` is not the right variable at least how its handled by `uv`. I probably understood the tooling sequence a bit wrong with `uv`, so far I experience that `uv sync` will correctly use the projects virtual env to install all packages (even when VIRTUAL_ENV points to a system python env... due to `UV_PROJECT_ENVIRONMENT`.
       `uv` is  a bit opaque and IMO has to many knobs already how it selects the python interpreter etc... although the behavior is more sane than `poetry`.
      
- 2. If you want to use another system python for the project, I think currently with:

   ```nix
   languages.python = {
      enable = true;
      venv.enable = true;
      uv = {
        enable = true;
        package = pkgs.uv;
        sync = {
          enable = true;
          allExtras = true;
        };
      }; 
    };
   ``` 
   it will directly create the venv. And you can override it with 
   `rm -rf .devenv/state/venv && UV_PYTHON=/mysystem/bin/python uv sync -v` which will pick the python and create a virtual env again in `.devenv/state/venv` etc.
   
 Currently I am bit lost, how one should configure python.  Maybe there should be an option like `package = null` (@domenkozar what do you think) which will not set `UV_PYTHON_DOWNLOAD=never` and place no `python` at top of `PATH`). This will make the Nix shell just have `uv` but not manage any python, which is a bit contrary to what we want with Nix, determinism and no system stuff. But python env. with AI Cuda/OpenGL pytorch and shit is just a hell of a mess ....
 
 Can we not just use `languages.python = { package =     (python3.withPackages (ps: [ ps.pytorch ])) }; }` or somewthing like that?

gabyx:
@frederiksemmel : I would not tinker with `UV_PYTHON_PREFERENCE` cause IMO its not needed, even if devenv would provide an option like `package = null` which will not populate a python interpreter etc...

frederiksemmel:
@gabyx Thanks for the context! Lets see how uv evolves in this regard. Maybe they add another knob that means: "Choose this python for creating the venv, and then work with the python from the venv for commands like `uv pip install`".

Just to clarify, we are setting `UV_PYTHON_PREFERENCE=system` to make uv choose the first python on `PATH` that is not uv provided. The one it finds is then consistently the one from nix (which is what we want). I am unsure if I explained this well in the previous comment.

Regarding the `languages.python = { package =     (python3.withPackages (ps: [ ps.pytorch ])) }; }`, that is probably the best solution to my situation in #1973. I think I didn't manage to make it work because I had issues with uv choosing the uv provided python. I think this will still require nixglhost or nixGL for cuda support on a non NixOS system, but it has the advantage of cuda support on a NixOS system (which I cannot test because I don't have one). I will test it and add my findings to the issue, in case anyone else runs into a similar situation.

gabyx:
> Just to clarify, we are setting UV_PYTHON_PREFERENCE=system to make uv choose the first python on PATH that is not uv provided. The one it finds is then consistently the one from nix (which is what we want). I am unsure if I explained this well in the previous comment.

Ah thanks for clarfication: In that case I think it would be good to add. I opened a PR on this: https://github.com/cachix/devenv/pull/2080

gabyx:
https://github.com/astral-sh/uv/issues/15709

PR #1961, opened 2025-06-27
Add readiness probes for elasticmq

penguincoder:
By default the SQS-REST endpoint will listen on port 9324. If you change this value, I give you the escape hatches to change where to ask the server `/health`. The test in `examples/` has been updated to query the server for the readiness endpoint, as well.

4 seconds may be a long time for a default installation, but that should (hopefully) be sufficient for any customized installations to finish booting.

### DIFF
--- a/examples/elasticmq/.test.sh
+++ b/examples/elasticmq/.test.sh
@@ -1,6 +1,10 @@
 #!/usr/bin/env bash
 set -ex
 
+wait_for_port 9324 60
+
+curl -s http://localhost:9324/health
+
 wait_for_port 9325 60
 
 QUEUE_NAME=$(curl http://localhost:9325/statistics/queues -s | jq .[].name -r)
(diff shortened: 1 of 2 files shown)


domenkozar:
Would we be able to get the host out of the settings?

penguincoder:
possibly, yes. there are two configuration stanzas for `elasticmq`. One is for the SQS endpoint, the other is for the HTTP point-and-click interface. The `rest-sqs` one, though, uses `bind-port` and `bind-hostname` which are unique slugs to that endpoint and that's great... but only if the user generated configuration file has defined them. Otherwise they won't exist and we will be back to using the defaults (these values are the default values).

is there an existing service that pulls the configuration value out of a file already?

domenkozar:
@penguincoder I wonder if these are not part of cfg.settings, we should expose other settings? Not sure what's the best way here :)

domenkozar:
#1961

PR #1992, opened 2025-07-07
nix: fix several regressions related to git introduced in v1.7

### LINKED ISSUE #1990: v1.7 doesn't detect file updates if used outside of a git repo
**Describe the bug**

The Nix bump in v1.7 introduced a regression in projects that aren't a git repository.
The caching system fails to pick up file changes, likely due to Nix mangling the paths somehow.

**To reproduce**

```
devenv init
# Modify devenv.nix and reload 
# Or view .devenv/input-paths.txt. It won't have devenv.nix listed.
```

**Version**

devenv 1.7.0 (aarch64-darwin)

### LINKED ISSUE #2001: devenv 1.7 getting stuck when entering shell
**Describe the bug**

update devenv from 1.6 to 1.7 ,  it getting stuck when entering shell

seems stuck in the step `evaluating file '<nix/derivation-internal.nix>`
Image

**Version**
❯ devenv version
devenv 1.7.0 (x86_64-linux)

### LINKED ISSUE #2002: v1.7 doesn't resolve relative non-flake inputs to a path on the filesystem
Nix 2.30 copies all git inputs to the store and then fails to resolve the paths. 2.24 would keep around the `outPath` and would resolve the `root` input to the local filesystem.

> Hey @sandydoo is that the case of [this issue](https://github.com/emdneto/opentelemetry-by-example/actions/runs/16168813864/job/45637008655#step:8:623) as well?
> 
> I'm using it in my project like this:
> ```yaml
> inputs:
>   nixpkgs-unstable:
>     url: github:NixOS/nixpkgs/nixpkgs-unstable
>   nixpkgs:
>     url: github:cachix/devenv-nixpkgs/rolling
>   root:
>     url: git+file://../
>     flake: false
> imports:
>   - root/common
>   - root/common/python.nix
> ```
> 
> After 1.7, my CI stopped working. It was working fine with 1.6. Is there any workaround for this use case? 

 _Originally posted by @emdneto in [#1990](https://github.com/cachix/devenv/issues/1990#issuecomment-3053061053)_

sandydoo:
Fixes several regressions in v1.7 to do with git.

#### Path resolution outside of git repos

Nix 2.30 introduced an issue where local paths were copied to the Nix store when not part of a git repo. This broke the caching system because, once copied to the store, Nix would lose track of the context by not using the file system accessor for paths.

#### Path resolution for non-flake git+file inputs

Some people uses `git+file` inputs to work around flake limitations and compose shared devenvs in a monorepo. Nix 2.30 broke the path resolution here as well, causing the paths to be resolved to the Nix store.

Testing https://github.com/cachix/nix/pull/7.

Fixes #1990.
Fixes #2002.
Fixes #2001.

##### Likely related (missing more information from OPs)

#1998


domenkozar:
Can we add a test so we can catch this next time?

sandydoo:
> Can we add a test so we can catch this next time?

Yeah, I'm thinking about how to do this.

We have integration tests for detecting files, but those use eval and this is a flake-specific issue.
And devenv-run-tests initializes a git repo to fix tests that use git-hooks, but that would fail to test for this issue.

sandydoo:
Added #1995 to test this edge-case. I've cherry-picked the commit here, so the eval-cache tests should fail in #1995 and pass here.

sandydoo:
Final test run.

PR #2010, opened 2025-07-15
Fix container compatibility with Gitlab CI

allonhadaya-maven:
This commit updates the default `config.containers.shell.entrypoint` to execute the startup command with exec rather than as a shell call. This allows Gitlab CI runnners to properly send scripts to CI jobs running the shell container image.

For [reference][gitlab], here's how Gitlab runners bring up a container and run the script:

> 1.The runner starts a Docker container using the defined entrypoint. The
> default from Dockerfile that may be overridden in the .gitlab-ci.yml file.
>
> ...
>
> 4. The runner sends the script to the container’s shell stdin and receives
> the output.

---

This pattern of integrating with Gitlab CI produces improved latency over the [current recommended pattern][devenv-ci]:
- pipelines build the shell once and then reuse the image in each shell CI job
- pipelines can skip unnecessary build jobs using CI rules
- shell CI jobs only load the shell image, not the large [devenv image][devenv-image]
- shell CI jobs run script commands as soon as the job container is ready

Users can implement this pattern as follows:

```nix
# devenv.nix

{ pkgs, ... }:
{
  env.GREET = "hello";

  packages = [ pkgs.jq ];

  containers.shell.name = "hello-world";
}
```

```yaml
# .gitlab-ci.yaml

# Build the shell container image and copy it to <registry>/hello-world
# Note: defining change rules allows users to skip unnecessary builds

build:
  stage: build
  image: ghcr.io/cachix/devenv/devenv:v1.6
  script:
    - devenv container copy shell
  rules:
    - changes:
      - devenv.nix
      - devenv.lock

.devenv-shell:
  image: <registry>/hello-world
  needs:
    - build

# Run CI workloads using the shell container image
# Note: script runs as soon as the job container is ready

greet:
  stage: test
  extends: .devenv-shell
  image: <registry>/hello-world
  script:
    - echo $GREET

jq-version:
  stage: test
  image: <registry>/hello-world
  script:
    - jq --version
```

[gitlab]: https://docs.gitlab.com/ci/docker/using_docker_images/#override-the-entrypoint-of-an-image
[devenv-ci]: https://github.com/cachix/devenv/blob/v1.6.1/docs/integrations/devenv-container.md?plain=1#L14-L20
[devenv-image]: https://github.com/cachix/devenv/pkgs/container/devenv%2Fdevenv/402654685?tag=v1.6

### DIFF
--- a/src/modules/containers.nix
+++ b/src/modules/containers.nix
@@ -30,9 +30,12 @@ let
     source ${shell.envScript}
 
     # expand any envvars before exec
-    cmd="`echo "$@"|${pkgs.envsubst}/bin/envsubst`"
+    expanded_args=()
+    for arg in "$@"; do
+        expanded_args+=("$(printf '%s' "$arg" | ${pkgs.envsubst}/bin/envsubst)")
+    done
 
-    ${bash} -c "$cmd"
+    exec "''${expanded_args[@]}"
   '';
   user = "user";
   group = "user";


domenkozar:
See ac2e743b492d0ad4fa11f6a0babd279d179fbf01 for why bash was used /cc @mcdonc

allonhadaya-maven:
> See [ac2e743](https://github.com/cachix/devenv/commit/ac2e743b492d0ad4fa11f6a0babd279d179fbf01) for why bash was used /cc @mcdonc

Ah, thanks for providing that context!

This will represent a breaking change for users who currently have startup commands defined like so:

```nix
# No longer supported:
{...}:
{
  containers.shell.startupCommand = "if [ 1 = 1 ]; then echo yup; fi";
}
```

Users who wish to use shell builtins in their startup command should call the shell explicitly:

```nix
# Do this instead:
{...}:
{
  containers.shell.startupCommand = "bash -c 'if [ 1 = 1 ]; then echo yup; fi'";
}
```

I'd recommend moving forward with this breaking change since moving back to `exec` enables container runtimes to properly supervise the resulting containerized process. See [Best practices `ENTRYPOINT`](https://docs.docker.com/build/building/best-practices/#entrypoint):

> For example, the [Postgres Official Image](https://hub.docker.com/_/postgres/) uses the following script as its ENTRYPOINT:
>
> _(code block omitted)_
>
> This script uses [the exec Bash command](https://wiki.bash-hackers.org/commands/builtin/exec) so that the final running application becomes the container's PID 1. This allows the application to receive any Unix signals sent to the container.

domenkozar:
Could we: exec bash -c $cmd?

domenkozar:
@allonhadaya-maven just wondering if that breaks semantics, otherwise I'm +1 for this change!

allonhadaya-maven:
> @allonhadaya-maven just wondering if that breaks semantics, otherwise I'm +1 for this change!

@domenkozar that might work! Towards the end of this week I'll have a chance to test that entrypoint configuration and also follow up with those [documentation updates](https://discord.com/channels/1036369714731036712/1392878645504118976/1394757329370746910) we discussed.

domenkozar:
ping :)

domenkozar:
@allonhadaya-maven hey, is this still needed?

PR #2020, opened 2025-07-17
rust: allow clippy git-hook to access internet by default

sandydoo:
`--offline` mode is set by default to work with `nix flake check`. For devenv, we need to allow clippy to fetch dependencies because there's no other method to do so ootb.


domenkozar:
We might need to use isFlake boolean here

sandydoo:
> We might need to use isFlake boolean here

Why though? You'd want to use online mode in `nix develop` as well.

PR #2021, opened 2025-07-17
tasks: add property tests

sandydoo:
Builds on top of #2018.

1. prop_prefix_matching_correctness - Tests that namespace prefix matching works correctly
2. prop_exact_task_matching - Tests that exact task names are matched correctly
3. prop_trailing_colon_equivalence - Tests that "ci" and "ci:" are equivalent
4. prop_namespace_exclusion - Tests that only tasks with the correct prefix are matched
5. prop_hierarchical_inclusion - Tests hierarchical namespace matching

Fixed a bug in the schedule() method where node indices were not being properly updated after creating a subgraph. This was causing test helper functions to fail when trying to extract task names from the graph.

### DIFF
--- a/devenv-tasks/src/tests/mod.rs
+++ b/devenv-tasks/src/tests/mod.rs
@@ -2519,7 +2519,10 @@ mod property_tests {
         }))
         .unwrap();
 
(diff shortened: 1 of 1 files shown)


domenkozar:
Shall we merge this?

PR #2024, opened 2025-07-19
fix: revert wrong `UV_PYTHON` env. variable

gabyx:
Fixes issues mentioned in #1959.


domenkozar:
Seems like it includes a lot of formatting as well, can we reduce it to the change itself?

Issue #1387, opened 2024-08-15
clang tools is added to packages does not add it to my dev shell


Redhawk18:
**Describe the bug**
I added a nix flake to this project, but I need to still run `nix-shell -p clang-tools` everytime or I get import errors. Does anyone know why?

**To reproduce**
https://github.com/Redhawk18/RandomX/blob/master/flake.nix

k3yss:
@Redhawk18 

`clangd` works for me without doing the `nix-shell` step which is a tool which comes under `clang-tools`.
` ``
[k3ys@nixos:~/sandbox/cachix/issues/flakes]$ clangd
clangd is a language server that provides IDE-like features to editors.

It should be used via an editor plugin rather than invoked directly. For more information, see:
https://clangd.llvm.org/
https://microsoft.github.io/language-server-protocol/

clangd accepts flags on the commandline, and in the CLANGD_FLAGS environment variable.

I[02:12:23.767] clangd version 18.1.8
I[02:12:23.767] Features: linux
I[02:12:23.767] PID: 210840
I[02:12:23.767] Working directory: /home/k3ys/sandbox/cachix/issues/flakes
I[02:12:23.767] argv[0]: clangd
I[02:12:23.767] Starting LSP over stdin/stdout

```

You should post the specific import error that you mentioned, otherwise we won't know what's the cause of the error.

Redhawk18:
> @Redhawk18
> 
> `clangd` works for me without doing the `nix-shell` step which is a tool which comes under `clang-tools`.
> 
> ```
> [k3ys@nixos:~/sandbox/cachix/issues/flakes]$ clangd
> clangd is a language server that provides IDE-like features to editors.
> 
> It should be used via an editor plugin rather than invoked directly. For more information, see:
> https://clangd.llvm.org/
> https://microsoft.github.io/language-server-protocol/
> 
> clangd accepts flags on the commandline, and in the CLANGD_FLAGS environment variable.
> 
> I[02:12:23.767] clangd version 18.1.8
> I[02:12:23.767] Features: linux
> I[02:12:23.767] PID: 210840
> I[02:12:23.767] Working directory: /home/k3ys/sandbox/cachix/issues/flakes
> I[02:12:23.767] argv[0]: clangd
> I[02:12:23.767] Starting LSP over stdin/stdout
> ```
> 
> You should post the specific import error that you mentioned, otherwise we won't know what's the cause of the error.

standard cpp files that are apart of the c++ standard libraries are saying not found
[...]

PR #2029, opened 2025-07-23
Fix clangd error: compile_commands.json not found when enabling languages.c and languages.cplusplus

Emtyffx:
Adding clang after clang-tools in dependency chain can cause issue with clangd of kind (compile_commands.json
Fixes issue from #1387 by adding clangd(clang-tools) as a dependency into languages.cplusplus / languages.c
This issue was also mentioned in Nixpkgs: https://github.com/NixOS/nixpkgs/issues/76486

Important note: clang-tools must be added before clang (there's some issue in initialization if otherwise)

### DIFF
--- a/src/modules/languages/c.nix
+++ b/src/modules/languages/c.nix
@@ -23,6 +23,7 @@ in
 
   config = lib.mkIf cfg.enable {
     packages = with pkgs; [
+      clang-tools
       stdenv
       gnumake
       ccls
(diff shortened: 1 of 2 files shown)


domenkozar:
Nice! Let's see if it works for macos and linux.

Emtyffx:
Tested this on linux:
when in devenv environment with clangd enabled in neovim

this works:
```
{
languages.cplusplus.enable = true;
packages = with pkgs; lib.mkBefore [clang-tools];
}
```
this doesn't(clangd just doesn't recognize std and LspLog shows that compile_commands.json not found which causes clangd to fail):
```
{
languages.cplusplus.enable = true;
packages = with pkgs; [clang-tools];
}
```

PR #2038, opened 2025-07-25
Remove --frozen and --no-install-workspace from default uv.sync arguments

extrange:
If `--frozen` is specified, then uv just uses the existing lockfile as-is, without checking `pyproject.toml`, defeating the purpose of the checksum calculated on `pyproject.toml`.


domenkozar:
cc @lamalex

Issue #998, opened 2024-03-06
mono repo importing


domenkozar:
With #745 it will be possible to import devenv.nix across directories in a git repo.

But it's a bit cumbersome to use:

```
inputs:
  root:
    url: git+file://.
    flake: false
imports:
- root/examples/compose/projectA
```

I wonder if we should come up with something like `root://` to support imports from top-level of the git repo.

domenkozar:
Maybe we could support `../projectA` syntax by infering the git root and pwd, then doing the path calculations.

A bit easier to support is `- /examples/compose/projectA`

rawkode:
I was literally opening an issue to request the ability to do:

```
imports:
- /
- /moduleA
```

Where `/` imports the git root devenv.nix and `/moduleA` is `./moduleA` from the git root.

rawkode:
Relative imports would also be appreciated for shared services deep within the monorepo.

How difficult would this be to support, @domenkozar ?

clotodex:
To add to this issue. I would like to lock some imports in a common way.
Simple example - I want the same EXACT version of flutter in projectA and projectB.
Relative Imports are the first step to import a common.nix
But I am currently running helper scripts to make sure that both projects use the same version, since the import is locked individually.
Does nix even allow this?

dsalaza4:
Looking forward to this! The `/absolute/path/from/root` way looks great

rawkode:
I believe this is blocked by #1548

domenkozar:
@dsalaza4 @rawkode does https://github.com/cachix/devenv/pull/2049 solve your use case?

domenkozar:
Coming out in 1.10!

PR #2049, opened 2025-07-29
Support monorepos

domenkozar:
Fixes https://github.com/cachix/devenv/issues/998

[docs preview](https://monorepos.devenv.pages.dev/guides/monorepo/)


domenkozar:
I've split up `devenv.yaml` merging into https://github.com/cachix/devenv/pull/2163

KevinEdry:
@domenkozar in a JS/TS monorepo it's usually the case to require a package within the monorepo, for example: https://yarnpkg.com/features/workspaces

The common scenario for us and many others is when our backend and frontend services both require a 3rd package within the monorepo (common utilities for example), and to use it we need to make sure that package is installed and built.

While this change helps share Nix configs, I don't think it will auto-run the nix for the 3rd package when it's required as a dependency by other packages in the monorepo. 

For instance, if I have:
```
/packages/shared-utils/  (needs to be built first)
/packages/frontend/      (depends on shared-utils)  
/packages/backend/       (depends on shared-utils)
```

When I run `devenv up` in `/packages/frontend/`, it should ideally:
1. Detect that `shared-utils` is a dependency 
2. First build/prepare `shared-utils` 
3. Then start the frontend environment

This is similar to how tools like Turborepo, Nx, or Rush handle task orchestration in monorepos - they automatically run prerequisite tasks for internal dependencies.

Without this, developers would need to manually ensure all internal dependencies are built in the correct order before starting their environment, which defeats much of the convenience that monorepo tooling is supposed to provide.

Is there a plan to add dependency-aware task orchestration, or would this need to be handled separately through something like `tasks` with explicit `before` dependencies?

domenkozar:
Merged in https://github.com/cachix/devenv/pull/2163

gabyx:
@domenkozar: Does that work in `flake.nix` currently with exactly 1.10 I see this as `null`. and `nix develop --no-pure-eval` etc

PR #2067, opened 2025-08-02
flakes: reload shell on `up`/`test` and remove the need for separate packages

### LINKED ISSUE #2065: Tasks and processes not detecting changes with direnv + flakes
> I have a related issue, I'm loading a devenv shell (defined via flake-parts) using direnv, but when running `devenv up` after making changes, it doesn't seem to reload the processes/tasks even though there is some nix evaluation happening. Most likely, it's inheriting `${DEVENV_TASKS}` from the shell, but I would expect the `devenv` shim to replace it with fresh values if it's already evaluating the shell.
> 
> `direnv reload` resolves this issue 

 _Originally posted by @mrene in [#2061](https://github.com/cachix/devenv/issues/2061#issuecomment-3145953738)_

### LINKED ISSUE #2079: `devenv up` using the `procfileScript` does not start up
**Describe the bug**
In latest `main`: (`v1.8.1`): 
when I load the flake and on the devShell in question do in `nix repl`

`:b devShells.x86_64-linux.test-services.config.procfileScript`

and then inspect the used `process-compse.yaml` in this file 

- it does not contain the env variables which where there on `v1.7`.
- it uses devenv-tasks (which might set the env variables?)
   to run the tasks but this command does not work:

  ```shell
  /nix/store/g6jkgjsl8j1xq52pf4df509dchvfcbjd-devenv-tasks-1.8.1/bin/devenv-tasks run --mode all devenv:processes:mongodb-configure
   Error: TaskNotFound("devenv:processes:mongodb-configure"
  ```

So the `procfileScript` which is used in `devenv up` (see [documentation](https://devenv.sh/guides/using-with-flakes/#multiple-shells)) in the flake integration wrapper, does not work as expected. It closes directly as `devenv-tasks` fails to lunch the tasks.

**Version**
1.8.1

sandydoo:
This is a large-ish refactor of the flakes compatibility module.

#### Evaluates the latest processes and tests

Uses `nix develop` to prepare the latest shell env before running processes and tests. This ensures that processes run inside the latest shell. 

direnv can only reload the shell _after_ the command, so running `devenv up` immediately after changing the flake would give you stale output unless you first reload. The extra `nix develop` gives you the latest changes at the cost of an extra eval.

This extra eval is disabled if we detect that direnv is being used.

Fixes #2065.
Fixes #2079.

#### Ideas from #1142

Q: This is related to https://github.com/cachix/devenv/pull/1142. Maybe we should expose packages instead of env vars?
A: Done!

#### Configurable command runner

Q: Should we make the `nix develop` configurable for those who don't want the performance penalty?
A: Implemented for the direnv use-case.

#### flake-parts and pure shell

Q: How does this integrate with flake-parts. If we have a devenv.root, we don't need the impure.
A: The existing implementation didn't handle this. We could check whether root is set and make a decision? Maybe it needs to be configurable.


flake-module.nix
```diff
@@ -60,7 +60,9 @@ devenvFlake: { flake-parts-lib, lib, inputs, ... }: {
                 { "${shellPrefix shellName}container-${containerName}" = container.derivation; }
```
sandydoo, line 68: Didn't know we had containers wired up in the flakes module. This should be exposed in the flake shim.

Issue #1973, opened 2025-07-02
Python: the wrapped python breaks nixglhost and NixGL (torch looses cuda support)


frederiksemmel:
**Describe the bug**

First of all, awesome project, thank you very much!

Using the python environment from devenv on a non-NixOs system doesn't work together with [nixglhost](https://github.com/numtide/nix-gl-host). The torch from pypi doesn't work with cuda, but with nixglhost it should work. Changing `--prefix` to `--suffix` in the python wrapper would fix this, should I open a PR for this?

nixglhost sets the `LD_LIBRARY_PATH` such that OpenGL/Cuda drivers are loaded correctly on non-NixOs systems. It is very useful for environments that use pytorch and tensorrt.

The python wrapper from devenv **prefixes** the `LD_LIBRARY_PATH` with `.devenv/profile/lib`. Therefore when `torch` loads the cuda libraries, it finds the stub from `.devenv/profile/lib` instead of the version provided by `nixghost`.

The python wrapper from devenv currently looks like this:

```
  package = pkgs.callPackage ../../python-wrapper.nix {
    python = cfg.package;
    requiredPythonModules = cfg.package.pkgs.requiredPythonModules;
    makeWrapperArgs = [
      "--prefix"
      "LD_LIBRARY_PATH"
      ":"
      libraries
    ] ++ lib.optionals pkgs.stdenv.isDarwin [
      "--prefix"
      "DYLD_LIBRARY_PATH"
      ":"
      libraries
    ];
  };
```

The devenv python works nicely with `nixglhost` if we change `--prefix` to `--suffix` in that python wrapper.

Note on a separate python issue: Since this PR #1959 the python wrapper is not used for `uv` anymore. This is a separate issue, in the gist below, I have set UV_PYTHON="" to work around that.

**To reproduce**

https://gist.github.com/frederiksemmel/acbb88c69f53e66e80c361a450c0a324

1. Run on system that is not NixOs with nixglhost installed, an Nvidia GPU, and working cuda setup
3. Set up devenv python with `uv.enable = true`
4. Use uv to add `torch` and do `uv sync` to set up the venv
5. Run `nixglhost python -c "import torch; print(torch.cuda.is_available())"`. This will print "False" and some error about a cuda sub file.

**Version**

devenv 1.6.1 (x86_64-linux)

frederiksemmel:
I cannot reproduce this anymore... No idea what went wrong, sorry for the noise.

PR #2069, opened 2025-08-04
python: LD_LIBRARY_PATH suffix instead of prefix

frederiksemmel:
This fixes the nixglhost and nixGL problems described in #1973.

It is relevant for systems that are not on NixOS and want to use cuda inside devenv.

I am not sure if there are any disadvantages of setting the python as a suffix instead of a prefix in `LD_LIBRARY_PATH`. If it is set as a prefix there is no way for the user to override what libraries are loaded, `./devenv/profile` always takes preference. As a suffix, we give more flexibility to the user. It is always possible to set the `LD_LIBRARY_PATH` variable in the environment variables in devenv.nix.

Let me know if you want me to provide more infos, or testing.

### DIFF
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -19,13 +19,13 @@ let
     requiredPythonModules = cfg.package.pkgs.requiredPythonModules;
     makeWrapperArgs =
       [
-        "--prefix"
+        "--suffix"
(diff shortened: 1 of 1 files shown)


frederiksemmel:
I cannot reproduce the issue anymore. No idea what was wrong. Sorry for the noise.

PR #2070, opened 2025-08-04
devenv: configure Nix to use the Cachix auth token

### LINKED ISSUE #2068: devenv cachix integration with cache that requires authorization fails on first run but succeeds on second
**Describe the bug**

When attempting to use the devenv cachix integration in a devenv nix that looks like this:

```nix

let
  deriv = pkgs.writeShellScriptBin "testderiv" ''
    echo yup
  '';
in
{

  packages = [ deriv ];

  scripts.hello.exec = ''
    echo hello
  '';

  enterShell = ''
    hello
  '';

  cachix.pull = [ "enfoldsystems" ];
  cachix.push = "enfoldsystems";

}
```

``enfoldsystems`` is not a public cache.  It requires authorization.

The first time you run ``devenv shell`` against this ``devenv.nix`` it will complain:

```
• Using Cachix caches: devenv, enfoldsystems, enfoldsystems
warning: unable to download 'https://enfoldsystems.cachix.org/nix-cache-info': HTTP error 401

         response body:

         Binary cache enfoldsystems doesn't exist or you're not authorized to access. Please check the name for typos and/or provide an auth token. Contact support on https://cachix.org if you're stuck!
```

The second time you run it even with ``--no-eval-cache`` or rm -rf'ing ``.devenv*``, it has no issue.

**To reproduce**

https://github.com/mcdonc/cachixauthdebug/blob/main/.github/workflows/cachix-populate.yaml

I am happy to give out access to this repo so you can run the action.

A failing test run is at https://github.com/mcdonc/cachixauthdebug/actions/runs/16701503514

May be subtly related to #2065 and/or #2059

**Version**

1.6.1, 1.7, and 1.8.1

sandydoo:
Sets up a netrc file for Nix to use for auth against binary caches when `CACHIX_AUTH_TOKEN` is provided and `cachix.pull` is non-empty.

The downside here is that `--netrc-file` is not additive. Nix won't concatenate the configured netrc files, so any credentials already configured, e.g. in `~/.config/nix/netrc`, won't be used.

Fixes #2068 (TDB).

### DIFF
--- a/devenv/src/nix.rs
+++ b/devenv/src/nix.rs
@@ -485,6 +485,7 @@ impl Nix {
         let mut final_args = Vec::new();
         let known_keys_str;
         let pull_caches_str;
+        let mut netrc_path_str: Option<String> = None;
         let mut push_cache = None;
 
         if !self.global_options.offline {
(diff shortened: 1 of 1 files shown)


devenv/src/nix.rs
```diff
@@ -526,10 +527,37 @@ impl Nix {
                             &known_keys_str,
                         ]);
                     }
+
+                    // Configure netrc file for auth token if available
+                    if !cachix_caches.caches.pull.is_empty() {
```
domenkozar, line 532: if these two conditionals are false, we should remove the file

PR #2085, opened 2025-08-08
devenv: support containers on macOS

sandydoo:
This unlocks the container restriction on macOS. You will need an `<arch>-linux` builder already set up in your nix.conf to build the container.


domenkozar:
Nice :) Let's merge, needs some rebasing :)

sandydoo:
After the TUI work, I think we should always pass commands through `CacheCommand`. Not for caching, but to get access to the parsed logs and capture the errors. This would let us detect the missing builder error and suggest ways to set one up.

domenkozar:
@sandydoo let's merge this?

PR #2091, opened 2025-08-11
fix: `nixpkgs` instead of `inputs.nixpkgs` should be used in `mkConfig`

gabyx:
I was stumbling accross the `inputs.nixpkgs` use in `mkConfig` and it seemed to me that it would make more sense to use the `nixpkgs` input from `devenv`'s flake instead of what the user inputs to `mkConfig { inputs = {...} }`.

Maybe I am missing the point of using `inputs.nixpkgs` which the users provided inside `mkConfig` :smile:
Shouldn't the devenv use its rolling version `evalModules` etc... with what it was tested to be more stable?

 Currently `mkConfig` needs `inputs.self`, `inputs.devenv` and `inputs.nixpkgs`. I think one can get rid of `inputs.nixpkgs` and replace it with `nixpkgs`.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -328,12 +328,12 @@
             moduleInputs = {
               inherit git-hooks;
             } // inputs;
-            project = inputs.nixpkgs.lib.evalModules {
+            project = nixpkgs.lib.evalModules {
               specialArgs = moduleInputs // {
                 inputs = moduleInputs;
               };
               modules = [
-                { config._module.args.pkgs = inputs.nixpkgs.lib.mkDefault pkgs; }
+                { config._module.args.pkgs = nixpkgs.lib.mkDefault pkgs; }
                 (self.modules + /top-level.nix)
                 (
                   { config, ... }:
--- a/flake.nix
+++ b/flake.nix
@@ -312,33 +312,46 @@
 
       lib = {
         mkConfig =
-          args@{ pkgs
-          , inputs
+          args@{ inputs
           , modules
+          , pkgs ? null
+          , system ? null
           ,
           }:
           (self.lib.mkEval args).config;
         mkEval =
-          { pkgs
-          , inputs
+          { inputs
           , modules
+          , pkgs ? null
+          , system ? null
           ,
           }:
           let
+            # If the user passed a nixpkg we use it to do the evaluation.
+            # By default we take the nixpkgs with which devenv was tested.
+            moduleLib = if nixpkgs.lib.hasAttr "nixpkgs" inputs then inputs.nixpkgs.lib else nixpkgs.lib;
+            # If the user passed a `pkgs` we used it.
+            # By default we take the pkgs which devenv was tested against.
+            modulePkgs =
+              if pkgs == null then
+                assert system != null "System must be specified when no `pkgs` argument is used.";
+                moduleLib.legacyPackages.${system} else pkgs;
+
             moduleInputs = {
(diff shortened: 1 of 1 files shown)


gabyx:
addendum: Als there is the question why not defaulting the `pkgs` to `pkgs ? nixpkgs.legacyPackages.${system}` which takes the correct inputs as devenv was tested with.
But then I guess the `mkConfig` needs a system paramter which is .... argh.....

sandydoo:
In other words, we always had an implicit `follows` because were always using the root flake's `nixpkgs` input.

The concern I have is what happens when a user deliberately passes in a nixpkgs input? Instead of ignoring the devenv flake's nixpkgs input, we're now ignoring the argument. 

I think the best approach here is to have a fallback. Use the flake's nixpkgs input by default. This makes `nixpkgs` an optional `input` in `mkConfig`. If we get a `nixpkgs` in `inputs`, then it takes precedence.

gabyx:
Jeah: thats true.
I can do this.

gabyx:
@sandydoo : I will introduce the fallback for the backwards compatibility:
My question is just, do you really want this this backwards compatibility or is it not just better to use always `nixpkgs`.

Pro: No fallback:
- Because its only used to evaluate the configuration and that functionality is bound to the version of `devenv`.
- The user already passes `pkgs` so it will construct it from `devenv/rolling` if one wants to be up-to-date with testing.

Contra: Fallback:
- The user should be able to choose which `nixpkgs.lib.evalModules` function gets used. Is this really the case? So you could have devenv on verison X and use version Z (Z > X) for the NixOS module system to do the eval?
  How would that work, cause devenvs module code in version X might not be compatible with the passed `nixpkgs` on verison `Z`.

With the Pro argument: the example in the docs is directly correct:

```nix
{
  inputs = {
    nixpkgs.url = "github:cachix/devenv-nixpkgs/rolling"; # The packages against devenv was tested.
  };

  nixConfig = {
    extra-trusted-public-keys = "devenv.cachix.org-1:w1cLUi8dv3hnoSPGAuibQv+f9TZLr6cv/Hm9XgU50cw=";
    extra-substituters = "https://devenv.cachix.org";
  };

  outputs = { self, nixpkgs, devenv, ... } @ inputs:
    let
      system = "x86_64-linux";
    in
    {
      packages.${system}.devenv-up = self.devShells.${system}.default.config.procfileScript;
      packages.${system}.devenv-test = self.devShells.${system}.default.config.test;

      devShells.${system}.default = devenv.lib.mkShell {
        inherit inputs pkgs;
        modules = [
          ({lib, pkgs, config, ... }: {
          
            # Important: `lib` is the nixpkgs lib with what devenv was tested.
            # You do not need `devenv.inputs.nixpkgs.follows = "nixpkgs"` anymore above.
            
            # This is your devenv configuration
            packages = [ pkgs.hello ];

            enterShell = ''
              hello
            '';

            processes.run.exec = "hello";
          })
        ];
      };
    };
}
```

gabyx:
@sandydoo : I tried to improve the `pkgs` and `inputs` arguments. Check the last commit, isn't this solution better?
So by default we use the stuff with what devenv was tested with.

sandydoo:
This code was refactored a while ago when fixing a `lib` incompatibility. `inputs.nixpkgs.lib` is no longer used. Users can provide an optional `lib` argument if necessary.

PR #2097, opened 2025-08-14
dynamodb: add sharedDb option

gregg-platogo:


src/modules/services/dynamodb-local.nix
```diff
@@ -13,7 +13,12 @@ let
 
     cd "${baseDir}"
 
-    ${config.services.dynamodb-local.package}/bin/dynamodb-local -port ${toString cfg.port} -dbPath ${baseDir} -disableTelemetry
+    extraFlags=""
+    if [[ "${toString cfg.sharedDb}" == "true" ]]; then
```
sandydoo: `toString` evaluates to `"1"` for true and `""` for false.
You're probably looking for [`lib.boolToString`](https://noogle.dev/f/lib/boolToString).

src/modules/services/dynamodb-local.nix
```diff
@@ -13,7 +13,12 @@ let
...
+      extraFlags="-sharedDb"
```
sandydoo:
```suggestion
      extraFlags+="-sharedDb"
```

src/modules/services/dynamodb-local.nix
```diff
@@ -32,6 +37,11 @@ in
       description = "Listen address for the Dynamodb-local.";
       default = 8000;
     };
+    sharedDb = lib.mkOption {
+      type = types.bool;
+      default = false;
+      description = "If true, enables the -sharedDb flag for DynamoDB Local.";
```
sandydoo: This should explain the effect of the option in practice. How does it change the functionality of dynamodb?

sandydoo (changes requested):

src/modules/services/dynamodb-local.nix
sandydoo: Or, better yet:
```suggestion
    if [[ "${toString cfg.sharedDb}" ]]; then
```

src/modules/services/dynamodb-local.nix
gregg-platogo: Addressed be726a01

src/modules/services/dynamodb-local.nix
gregg-platogo: Addressed be726a01

src/modules/services/dynamodb-local.nix
gregg-platogo: Addressed be726a01

sandydoo:
Thanks!

PR #2102, opened 2025-08-15
Fix rubyVersionFile example

teoljungberg:
The common naming convention is `.ruby-version` among rbenv(1), asdf(1),
etc.


sandydoo:
Thanks!

PR #2115, opened 2025-08-22
python: use uv2nix for packaging

domenkozar:
```nix
{ pkgs, lib, config, ... }: {

  outputs.myapp = config.languages.python.import ./. {};
}
```

### DIFF
--- a/src/modules/languages/python.nix
+++ b/src/modules/languages/python.nix
@@ -48,6 +48,27 @@ let
     follows = [ "nixpkgs" ];
   };
 
+  uv2nix = config.lib.getInput {
+    name = "uv2nix";
+    url = "github:pyproject-nix/uv2nix";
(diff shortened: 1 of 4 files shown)


rawkode:
Excited to try and use this! I've got a uv project. Subscribed.

takeda:
@rawkode I actually made a module for my own use https://github.com/takeda/devenv-uv2nix (it basically sets up the environment in such way that all the packages are available in the shell) but I'm also excited about it, perhaps this implementation will be done better and I won't need it at all.

Bullish-Design:
Ha, I'm finally realizing that I'm unable to use this for python because its not yet implemented! 

Is there anything holding up merging in the languages.python.import functionality? I promise you have at least one excited user!

domenkozar:
Let's merge this, at least the basics work :)

PR #2129, opened 2025-08-30
services/pocket-id: init

deadbaed:
I am using devenv for the first time, and I wanted to use [Pocket ID](https://pocket-id.org) for my project.

This PR adds a service for Pocket ID, with the ability to:
- Make it listen to a UNIX socket
- Store files inside `$DEVENV_STATE`
- Disable analytics

An example was added: Running Pocket ID on a UNIX socket with Caddy used as a reverse proxy


domenkozar:
Thanks! Could you merge the example into one, we unfortunately have a limit how many examples we can run

deadbaed:
done!

deadbaed:
The Claude CI gives a 401, nothing I can do about it.

Tests are failing, but it does not seem related to the `pocket-id` service.

The documentation CI gives the error `sed: can't read ./services/pocket-id.md: No such file or directory`, do I need to generate the docs locally and commit the markdown file?

domenkozar:
Looks like we're missing automation that generates new docs for new services.

PR #2146, opened 2025-09-10
init: add shellcheck directive to default .envrc

### LINKED ISSUE #2121: Autogenerated .envrc fails shellcheck hook in default configuration
Minor onboarding nitpick: 
.envrc file autogenerated during `devenv init` doesn't pass shellcheck by default so enabling the shellcheck git-hook makes it impossible to commit both files from within the devenv shell. It would be nice if either of these three things were implemented:
1. Autogenerated .envrc contained shellcheck disable directives
2. shellcheck hook excluded .envrc by default
3. Documentation for shellcheck hook included warning that this is the case and explained how to exclude .envrc manually.

I can submit a PR for this just tell me which approach would be preferable

sandydoo:
Tells shellcheck that the .envrc file is bash out of the box.

Fixes #2121.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,6 +1,5 @@
-#!/usr/bin/env bash
+# shellcheck shell=bash
 # ^ for code highlighting
-# Used by https://direnv.net
 set -euo pipefail
 
 # External users should eval `devenv direnvrc` or use `source_url` to load this file
--- a/devenv/init/.envrc
+++ b/devenv/init/.envrc
@@ -1,10 +1,12 @@
+# shellcheck shell=bash
+
 export DIRENV_WARN_TIMEOUT=20s
 
 eval "$(devenv direnvrc)"
 
 # `use devenv` supports the same options as the `devenv shell` command.
 #
-# To silence the output, use `--quiet`.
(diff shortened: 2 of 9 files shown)


.envrc
```diff
@@ -1,6 +1,5 @@
-#!/usr/bin/env bash
+# shellcheck shell=bash
```
domenkozar: What happens if you don't have shellcheck, I wonder if removing shebang works?

sandydoo: > What happens if you don't have shellcheck

This is a comment, so shellcheck isn't called. This is something shellcheck looks for.

> I wonder if removing shebang works?

We _used_ to have a shebang here that was added to specifically fix the shellcheck hook. It's not needed for direnv to run the file.

sandydoo: Looking back at this, maybe we should just add `#!/usr/bin/env bash` to all of our `.envrc` files?

There doesn't seem to be any harm in doing so and I don't like having a linter-specific comment in all the files we ship.

domenkozar: I think it's good to be explicit yeah

PR #2147, opened 2025-09-10
Don't run bun install with `--yarn` argument.

mdarocha:
Starting with Bun v1.2, Bun uses its own text lockfile, `bun.lock`.

See https://bun.sh/blog/bun-lock-text-lockfile

This means that the logic devenv used, which called `bun install --yarn`, is no longer necessary.

This PR intelligently picks either a new Bun install script that uses `bun install --save-text-lockfile`, or the existing script to preserve backwards compatibility.

### DIFF
--- a/src/modules/languages/javascript.nix
+++ b/src/modules/languages/javascript.nix
@@ -1,9 +1,15 @@
-{ pkgs, config, lib, ... }:
+{ pkgs
+, config
+, lib
+, ...
+}:
 
 let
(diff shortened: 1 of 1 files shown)


domenkozar:
Thanks for this! You don't have to do it, but ideally we'd start using https://devenv.sh/tasks/ for these kind of things as it already offers caching via `execIfModified`. Happy to merge it as is :)

Issue #2137, opened 2025-09-04
Profiles


domenkozar:
We currently have a [profiles](https://devenv.sh/guides/profiles/#defining-profiles) guide that implements a way to conditionalize devenv based on an enum.

Let's explore the design space where we could use the language to generate profiles, for example using different python versions:

```nix
{ pkgs, config, ... }: {
  languages.python = {
    enable = true;
    version = "3.15";
  }

  profiles."python-3.14" = {
    languages.python.version = "3.14";
  };
}
```

Then something like `devenv --profile python-3.14 shell` would join the profile config together with the defaults.

sandydoo:
[Released in v1.9](https://github.com/cachix/devenv/releases/tag/v1.9)

PR #2150, opened 2025-09-11
Profiles

domenkozar:
Fixes https://github.com/cachix/devenv/issues/2137

[docs](https://profiles.devenv.pages.dev/profiles/)


docs/profiles.md
```diff
@@ -0,0 +1,194 @@
...
+```nix
+{
+  name = "myproject";
+
+  packages = [ pkgs.git pkgs.curl ];
+  languages.nix.enable = true;
+
+  profiles = {
+    backend = {
+      extends = [ "base" ];
```
sandydoo: There's no `base` profile in this example. Made me think `base` was referring to the actual base config for a second.
```suggestion
```

docs/profiles.md
```diff
@@ -0,0 +1,194 @@
...
+      config = {
+        services.postgres.enable = true;
+        services.redis.enable = true;
+      };
+    };
+
+    frontend = {
+      extends = [ "base" ];
```
sandydoo:
```suggestion
```

src/modules/top-level.nix
```diff
@@ -252,6 +252,112 @@ in
...
+                  default = [ ];
+                  example = [ "base" "backend" ];
+                };
+                config = lib.mkOption {
+                  type = types.deferredModule;
+                  description = "Configuration to merge when this hostname matches.";
+                  default = { };
+                };
+              };
+            });
```
sandydoo: Lets reuse the type for the freeform, hostname, and username profiles.
The description on hostname/username should be enough.

src/modules/top-level.nix
```diff
@@ -252,6 +252,112 @@ in
         internal = true;
       };
     };
+
+    profiles = lib.mkOption {
```
sandydoo: This should go into `profiles.nix`

devenv/src/flake.tmpl.nix
```diff
@@ -101,6 +102,59 @@
...
+              collectProfileModules = profileName: profileType: visited:
+                let
+                  profileId = "${profileType}:${profileName}";
+                in
+                if builtins.elem profileId visited then
+                  throw "Circular dependency detected in profile extends: ${lib.concatStringsSep " -> " visited} -> ${profileId}"
+                else
+                  let
+                    profile =
+                      if profileType == "manual" then
```
sandydoo: So `profileType` is used just for matching? Would a `prefix` here simplify the matching logic? That `prefix` could then be passed down recursively for the `extends`.

src/modules/top-level.nix
domenkozar: Not sure why that's better than having it typed?

domenkozar:
merged!

Issue #2156, opened 2025-09-16
Custom service readiness probes


attilaolah:
Enabling e.g. TLS for Prometheus causes readiness probes to fail, since those are sent as HTTP requests. Using a custom probe would allow developers to implement e.g. a simple `curl` replacement that picks up the right CA bundle and certificates (for client auth).

This could also be used as a workaround for #2155.

attilaolah:
Actually, [the OTEL collector documentation](https://devenv.sh/reference/options/#servicesopentelemetry-collectorconfigfile) suggests the readiness probe can be set for the corresponding process.

I gave it a quick try but for me it doesn't appear to be working.

sandydoo:
> Actually, [the OTEL collector documentation](https://devenv.sh/reference/options/#servicesopentelemetry-collectorconfigfile) suggests the readiness probe can be set for the corresponding process.
> 
> I gave it a quick try but for me it doesn't appear to be working.

This is a bug. The type used for the options wasn't allowing any overrides (#2158)

Otherwise, this is generally possible (with the above fix). The discoverability, however, is terrible. It may be time to introduce common healthcheck options throughout the services.

domenkozar:
Shipped in 2.0 at https://devenv.sh/processes/#ready-probes

Issue #2157, opened 2025-09-16
`Error: NotPresent` with `processes.caddy.process-compose.is_elevated`


RafaelKr:
**Describe the bug**

I have a devenv environment for a caddy reverse proxy to be able to serve multiple devenv projects simultaneously. As I also need HTTPS for some testing, caddy needs to be run as root.

I just tried to update to the latest devenv and process-compose versions, now caddy doesn't start anymore, it shows `Error: NotPresent`.

**To reproduce**
My current devenv.yaml, nixpkgs-unstable for process-compose v1.75.1:
```yaml
# yaml-language-server: $schema=https://devenv.sh/devenv.schema.json
inputs:
  nixpkgs:
    url: github:NixOS/nixpkgs/nixpkgs-unstable
```

Simplified devenv.nix for minimal reproduction:
```nix
{
  ...
}:
{
  services.caddy = {
    enable = true;
  };

  processes = {
    caddy.process-compose = {
      # required to install its unique root certificate into the trust store
      is_elevated = true;
    };
    # process-compose immediately exits when there is no running process.
    # But if caddy crashes we still want to be able to see the logs.
    sleep.exec = "sleep 3600";
  };
}
```

<details>
<summary>Actual structure of my devenv.nix</summary>

```nix
{
  ...
}:
let
  vhosts = {
    "https://project1.localhost" =
      "unix//path/to/project1/.devenv/run/caddy.sock";

    "https://project2.localhost" =
      "unix//path/to/project2/.devenv/run/caddy.sock";
  };

  caddyHosts =
    let
      vhostMapFn = (
        address: upstreams: {
          extraConfig = ''
            # force self-signed certificates
            tls internal

            reverse_proxy * {
              to ${upstreams}
              transport http {
                proxy_protocol v2
              }
            }
          '';
        }
      );
    in
    builtins.mapAttrs vhostMapFn vhosts;
in
{
  services.caddy = {
    enable = true;
    config = ''
      {
        auto_https disable_redirects
      }
    '';
    virtualHosts = caddyHosts;
  };

  processes = {
    caddy.process-compose = {
      # required to install its unique root certificate into the trust store
      is_elevated = true;
    };
    # process-compose immediately exits when there is no running process.
    # But if caddy crashes we still want to be able to see the logs.
    sleep.exec = "sleep 3600";
  };
}
```
</details>

[...]

PR #2158, opened 2025-09-16
process-compose: fix overriding per-process options

sandydoo:
`type.attrs` does not merge options, making it impossible to override.

Partial fix for #2156.


RafaelKr:
Not sure, but maybe also related to #2157?

PR #2162, opened 2025-09-17
process-compose: check the socket for liveness before attaching

sandydoo:
We've had a lot of complaints about `process-compose` trying to attach to non-existent instances in recent months.
I think the issue is this commit: https://github.com/cachix/devenv/commit/053c79211a274098b5a48ccf2ff47873316e048b

`process-compose` will skip cleaning up the socket file if it exits abnormally.
While `process-compose up` is capable of restarting in this situation, `process-compose attach` is not and will continue trying to reconnect. Our simple file check is not a robust enough check.

This PR uses `socat` to test for all 3 conditions: the file exists, the file is a socket, the socket has a bound listener.

#### Related

We've also had feedback that it wasn't intuitive that `devenv up` would try to reconnect to another `process-compose` instance.
This is believable because the "Attaching to x" message is essentially impossible to see before TUI goes up. The TUI also doesn't seem to indicate that it's in any kind of "client mode" either.

### DIFF
--- a/src/modules/process-managers/process-compose.nix
+++ b/src/modules/process-managers/process-compose.nix
@@ -97,16 +97,18 @@ in
       # Ensure the log directory exists
       mkdir -p "${config.env.DEVENV_STATE}/process-compose"
 
-      ${if cfg.unixSocket.enable then ''
-      # Check if process-compose server is already running on the socket
-      if [ -S "${cfg.unixSocket.path}" ]; then
-        echo "Attaching to existing process-compose server at ${cfg.unixSocket.path}"
-        exec ${cfg.package}/bin/process-compose --unix-socket "${cfg.unixSocket.path}" attach "$@"
+      ${lib.optionalString cfg.unixSocket.enable ''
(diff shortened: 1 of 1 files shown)


domenkozar:
cc @mcdonc

domenkozar:
Note that it seems like (might) be fixed upstream: https://github.com/F1bonacc1/process-compose/issues/364

sandydoo:
> Note that it seems like (might) be fixed upstream: [F1bonacc1/process-compose#364](https://github.com/F1bonacc1/process-compose/issues/364)

Good fixes, but looks unrelated to the bulk of our reports.

domenkozar:
We should then open an issue upstream and link it in the comments - so hopefully it's fixed there :)

sandydoo:
> We should then open an issue upstream and link it in the comments - so hopefully it's fixed there :)

Hmm maybe we can get a "fail-fast" flag of some sort

domenkozar:
There should be a socket timeout for sure.

PR #2060, opened 2025-08-01
TUI mode

domenkozar:
Fixes https://github.com/cachix/devenv/issues/2042 
Fixes https://github.com/cachix/devenv/issues/1848
Fixes #1887
Fixes #2178

tui

- [x] duplicated `Evaluating`
- [x] fix terminal resizing
- [x] support tasks
- [x] fix graceful shutdown

PR #2164, opened 2025-09-18
tasks: fix broken restarts for task processes in process-compose

### LINKED ISSUE #2141: Restarting process stopped working after v1.8
**Describe the bug**

Restarting processes is broken in newer versions than 1.8.0, in version 1.8.0 it works as intended.

There was a problem in v1.8.1 that caused processes to not get properly killed but restarting them doesn't work either.
I don't know if this is caused by the fix or if the same bug that broke the shutdown, broke the restart behaviour as well.
https://github.com/cachix/devenv/pull/2045

**To reproduce**

1. Press `CTRL+R` on a process in the process-compose TUI.
2. Said process only stops but doesn't restart.

**Version**

Flakes with devenv version >1.8.0

sandydoo:
- Properly wait on child processes to ensure they are reaped
- Remove manual timeout of process shutdown
- Perform a textbook unix signal shutdown by restoring the default handler and re-raising the signal. If we're going to manage processes, we should do it properly.

Fixes #2141.

### DIFF
--- a/devenv-tasks/src/main.rs
+++ b/devenv-tasks/src/main.rs
@@ -62,15 +62,18 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
                 run_mode: mode,
             };
 
-            // Create shared signal handler
+            // Create a global signal handler
             let signal_handler = SignalHandler::start();
-            let cancellation_token = signal_handler.cancellation_token();
 
             let mut tasks_ui = TasksUi::builder(config, verbosity)
-                .with_cancellation_token(cancellation_token)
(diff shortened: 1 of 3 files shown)


domenkozar:
This will be painful to resolve with #2060 that removes this logic completely :(

sandydoo:
> This will be painful to resolve with #2060 that removes this logic completely :(

It'll be fine. The child shutdown will merge without issues.

As for the top-level signal handling, I think #2060 removed a bit too much. I only see ctrl+c handling left, which isn't enough.  I don't see any issue integrating the tokio graceful stuff with this.

PR #2166, opened 2025-09-18
starship: add inline configuration option

Yethal:
Main use case for this is being able to merge in configuration from imports and profiles


domenkozar:
Can you modify existing example to show both can be set? We prefer that over having multiple examples. Thank you!

Yethal:
Done

PR #2200, opened 2025-10-05
Improve process management with conflict detection and error handling

### LINKED ISSUE #2198: `devenv up -d` overwrites pid file if processes are already running
**Describe the bug**
When running `devenv up -d` consecutively without stopping previously started processes, `devenv` overwrites the pid file stored in `.devenv/processes.pid`, which in turn will break `devenv processes down`.

**To reproduce**
1. Create default configuration with `devenv init`
2. Configure a process to start, e.g. with `services.mysql.enable = true;`
3. Run `devenv up -d`
4. Run `devenv up -d`
5. Run `devenv processes down`

**Expected result:**
`devenv` stops the started processes.

**Actual result:**
```
Stopping process with PID 64353
Error:   × Process with PID 64353 not found.
```
And the processes are still running, because the PID file was overwritten by the second invocation of `devenv up -d`.

[Example files](https://gist.github.com/chibidev/1c2cd9f35f211014013bd6e4738298bb)

**Version**
devenv 1.9.0 (aarch64-darwin)

domenkozar:
Prevents multiple process-compose instances from running simultaneously by checking PID file before starting. Handles stale PID files gracefully and replaces .expect() calls with proper error propagation using miette diagnostics.

Fixes #2198

### DIFF
--- a/devenv/src/devenv.rs
+++ b/devenv/src/devenv.rs
@@ -1082,19 +1082,67 @@ impl Devenv {
                     .await?
             };
 
+            // Check if processes are already running (prevents conflicts in both detached and non-detached modes)
+            let processes_pid_path = self.processes_pid();
+            if processes_pid_path.exists() {
+                let existing_pid_str = fs::read_to_string(&processes_pid_path)
+                    .await
+                    .into_diagnostic()
+                    .wrap_err("Failed to read processes.pid file")?;
+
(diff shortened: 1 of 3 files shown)


domenkozar:
This is trickier to solve in general case because we also need to support workflow when `devenv up -d` creates a detached mode and then `devenv up` should just attach.

domenkozar:
Let's do this only in detached mode: https://github.com/cachix/devenv/pull/2203

PR #2204, opened 2025-10-07
Make it posible to override pre-commit binary

leiserfg:
I replaced the name of the binary 'cause I don't feel comfortable naming `prek` as `pre-commit` let me know if you don't like it.


domenkozar:
We should probably get the name of the executable and then link that one instead of suffixing `-bin`

leiserfg:
Not sure if this one works, if not, can you point me to how to run it?

domenkozar:
Looking to set it as default for devenv in https://github.com/cachix/devenv/pull/2304

PR #2208, opened 2025-10-08
ci: allow overriding the nixpkgs used in tests

sandydoo:
This is to allow reusing these workflows in devenv-nixpkgs.


domenkozar:
This is for testing nixpkgs bumps?

sandydoo:
> his is for testing nixpkgs bumps?

Yeah, I want to dedupe the tests harness and call this workflow from devenv-nixpkgs.

PR #2223, opened 2025-10-15
feat: add and document devenv usage with standard nix

### LINKED ISSUE #2126: Guide for using with npins
Please add a possibility for using devenv shells with npins.
There are 2 supported ways of using devenv that I know of:
1. using the devenv cli calling `devenv shell` which supports specifying inputs with devenv.yaml
2. using flakes calling `nix develop --impure` after defining a devenv dev shell in the flake.

The issue with the first method is that I can't use the same input specification across my NixOS and home manager modules, it's only for devenv. The issue with the second method is that is requires using flakes, which is non-standard nix.

It would be great to have an officially supported way to use devenv with npins, something like:

```nix
# default.nix
let
  sources = import ./npins;
  pkgs = import sources.nixpkgs { ... };
  devenv = import sources.devenv { ... }; # maybe inject nixpkgs here?
in
{
  # activate with `nix-shell -A devshell`
  devshell = devenv.mkShell {
    # or somehow inject pkgs here
    imports = [ ./devenv.nix ];
  };
}
```

If this possibility already exists please document it. If the documentation already exists I apologise and kindly ask that you point me to where I can find it.

balintbarna:
Fixes #2126

This PR add a thin wrapper to be used with standard nix files, such as a common project level `default.nix` file, to define devenv shells for the project. There is also a concise guide added to the documentaiton describing the usage with this new wrapper and npins to pin dependencies.

### DIFF
--- a/docs/src/guides/using-with-standard-nix.md
+++ b/docs/src/guides/using-with-standard-nix.md
@@ -1,0 +1,89 @@
+# Using devenv with standard Nix
+
+[Nix Flakes](https://wiki.nixos.org/wiki/Flakes) are not standard nix, and for those who prefer not to,
+or can't use them but still want to pin and share dependencies across multiple targets, there is a way.
+Without flakes or the devenv CLI, we can download inputs using built-in functions like fetchGit or fetchTarball,
+or using popular dependency pinning tools like [niv](https://github.com/nmattia/niv) or [npins](https://github.com/andir/npins).
+Then we can define devenv shells as output attributes like anything else in Nix.
+
+!!! note "For those new to devenv and Nix"
+    If you're new to both devenv and Nix, starting with the standard devenv CLI approach will provide the smoothest experience. [Getting started with devenv.](../getting-started.md)
+
+You can integrate the devenv module system (languages, processes, services, etc.) into a Nix module as an attribute.
+This allows using devenv without flakes or the devenv CLI and replace standard nix shells in a painless way within
+your existing nix-based development environments.
+
+Creating a devenv shell via the non-flake wrapper is not a first-class option and might have some limitation compared to the other methods.
+It is only recommended to experienced Nix users who prefer to not use flakes.
+
+
+## Getting started
+
+In this guide we will show how to write a basic `default.nix` using the built-in
+fetchTarball to fetch sources from GitHub and a devenv shell definition.
+
+Create a default.nix file in the project root with the following content:
+
+```nix title="default.nix"
+let
+  nixpkgs = fetchTarball "https://github.com/cachix/devenv-nixpkgs/archive/rolling.tar.gz";
+  pkgs = import nixpkgs { };
+  devenv-src = fetchTarball "https://github.com/cachix/devenv/archive/main.tar.gz";
+  devenv = (import devenv-src).lib.mkStandardShell;
+in
+{
+  shell = devenv {
+    inherit pkgs;
+    modules = [ ./devenv.nix ];
+  };
+}
+```
+
+Finally, create a standard devenv configuration in `devenv.nix` [as normal](https://devenv.sh/basics/).
+
+
+### Entering the shell
+
+Create and enter the `devenv` shell with:
+
+```console
+nix-shell
+```
+
+
+### Further considerations
+
+The created `default.nix` file is standard Nix.
+You can freely define multiple attributes, such as more shells, packages, NixOSand Home Manager configurations,
+among other things, as well as add more input sources as necessary.
+
+The non-flake warpper is a thin wrapper created around the flake based shell function, which was written with flakes in mind.
+In some cases, such as Rust channel configurations, devenv requires extra inputs like the rust-overlay flake.
+If that input is not provided, devenv will print error messages with some instructions meant for a flake-based setup.
+This scenario can be handled the following way with a non-flake setup:
+
+```nix title="default.nix"
+let
+  nixpkgs = fetchTarball "https://github.com/cachix/devenv-nixpkgs/archive/rolling.tar.gz";
+  pkgs = import nixpkgs { };
+  devenv-src = fetchTarball "https://github.com/cachix/devenv/archive/main.tar.gz";
+  devenv = (import devenv-src).lib.mkStandardShell;
+  flake-compat = import (fetchTarball "https://github.com/NixOS/flake-compat/archive/master.tar.gz");
+  rust-overlay-src = fetchTarball "https://github.com/oxalica/rust-overlay/archive/master.tar.gz";
+in
+{
+  shell = devenv {
+    inherit pkgs;
+    modules = [ ./devenv.nix ];
+    inputs = {
+      nixpkgs = pkgs;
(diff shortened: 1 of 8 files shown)


src/non-flake-wrapper.nix
```diff
@@ -0,0 +1,14 @@
+{ mkShell }:
```
sandydoo: 1. Our flake exports a `mkShell` like this.
2. I don't particularly like the renamed arguments: `nixpkgs` and `configuration`. I would stick with the API we already expose (see 1.).

docs/src/guides/using-with-standard-nix.md
```diff
@@ -0,0 +1,57 @@
...
+```console
+nix-shell -p npins
+npins init
+npins add github cachix devenv 
+```
+
+Then create a default.nix file in the project root with the following content:
+
+```nix
+# default.nix
```
sandydoo:
```suggestion
```nix title="default.nix"
# default.nix
```

sandydoo (changes requested):
Thanks for writing this up.

Would you also be able to add a test for this?

In tests:
1. Create a folder with the `default.nix`
2. `.test-config.yml`: `use_shell: false`
3. `.test.sh`: Run commands to launch the shell. If possible with npins, don't commit the lock.

balintbarna:
Thany you @sandydoo for the comments and guidance. I pushed a new commit that tries to address some your concerns and also added some more documentation on how to handle more complex scenarios with this wrapper.

I still have to add the tests.

docs/src/guides/using-with-standard-nix.md
balintbarna: This is added in the last commit

src/non-flake-wrapper.nix
balintbarna: 1. the mkShell that is exported in the flake is not autmatically available here because there's no `self`. Therefore this function accepts it as an argument and the mkShell you referenced here is passed in in flake.nix [here](https://github.com/cachix/devenv/pull/2223/files#diff-206b9ce276ab5971a2489d75eb1b12999d4bf3843b7988cbe8d687cfde61dea0R347)
2. I now updated those arguments to align with the other mkShell function (pkgs, inputs, modules)

balintbarna:
@sandydoo I now pushed the tests as well, let me know if this is how you wanted it. I ran the test bash files locally and that works well but I'm not sure how your CI is structured.

balintbarna:
I'm having a bit of difficulty understanding what's wrong with the tests.

sandydoo:
I've fixed the tests here. I dropped the `advanced` test as it doesn't add much coverage for us and just kept the `simple` one (renamed to `npins`).

sandydoo (approved):
Now that we have a test for this, I'm happy. Not entirely sold on the naming of `nonFlakeMkShell`, but tentatively approving.

Thoughts, @domenkozar?

tests/non-flake-simple/default.nix
```diff
@@ -1,7 +1,7 @@
 let
   sources = import ./npins;
   pkgs = import sources.nixpkgs { };
-  devenv = (import ../..).lib.nonFlakeMkShell ./.;
+  devenv = (import sources.devenv).lib.nonFlakeMkShell ./.;
```
balintbarna: @sandydoo I think this is incorrect and will lead to issues.
Above in your changes you called `npins add github cachix devenv` which will add the master branch version of devenv. This test will never run successfully on this branch before it is merged into master because the version on master doesn't have the feature this depends on yet. It has to reference the devenv flake in the same repo, not downloaded from GitHub.

balintbarna:
> Now that we have a test for this, I'm happy. Not entirely sold on the naming of `nonFlakeMkShell`, but tentatively approving.

I also dislike the name, feel free to change it. mkShell is already taken unfortunately. Could be mkStandardShell for example.

tests/non-flake-simple/default.nix
sandydoo: These tests are designed to be run with `devenv-run-tests`. Each test is run from a fresh temporary directory, so the relative import won't work.

I've modified `devenv-run-tests` to override the devenv source using `NPINS_OVERRIDE_devenv`. We do something similar for the other tests.

balintbarna:
@sandydoo should we try to get this merged in in the next couple days?

domenkozar:
I'm against having npins guide part of devenv, because don't want to support that (mostly because already supporting Nix inputs is a lot of work).

We can expose things on devenv side and the guide should be upstreamed to npins instead, to claim it has devenv support and how to use it.

Does this make sense?

balintbarna:
@domenkozar Thanks for your input. I don't think it makes sense to add anything to npins regarding this change. This PR adds a function to devenv and I think that having the documentation here on how to use this new function is very useful and seems to me like the most reasonable place to have it.

I used npins in the guide because I think it provides the best user experience but this function should work with any standard nix workflow, using npins or niv or fetchTarball, it shouldn't matter.

balintbarna:
Perhaps we could change the documentation and the tests to use fetchgit instead of npins, is that something you would prefer?

domenkozar:
> @domenkozar Thanks for your input. I don't think it makes sense to add anything to npins regarding this change. This PR adds a function to devenv and I think that having the documentation here on how to use this new function is very useful and seems to me like the most reasonable place to have it.
> 
> I used npins in the guide because I think it provides the best user experience but this function should work with any standard nix workflow, using npins or niv or fetchTarball, it shouldn't matter.

Imagine a scenario where npins makes a breaking change, then documentation for npins and devenv needs to be tested and upated. All of that should be handled by npins, that's why I prefer that maintenance burden is not transfered to us because we don't have the capacity to handle it.

We should definitely expose the bits on the devenv side for npins to integrate well and I'm also fine having a guide link to npins+devenv guide as a link.

balintbarna:
This feature of devenv doesn't really have anything to do with npins so I won't submit any changes to npins.

I will update the documentation in this PR to not use npins but instead use built in nix functions like fetchGit or fetchTarball. That way there won't be any npins related maintenance burden on you.

balintbarna:
I updated the tests and documentation to work with the built-in Nix function fetchTarball. No more npins, except it's mentioned as one of the tools that can be used in the documentation.

I'm not sure how to update the test override so that it would take the current branch of devenv instead of main. @sandydoo  can you help with that?

balintbarna:
Hi everyone!

I understand this is not a high priority ticket for your team, but for me it would be really nice to not have to use my fork of devenv anymore. I removed all the npins related documentation/guide and replaced it with built-in nix functions.

Could we take a final look and push and see if we can get it merged in? I'm happy to make any final changes if you have any further concerns. Let me know what you think.

@domenkozar @sandydoo

balintbarna:
I simplified the shell function by passing in `self` directly, and updated the test configuration to use the DEVENV_REPO environment variable.

balintbarna:
@sandydoo I rebased the branch on upstream main, would love to get it merged. Anything else needed to be changed?

balintbarna:
Just rebased, still working without any issues

PR #2227, opened 2025-10-20
devenv: improve jobs/cores defaults

### LINKED ISSUE #2213: Global setting for `--cores`  and/or `--max-jobs`
Is there a way to globally configure all `devenv` shells to use all cores available? Currently, if I run `devenv init` I have to kill the process, then update the `.envrc` to contain `--cores "$(nproc)"` and it would nice if there was a way to configure this by default with an environment variable like `export DEVENV_CORES="$(nproc)"`, and similarly with `--max-jobs`, `export DEVENV_MAX_JOBS=12`

sandydoo:
- Improves the default values for `--cores` and `--max-jobs`, trying to strike a balance between the two. Previously, `cores` was capped to `2`, making any builds unnecessarily slow.
- Adds the ability to configure `--cores` and `--max-jobs` using the environment variables `$DEVENV_CORES` and `DEVENV_MAX_JOBS`, respectively.

Fixes #2213.

### DIFF
--- a/devenv/src/cli.rs
+++ b/devenv/src/cli.rs
@@ -63,7 +63,9 @@ pub struct GlobalOptions {
     pub log_format: LogFormat,
 
     #[arg(short = 'j', long,
-        global = true, help = "Maximum number of Nix builds at any time.",
+        global = true,
+        env = "DEVENV_MAX_JOBS",
+        help = "Maximum number of Nix builds to run concurrently.",
         default_value_t = NixBuildDefaults::compute().max_jobs)]
     pub max_jobs: u8,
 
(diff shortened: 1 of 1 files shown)


devenv/src/cli.rs
```diff
@@ -63,15 +63,19 @@ pub struct GlobalOptions {
     pub log_format: LogFormat,
 
     #[arg(short = 'j', long,
-        global = true, help = "Maximum number of Nix builds at any time.",
-        default_value_t = max_jobs())]
+        global = true,
+        env = "DEVENV_MAX_JOBS",
```
domenkozar, line 67: I'm not sure we should allow env variables here.

We should think about a global user config, because this is about the machine we're running on.

sandydoo: I was thinking of adding env var support to a bunch of our settings, [uv-style](https://docs.astral.sh/uv/reference/environment).

> We should think about a global user config, because this is about the machine we're running on.

Yes, we need a global config too. I think this is complementary?

Issue #1694, opened 2025-01-28
dotenv integration leaks secrets into the nix store


pkel:
**Describe the bug**

The `.env` file may contain secrets. Its contents must remain confidential. [source](https://www.dotenv.org/docs/security/env)
According to nix philosophy the nix store is not confidential. [source](https://nixos.wiki/wiki/Comparison_of_secret_managing_schemes)
The devenv/dotenv integration copies the contents of the .env file into the nix store.
As a result, the secrets in `.env` do not remain confidential.

Related:
- #775

**To reproduce**

Create `.env` file with content `SECRET=batteryhorsestaples`.
Set `dotenv.enable` in `devenv.nix`, build the shell.
Search (and find) `batteryhorsestaples` in `/nix`.

**Version**
 
devenv 1.3.1 (x86_64-linux)

itpropro:
Any updates on this? Currently, I couldn’t use devenv in any development environment with such a security flaw.

domenkozar:
We have addressed this by releaseing secretspec: https://devenv.sh/blog/2025/07/21/announcing-secretspec-declarative-secrets-management/

pkel:
Imo you should (and should have) put a warning on https://devenv.sh/integrations/dotenv/.

domenkozar:
Happy to accept a PR for someone to document this one!

pkel:
> Happy to accept a PR for someone to document this one!

Happy to address your comments over there: #2228

domenkozar:
Note that a proper fix would be to move the parsing of `.env` files into Rust code and then populating Nix variables as inputs.

I suggest moving to https://devenv.sh/integrations/secretspec/ as long-term solution.

domenkozar:
We've added support for path secrets in secretspec 0.4.0 to further help with this: https://devenv.sh/blog/2025/11/26/devenv-111-module-changelogs-and-secretspec-040/

domenkozar:
This is now documented, use secretspec.dev please.

PR #2228, opened 2025-10-20
Document .env secret leak to nix store

pkel:
This PR adds documentation for the security issue with the `.env` integration, as described in #1694.


pkel:
Wow, that was fast. :rocket: Thanks

PR #2230, opened 2025-10-20
tasks: fix process output being red

### LINKED ISSUE #2211: devenv makes all process output red
**Describe the bug**

Define some processes, disable TUI, and invoke `devenv up`.

Image

**To reproduce**

Minimal reproducible example pushed to this repo: https://github.com/jcf/devenv-red

**Version**

``` console
$ devenv version
devenv 1.9.1 (aarch64-darwin)
```

domenkozar:
Fixes #2211 by changing process output from stderr (eprintln!) to stdout (println!).

Previously, all process output was printed to stderr which appears red in most terminals. Now process output goes to stdout while maintaining the verbose mode behavior with task name prefixes.

### DIFF
--- a/devenv-tasks/src/task_state.rs
+++ b/devenv-tasks/src/task_state.rs
@@ -366,8 +366,10 @@ impl TaskState {
                 result = stdout_reader.next_line(), if !stdout_closed => {
                     match result {
                         Ok(Some(line)) => {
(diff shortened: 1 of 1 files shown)


sandydoo:
I can replicate this with `process.managers.process-compose.tui.enable = false;`. The question I have is: why does process-compose print all stderr output in red when in non-tui mode?

PR #2234, opened 2025-10-22
python: support `withPackages` with venvs and prepare for upstream patch

### LINKED ISSUE #1218: `languages.python.package = pkgs.python310.withPackages (…)` does not work
Thank you for developing and maintaining this awesome project! :slightly_smiling_face:

**Describe the bug**

I'm trying to specify Python packages by

```
languages.python = {
  enable = true;
  package =
    pkgs.python310.withPackages (ps: [
      ps.xlrd
    ]);
};
```

but this does not work, i.e.

```
nix develop --impure --command python -c 'import xlrd; print("yeah!")'
```

yields

```
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'xlrd'

```

Interestingly, as also described in [this comment](https://github.com/cachix/devenv/issues/601#issuecomment-2014621906), instead specifying

```
packages = with pkgs.python310Packages; [ xlrd ];
```

*does* work but does not seem to be preferable due to the uncoupling of Python packages and interpreter. Or maybe I misunderstood something? Of course I could use `let` and friends to extract `python310` somehow but that seems clunky.

**To reproduce**

https://gist.github.com/dpaetzel/303644fc1aa7863a4d0194c9f205b4b1

**Version**

I'm using Flakes.

### LINKED ISSUE #1486: Selenium for python is broken on devenv
**Describe the bug**
The selenium python package does not work when installed using devenv. Installing directly with nix works.

**To reproduce**

https://gist.github.com/theoahfeldt/cfca82103579841cc4f5bca5b1623210

**Version**
 
devenv 1.2.0 (x86_64-linux)

### LINKED ISSUE #1792: Python: Matplotlib & Tkinter
I know that Python is a great source of headaches, because of system deps (as in #773 and related), and also because of graphic tools (as in #1587).

However, while using `matplotlib` requires both, in principle it could run displaying plots in a window, based on the built-in [`tkinter`](https://docs.python.org/3/library/tkinter.html) module.
Now, even `tkinter` has its own complications, including not being distributed by default with `nixpkgs.python3X` (cf. https://github.com/NixOS/nixpkgs/issues/1819, https://github.com/NixOS/nixpkgs/issues/7134, https://github.com/NixOS/nixpkgs/issues/19255, https://github.com/NixOS/nixpkgs/pull/19309, and possibly more...). As it is written clearly in the [Nixpkgs manual](https://nixos.org/manual/nixpkgs/stable/#missing-tkinter-module-standard-library)
> #### Missing `tkinter` module standard library {#missing-tkinter-module-standard-library}
>
> To reduce closure size the `Tkinter`/`tkinter` is available as a separate package, `pythonPackages.tkinter`.

However, it comes as a dependency of the `matplotlib` package in `nixpkgs`, and thus it is brought in when the package is requested among Python packages.
But this is not happening under `devenv`, and I'm not sure why.

Indeed, defining two development shells as follows:

<details><summary>Minimal flake with vanilla and Devenv shells</summary>
<p>

```nix
{
  inputs = {
    nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
    devenv = {
      url = "github:cachix/devenv";
      inputs.nixpkgs.follows = "nixpkgs";
    };
  };

  outputs = {
    self,
    nixpkgs,
    devenv,
    ...
  } @ inputs: {
    devShells = (
      nixpkgs.lib.genAttrs ["x86_64-linux"]
      (system: let
      in {
        default = pkgs.mkShell {
          packages = with pkgs; [
            (python3.withPackages (py:
                with py; [matplotlib numpy ipython]))
          ];
        };

        devenv = devenv.lib.mkShell {
          inherit inputs pkgs;
          modules = [
            ({pkgs, ...}: {
              languages.python = {
                enable = true;
                package =
                  pkgs.python3.withPackages (py:
                    with py; [matplotlib numpy ipython]);
              };
            })
          ];
        };
      })
    );
  };
}
```

</p>
</details> 

results in the `default` one succeeding to import `tkinter` (and then makes `matplotlib` backend accessible), while the `devenv` alternative is failing

<details><summary>Importing & failing</summary>
<p>

```py
$ ipython
Python 3.12.9 (main, Feb  4 2025, 14:38:38) [GCC 14.2.1 20241116]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.32.0 -- An enhanced Interactive Python. Type '?' for help.

[ins] In [1]: import tkinter

[ins] In [2]:
Do you really want to exit ([y]/n)?

$
exit

❯ : nix develop .#devenv --impure
Running tasks     devenv:enterShell
Succeeded         devenv:enterShell 8ms
1 Succeeded                         8.86ms

(devenv)
$ ipython
ipython: command not found
(devenv)
$ python
Python 3.12.9 (main, Feb  4 2025, 14:38:38) [GCC 14.2.1 20241116] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tkinter
  File "<stdin>", line 1, in <module>
  File "/nix/store/iwksr4mrmp6gz4q8w1yzzrhhxqw1fmjx-python3-3.12.9-env-env/lib/python3.12/tkinter/__init__.py", line 38, in <module>
    import _tkinter # If this fails your Python may not be configured for Tk
    ^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named '_tkinter'
>>>
```

</p>
</details> 

---

In full honesty, my personal goal would be to make it working through a Poetry installation (which should also handle `matplotlib`).
But, as a starting point, I'd be happy to understand what is going wrong in the previous example...

### LINKED ISSUE #1807: Python native "libraries" option not working with Poetry
I'm using devenv with poetry for a python project. I use python-magic, which depends on the libmagic library. python-magic is installed using poetry (pyproject.toml), so I manually added the `file` package to `libraries`. This installs the `file` binary as well as the library, however the magic package (using ctypes util find library on unix) cannot find the library.

```
  languages.python = {
    enable = true;
    version = "3.13.1";
    poetry = {
      enable = true;
      activate.enable = true;
      install = {
        enable = true;
        groups = [ "dev" ];
      };
      package = pkgs-unstable.poetry;
    };
    libraries = [
      # for libmagic
      pkgs.file
    ];
  };
```

Opening an interpreter in the devenv shell and printing os.environ["LD_LIBRARY_PATH"] only returns `.devenv/profile/lib` and `...-gcc-13.3.0-lib/lib`. However `builtins.trace (lib.makeLibraryPath [ pkgs.file ])` gives the expected result path.

If I manually add this to the env before starting python, everything works. If I remove the poetry module, everything works as well (apart from poetry...), so like this:

```
  languages.python = {
    enable = true;
    version = "3.13.1";
    libraries = [
      # for libmagic
      pkgs.file
    ];
  };
```

Version: devenv 1.4.1 (x86_64-linux)

**TLDR: I think the setting of the LD_LIBRARY_PATH is not working when using poetry.**

sandydoo:
This prepares us for the arrival of several upstream patches (including https://github.com/NixOS/nixpkgs/pull/442540) that will make it possible to drop the python wrapper.

- `withPackages` are now propagated through the wrapper
- Adds a `patches.buildEnv.enable` option to toggle the wrapper

Fixes #1807.
Fixes #1792.
Fixes #1486. (possibly earlier)
Fixes #1218.

### DIFF
--- a/docs/src/supported-languages/python.md
+++ b/docs/src/supported-languages/python.md
@@ -135,13 +135,13 @@ boolean
 
 Whether to apply fixes to Python’s ` buildEnv ` for correct runtime initialization:
 
- - Executables use ` --inherit-argv0 ` to ensure Python initializes with correct ` sys.prefix ` and ` sys.base_prefix `
+ - Executables use ` --inherit-argv0 ` and ` --resolve-argv0 ` to ensure Python initializes with correct ` sys.prefix ` and ` sys.base_prefix `
  - Python package scripts are unwrapped to invoke the environment’s interpreter directly
 
 Without these fixes, venvs cannot access environment packages via ` --system-site-packages `\.
 
-Enabled by default for nixpkgs versions prior to 25\.11\.
-Newer nixpkgs releases include upstream fixes that make this patch obsolete\.
+Enabled by default\.
+Newer nixpkgs releases may include upstream fixes that make this patch obsolete\.
 
 
 


src/modules/languages/python.nix
```diff
@@ -343,6 +405,19 @@ in
         default = false;
         description = "Whether `pip install` should avoid outputting messages during devenv initialisation.";
       };
+      wrapper.enable = lib.mkOption {
+        type = lib.types.bool;
+        default = legacyWrapperEnabled;
+        description = ''
+          Whether to enable a wrapper that allows Python to import packages from the virtual environment.
```
gabyx: question: I dont understand here the following:

The legacy wrapper never actually made importing packages from virtual environment possible, or did it exactly that?

Was that wrapper exactly for this purpose because otherwise you could not import any `python -c "import pyyaml` where `pyyaml` was installed with `uv add pyyaml` for example?

src/modules/languages/python/default.nix
```diff
@@ -582,7 +657,7 @@ in
     };
 
     enterShell = ''
-      export PYTHONPATH="$DEVENV_PROFILE/${package.sitePackages}''${PYTHONPATH:+:$PYTHONPATH}"
+      export PYTHONPATH="$DEVENV_PROFILE/${cfg.package.sitePackages}''${PYTHONPATH:+:$PYTHONPATH}"
```
gabyx, line 660: question: Why do we even need to set `PYTHONPATH`. We should never tinker with that if possible?

tests/python-with-packages/devenv.nix
```diff
@@ -0,0 +1,19 @@
+{ pkgs, ... }:
+{
+  languages.python = {
+    enable = true;
+    package = pkgs.python3.withPackages (ps: [
```
gabyx, line 7: awesome!

src/modules/languages/python/default.nix
gabyx: As I see this resolves to `DEVENV_PROFILE/lib/python.../site-packages`. for what is that path and what does it contain?

src/modules/languages/python.nix
```diff
@@ -280,6 +292,56 @@ in
...
+        in
+        # Apply the venv support wrapper if enabled.
+        if cfg.venv.wrapper.enable then
+          lib.pipe drv [
+            (
+              drv:
+              drv.overrideAttrs (
+                prevAttrs:
+                let
+                  buildEnv = pkgs.callPackage ../../python-wrapper.nix {
```
gabyx: suggestion: maybe call this file `python-wrapper-legacy.nix` now

src/modules/languages/python.nix
```diff
@@ -280,6 +292,56 @@ in
...
+                (lib.makeLibraryPath libraries)
+              ];
+            });
+        in
+        # Apply the venv support wrapper if enabled.
+        if cfg.venv.wrapper.enable then
+          lib.pipe drv [
+            (
+              drv:
+              drv.overrideAttrs (
```
gabyx: suggestion: take this lambda to the `let` block above, then the pipe becomes cleaner

gabyx (changes requested):

src/modules/languages/python/default.nix
sandydoo: No clue. Maybe an old attempt at doing what the wrapper does?

sandydoo: https://github.com/cachix/devenv/commit/f151daf3a821e8f6cfce0494d670bb5ef5b0b8d8#diff-f5bb89f776eda1a758db67c6c75b19ff1665b2f10998d17d1d98479835fc9e2eR482

This is from a large tasks PR. Let's try removing it.

sandydoo: This `PYTHONPATH` is critical in making site packages  (`withPackages`) available to venvs created with `venv.enable`.

src/modules/languages/python.nix
sandydoo: Yes, I completely confused what the point of the wrapper was.

src/modules/languages/python/default.nix
gabyx: ah weird: ok, so its an empty folder which gets filled by venv.enable I guess

sandydoo:
Added a test to verify `sys.base_prefix` and the interaction between `withPackages` and `venv.enable`.

We now also patch `buildEnv` in a way that avoids nesting envs.

This PR still needs to be tested against the upstream nixpkgs PR. Once it's out of staging, we can verify this easily without rebuilding the universe.

tests/python-with-packages/devenv.nix
```diff
@@ -0,0 +1,51 @@
...
+      ps.matplotlib
+      ps.numpy
+      ps.ipython
+      ps.tkinter
+    ]);
+    patches.buildEnv.enable = true;
+    venv.enable = true;
```
gabyx: suggestion: could we add a `uv` test as well? with `uv.enable = true` and `uv.sync = true` or so

`pyproject.toml`

```yaml
[project]
name = "mymodule"
description = "test"
version = "0.0.1"

dependencies = [
    "request",
    "pytest",
]
```

tests/python-with-packages/devenv.nix
```diff
@@ -0,0 +1,51 @@
...
+        python <<'EOF'
+    import sys
+    import os
+
+    print("sys.base_prefix:", sys.base_prefix)
+    print("sys.executable:", sys.executable)
+
+    # Check that sys.base_prefix points to the -env buildEnv, not the bare interpreter
+    assert "-env" in sys.base_prefix, \
+        f"sys.base_prefix ({sys.base_prefix}) should point to python-env with packages, not bare interpreter"
```
gabyx: Awesome!

gabyx (changes requested):

domenkozar:
Shall we merge this?

gabyx:
suggestion: we need another uv test as uv is the predominant tool
in the future

gabyx:
but should not hold it back!

tests/python-with-packages/devenv.nix
sandydoo: Done!

PR #2235, opened 2025-10-22
Make Crystal and Shards packages configurable

straight-shoota:
This allows overriding the package version.
For example:
```nix
languages.crystal = {
  enable = true;
  package = pkgs.crystal_1_16
};
```

### DIFF
--- a/src/modules/languages/crystal.nix
+++ b/src/modules/languages/crystal.nix
@@ -14,11 +14,17 @@ in
       description = "The Crystal package to use.";
     };
 
-    shards_package = lib.mkOption {
-      type = lib.types.package;
-      default = pkgs.shards;
-      defaultText = lib.literalExpression "pkgs.shards";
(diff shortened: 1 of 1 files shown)


src/modules/languages/crystal.nix
```diff
@@ -6,15 +6,29 @@ in
...
     enable = lib.mkEnableOption "Enable tools for Crystal development.";
+
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.crystal;
+      defaultText = lib.literalExpression "pkgs.crystal";
+      description = "The Crystal package to use.";
+    };
+
+    shards_package = lib.mkOption {
```
domenkozar: let's use shards.package

straight-shoota: I pushed a change to do that, but I'm not entirely sure if that's what you meant. Any recommendation is welcome.

PR #1679, opened 2025-01-21
feature/treefmt - integrate treefmt-nix

jmuchovej:
attempting to continue the work from #1317.

PR #2238, opened 2025-10-23
feat: add treefmt integration

tarc:
This is a continuation of #1679.

Improvements:
- `treefmt`'s input is handled with `config.lib.getInput` in order to make it optional.
- The main option is defined as `formatting` - this way, `treefmt` is defined therein (as `formatting.treefmt`).
- Example and test are added.


domenkozar:
Docs should be under `integrations` and same for the module name, we're having a flat namespace.

tarc:
> Docs should be under `integrations` and same for the module name, we're having a flat namespace.

I get rid of the `formatting` top level - is that what you meant, @domenkozar? (sorry, I'm pretty new to nix module hacking and terminology).

PR #2239, opened 2025-10-23
devenv: add ability to output JSON traces

sandydoo:


devenv/src/log.rs
```diff
@@ -51,14 +56,32 @@ pub enum LogFormat {
...
     TracingPretty,
+    /// A JSON log format used for machine consumption.
+    TracingJson,
+}
+
+macro_rules! json_export_layer {
+    ($export_file:expr) => {
+        $export_file.map(|file| {
+            tracing_subscriber::fmt::layer()
+                .with_writer(file)
```
domenkozar, line 67: Maybe it should go to stderr by default? Then --log-format-file overrides that.

domenkozar (changes requested):

devenv/src/log.rs
sandydoo: I'd like to have an option to generate traces separately from what the UI is doing. This'll be convenient once TUI ships.

When `--trace-export-file` is set, we output JSON traces to the file, regardless of what we print to stderr.
If we need anything other than JSON, we can add an option here.
`--trace-export-file` should probably become `--trace-export-to` and a URI for our later use-cases.

PR #2247, opened 2025-10-28
treefmt: make config option definition conditional

sandydoo:
Since treefmt is an optional input, we need to make the submodule definition optional as well.

This fixes tools that traverse options, like doc gen and `devenv search`.

### DIFF
--- a/src/modules/integrations/treefmt.nix
+++ b/src/modules/integrations/treefmt.nix
@@ -7,22 +7,25 @@
 let
   cfg = config.treefmt;
 
-  treefmt-nix = config.lib.getInput {
+  inputArgs = {
     name = "treefmt-nix";
     url = "github:numtide/treefmt-nix";
     attribute = "treefmt";
     follows = [ "nixpkgs" ];
   };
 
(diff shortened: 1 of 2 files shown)


sandydoo:
Hmm this then prints a less than ideal error in the usual case where the user enables treefmt, but hasn't yet added the input.

domenkozar:
This one is similar to git-hooks, I'm not sure we have other external submodules.

sandydoo:
> This one is similar to git-hooks, I'm not sure we have other external submodules.

I think this'll work now. It's definitely doable, just a question of figuring out a decent API for us. We can apply the same pattern to git-hooks to make it optional in the future.

domenkozar:
Then we need to make sure that docs do have the inputs so we can generate the options?

sandydoo:
> Then we need to make sure that docs do have the inputs so we can generate the options?

Already done. This is our dedicated doc gen shell: https://github.com/cachix/devenv/blob/main/docs/gen/devenv.yaml

Issue #2149, opened 2025-09-11
Secretspec 1Password Signin breaks Devenv


bromanko:
**Describe the bug**
I have enabled Secretspec with a 1Password provider. If `op` is not signed in, `devenv shell` fails:

```
❯ devenv shell
✓ Building shell in 72.3ms
Error:   × Provider operation failed: [ERROR] 2025/09/11 15:13:20 account is not signed in
  │
```

I don't think this should be an error. I think it should be a warning. My `op` session only lasts for 30 minutes and it's a significant disruption to workflow to have devenv fail to load in that circumstance.  

**To reproduce**

1. Create a devenv using secretspec with a 1Password backend
2. Don't signin to `op`
3. Attempt to run `devenv shell`

**Version**
devenv 1.8.2 (aarch64-darwin)

domenkozar:
If you enable `secretspec` integration, then it assumes that you can provide secrets for the environment otherwise it fails. Would you expect environment to be activated without secrets?

bromanko:
I get where you're coming from. But the current workflow leads to some awkward developer experiences. For example, I recently added a new required secret to `secretspec.yml`. I need to then use the `secretspec` cli to add the secret. But, I'm not able to load the devenv shell which contains it. 

```
❯ cd ooo
direnv: loading ~/Code/ooo/.envrc
direnv: using devenv
direnv: ([/nix/store/qdsl1sp5p4hpz1ihifkgx3dnn66yca26-direnv-2.37.1/bin/direnv export fish]) is taking a while to execute. Use CTRL-C to give up.
Error:   × Required secrets are missing: OOO_GOOGLE_CLIENT_SECRET, OOO_GOOGLE_CLIENT_ID (provider: onepassword, profile: development)
```

So the failure to load secrets prevents me from getting access to the tools I need to fix the issue.

domenkozar:
@bromanko would it make sense if instead of failing that a secret is missing, it would ask you for the secrets?

domenkozar:
@bromanko I've made the change in https://github.com/cachix/devenv/commit/871945b5fa79e9c03335cdbd90d1501f10fe9078, let me know if this doesn't address your issue.

PR #2248, opened 2025-10-29
fix: add warning messages when required secrets are not set

Morsicus:
# Context

Currently, when required secrets (configured with secretspec) are missing, running `devenv shell` fails to build the environment.

It leads to a strange developer experience because adding a secret is not easy. Indeed the `secretspec check` command is accessible in the environment but the environment is not built.

It also give clunky developer experience during onboarding (ie. When the secrets are not yet been set)

# Proposed solution

I understand that we probably don't want to have a shell that has been built without `required` secrets. Because the shell won't completely be working. 

However it's a Chicken-Egg problem that need to be solved to ease the process of adding new secrets and onboard new developers.

My current implementation add a "Status concept" in the devenv structure. This status is updated when we assemble the environment. If we find "missing secrets", rather than bailing out, we change the status to  `MissingSecrets` and the rest of the process of "assembling the environment" continue.

When we exec inside the shell, we can actually read the status and print a warning message to the user with some hints to fix the problem.

## Before

image

## After

image

# Testing

To test this change:

- Configure a project with secretspec.toml requiring secrets:

```
[project]
name = "Test Project"
revision = "1.0"

[profiles.default]
SECRET_1 = { description = "A secret", required = true }
SECRET_2 = { description = "A secret", required = true }
SECRET_3 = { description = "A secret", required = true }
SECRET_4 = { description = "A secret", required = true }
SECRET_5 = { description = "A secret", required = true }
SECRET_6 = { description = "A secret", required = true }
SECRET_7 = { description = "A secret", required = true }

[profiles.development]
```
- Don't set the required secrets
- Run `devenv shell`
- Observe: Shell opens (instead of failing) with red warning
- Run `secretspec check` inside the shell and configure secrets
- Exit and re-run `devenv shell`
- Observe: Shell opens without warning

# Future works (if we want to)

- We could reuse the status for different needs
- We could print the status in the TUI
- We could add a command `devenv status` or `devenv doctor` `devenv healthcheck` 

# Resources

- [Discord discussion about this problem](https://discord.com/channels/1036369714731036712/1391835385805668432/1432491133329473717)
- I think this PR could be a decent workaround for https://github.com/cachix/devenv/issues/2149#issue-3408178673

Let me know what I could change or if you have a better idea - I might have missed some side effects (I'm just ramping up on the code base :D)

### DIFF
--- a/devenv/src/devenv.rs
+++ b/devenv/src/devenv.rs
@@ -4,6 +4,7 @@ use ::nix::unistd::Pid;
 use clap::crate_version;
 use cli_table::Table;
 use cli_table::{WithTitle, print_stderr};
+use console::style;
 use include_dir::{Dir, include_dir};
 use miette::{IntoDiagnostic, Result, WrapErr, bail, miette};
 use once_cell::sync::Lazy;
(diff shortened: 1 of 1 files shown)


Morsicus:
:question: I'm thinking about an alternative approach: Using the "ad-hoc environment" with `secretspec` installed in order to set the secrets (and workaround the current behavior that bail out).

I don't know if it's a good idea... :x Just sharing here if anyone has any thought :)

domenkozar:
Maybe devenv should be running `check` so it asks for passwords directly?

domenkozar:
@Morsicus what do you think? :)

Morsicus:
Hello @domenkozar, 

Sorry for the delay, busy weeks :x. 

> @Morsicus what do you think? :)

Sure. It will be more straightforward. I'll open an other PR today or tomorrow.

Morsicus:
Closed in favor of #2270

PR #2251, opened 2025-10-29
treefmt: fix conflicting tree-root options when wrapping treefmt

### LINKED ISSUE #2249: treefmt broken because of redundant flag
**Describe the bug**

b7080bee7c606739be5fb3d6c506e70474d8d3bc seems to break `treefmt` within devenv.

https://github.com/cachix/devenv/blob/b59500b629c08829825d4f4eca68169ea3a8821f/src/modules/integrations/treefmt.nix#L35 adds the flag `--tree-root`.  
treefmt-nix already defines the flag `--tree-root-file` (https://github.com/numtide/treefmt-nix/blob/2eddae033e4e74bf581c2d1dfa101f9033dbd2dc/module-options.nix#L246) in their own wrapper.  

Calling the wrapped binary with a wrapper produces an error since the two flags are mutually exclusive.

**To reproduce**

- add to `devenv.nix` 
  ```nix
    treefmt = {
      enable = true;
    };
  ```
- run `treefmt`
- error:
  ```
  Error: if any flags in the group [tree-root tree-root-cmd tree-root-file] are set none of the others can be; [tree-root tree-root-file] were all set
  ```

**Version**

devenv 1.10.0 (x86_64-linux)

sandydoo:
Fixes #2249.

Uses the unwrapped `treefmt` executable in our wrapper to avoid providing conflicting tree root options.

This also adds proper support for `projectRootFile`, should the user decide to use it.

### DIFF
--- a/examples/treefmt/devenv.nix
+++ b/examples/treefmt/devenv.nix
@@ -5,38 +5,36 @@
     pkgs.diffutils
   ];
 
+  git-hooks.hooks = {
+    treefmt.enable = true;
+  };
+
   treefmt = {
     enable = true;
-
-    config.projectRootFile = "devenv.nix";
 
     config.programs = {
       nixfmt.enable = true;
       rustfmt.enable = true;
     };
-  };
 
-  git-hooks.hooks = {
-    treefmt.enable = true;
-  };
-
-  treefmt.config.settings.formatter = {
-    "yq-json" = {
-      command = "${lib.getExe pkgs.bash}";
-      options = [
-        "-euc"
-        ''
-          for file in "$@"; do
-            ${lib.getExe pkgs.yq-go} -i --output-format=json $file
-          done
-        ''
-        "--" # bash swallows the second argument when using -c
-      ];
-      includes = [ "*.json" ];
-      excludes = [
-        ".git/*"
-        ".devenv/*"
-      ];
+    config.settings.formatter = {
+      "yq-json" = {
+        command = "${lib.getExe pkgs.bash}";
+        options = [
+          "-euc"
+          ''
+            for file in "$@"; do
+              ${lib.getExe pkgs.yq-go} -i --output-format=json $file
+            done
+          ''
+          "--" # bash swallows the second argument when using -c
+        ];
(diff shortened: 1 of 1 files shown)


examples/treefmt/devenv.nix
```diff
@@ -8,10 +8,9 @@
   treefmt = {
     enable = true;
 
-    config.projectRootFile = "projectRootFile";
+    config.projectRootFile = "devenv.nix";
```
domenkozar: This should be deleted, it's not needed since we set the default

src/modules/integrations/treefmt.nix
```diff
@@ -30,10 +30,22 @@ let
...
+  # treefmt-nix provides a `config.build.wrapper` that uses `config.projectRootFile` to find the tree root.
+  # We instead set the tree root directory based on the information provided by the CLI.
+  # If the user sets a custom `projectRootFile`, we'll use that.
+  treefmtWrapper =
+    let
+      treeRootOption =
+        if cfg.config.projectRootFile != "" then
+          "--tree-root-file " + lib.escapeShellArg cfg.config.projectRootFile
+        else
+          "--tree-root " + lib.escapeShellArg treeRoot;
```
basbebe, line 44: As I read https://github.com/numtide/treefmt-nix/blob/2eddae033e4e74bf581c2d1dfa101f9033dbd2dc/module-options.nix#L246, the wrapper of treefmt-nix will always add `--tree-root-file`, creating a conflict if the user doesn't set `config.projectRootFile`.  
(`treefmt` will fail with an error in this case since the two flags are mutually exclusive)

sandydoo: The fix for that is [in the script below](https://github.com/cachix/devenv/pull/2251/files#diff-a42412996345a0c1d9a8e3f56c0a0cbb54cd0a18f1b74ed0e8d2d09b1fbba639R47): we use the treefmt package directly, not the wrapper.

basbebe: Oversaw that part.
My bad!

PR #2256, opened 2025-11-01
refactor: use struct for passing args to Nix flake template

domenkozar:
Replace manual string formatting in devenv.rs with a typed struct approach for constructing Nix flake template arguments.

### DIFF
--- a/devenv/src/devenv.rs
+++ b/devenv/src/devenv.rs
@@ -1482,10 +1482,12 @@ impl Devenv {
         let git_root = config.git_root.clone();
 
         // Create the Nix arguments struct
+        let project_input_ref = format!("path:{}", self.devenv_root.display());
         let args = NixArgs {
             version: crate_version!(),
             system: &self.global_options.system,
             devenv_root: &self.devenv_root,
+            project_input_ref: &project_input_ref,
             devenv_dotfile: &self.devenv_dotfile,
             devenv_dotfile_path: &dotfile_relative_path,
             devenv_tmpdir: &self.devenv_tmp,
--- a/devenv/src/nix_args.rs
+++ b/devenv/src/nix_args.rs
@@ -20,6 +20,9 @@ pub struct NixArgs<'a> {
     /// Serialized as a Nix path literal by ser_nix
     pub devenv_root: &'a Path,
 
+    /// Project input reference as string
+    pub project_input_ref: &'a str,
+
     /// Absolute path to the devenv dotfile directory (.devenv)
     /// Serialized as a Nix path literal by ser_nix
     pub devenv_dotfile: &'a Path,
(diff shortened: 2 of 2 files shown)


devenv/src/nix_args.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub devenv_direnvrc_latest_version: u8,
+
+    /// Container name if building/running a container, otherwise null
+    pub container_name: Option<String>,
+
+    /// List of active profiles to enable
+    pub active_profiles: Vec<String>,
+
+    /// Current system hostname
+    pub hostname: String,
```
sandydoo: Hostname should be an option. We can then remove the unwrap to "unknown".
```suggestion
    pub hostname: Option<String>,
```

devenv/src/nix_args.rs
```diff
@@ -0,0 +1,59 @@
...
+
+    /// Current username
+    pub username: String,
```
sandydoo: Same as hostname.
```suggestion
    pub username: Option<String>,
```

devenv/src/devenv.rs
```diff
@@ -1492,51 +1492,38 @@ impl Devenv {
         let username = whoami::username();
```
sandydoo: This'll return an `Option<String>` vs some random "Unknown" string.
```suggestion
        let username = whoami::fallible::username().ok();
```

devenv/src/devenv.rs
```diff
@@ -1492,51 +1492,38 @@ impl Devenv {
...
+            devenv_tmpdir: PathBuf::from(self.devenv_tmp.clone()),
+            devenv_runtime: self.devenv_runtime.clone(),
+            devenv_istesting: is_testing,
+            devenv_direnvrc_latest_version: *DIRENVRC_VERSION,
+            container_name: self.container_name.clone(),
+            active_profiles: if self.global_options.profile.is_empty() {
+                vec![]
+            } else {
+                self.global_options.profile.clone()
+            },
```
sandydoo: What about this?
```suggestion
            active_profiles: self.global_options.profile.clone(),
```

devenv/src/nix_args.rs
```diff
@@ -0,0 +1,59 @@
...
+    /// Serialized as a Nix path literal by ser_nix
+    pub devenv_root: PathBuf,
+
+    /// Absolute path to the devenv dotfile directory (.devenv)
+    /// Serialized as a Nix path literal by ser_nix
+    pub devenv_dotfile: PathBuf,
+
+    /// Relative Nix path to the dotfile directory (e.g., ".devenv")
+    /// Serialized as a Nix path literal by ser_nix
+    pub devenv_dotfile_path: PathBuf,
```
sandydoo:
```suggestion
    /// Relative Nix path to the dotfile directory (e.g., ".devenv")
    /// Serialized as a Nix path literal by ser_nix
    /// TODO: remove in the next release
    pub devenv_dotfile_path: PathBuf,
```

sandydoo (changes requested):

sandydoo:
@domenkozar I've modified `NixArgs` to not allocate anything and added a test.

The test highlighted an issue with `ser_nix`: it doesn't serialize `None` fields, which won't work with our current templating setup. `ser_nix` doesn't currently expose a way to control this.

sandydoo:
I fixed the issue with `null` not being serialized, but the next issue is more complicated.

Due to flake eval issues, we need the `devenv_dotfile_path` to be a path, not a string. Getting serde to generate strings _without_ quotes might be messy.

sandydoo:
Added very wip support for paths in ser_nix. Let's see what the tests say.

PR #2270, opened 2025-11-08
fix: ask for required secrets rather than bailing out

Morsicus:
# Context

Currently, when required secrets (configured with secretspec) are missing, running `devenv shell` fails to build the environment.

It leads to a strange developer experience because adding a secret is not easy. Indeed the secretspec check command is accessible in the environment but the environment is not built.

It also give clunky developer experience during onboarding (ie. When the secrets are not yet been set)

This PR wants to resolve that by following the recommendation from @domenkozar (in [this previous PR](https://github.com/cachix/devenv/pull/2248#issuecomment-3473519017)). 

# Solution

  When one or more required secrets are not configured, `devenv shell` now prompts the user interactively for the missing secrets instead of failing immediately.

# Implementations

## Extract secretspec validation into dedicated method

  - Handles all secretspec validation logic in one place
  - Is idempotent (checks if secrets are already resolved and returns early)
  - Can be called early in the execution flow to handle interactive prompts before other operations

## Call secretspec check before building shell

  Updated `prepare_shell()` to call `check_secretspec()` before `get_dev_environment()`, ensuring the interactive secret prompts complete before any "Building shell" messages appear.

## Continue after interactive secret configuration

  - Run `secrets.check()` when validation fails (prompts user interactively)
  - After successful configuration, re-validate and store the secrets
  - Continue with normal execution instead of bailing

# Open Questions

## Continuing after secret configuration vs asking user to retry

I did few tests and after the secrets update the flake needs to be rebuild and it seems to be slow. It happens that it failed. If we run `devenv shell` after, it works well.

Wouldn't it be better to ask the user to just re-execute `devenv shell`? 

What do you think? :)

### DIFF
--- a/devenv/src/devenv.rs
+++ b/devenv/src/devenv.rs
@@ -324,6 +324,8 @@ impl Devenv {
         cmd: &Option<String>,
         args: &[String],
     ) -> Result<process::Command> {
+        self.check_secretspec().await?;
+
         let DevEnv { output, .. } = self.get_dev_environment(false).await?;
 
         let bash = match self.nix.get_bash(false).await {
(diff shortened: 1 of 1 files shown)


domenkozar:
Thanks for this! I've noticed code can be simplified if `check()` returns secrets so I've commited a slightly different solution in https://github.com/cachix/devenv/commit/871945b5fa79e9c03335cdbd90d1501f10fe9078

Issue #2286, opened 2025-11-13
"execIfModified" hang on **/*.<extension>


shikanime:
When using the `devenv` task with `execIfModified`, the task currently watches files that match the provided patterns (e.g. '**/*.ts'), even if those files or their containing directories are specified in the project's '.gitignore' file. This can cause the command to hang such as against `.devenv` or `.direnv` if they contain a large number of files including go mod or python venv.

PR #2287, opened 2025-11-13
fix: recursive glob matching hidden directories

shikanime:
Task glob should be stricter and ignore Linux hidden files unless specified.

Fixes: https://github.com/cachix/devenv/issues/2286

### DIFF
--- a/devenv-tasks/src/task_cache.rs
+++ b/devenv-tasks/src/task_cache.rs
@@ -7,7 +7,7 @@ use devenv_cache_core::{
     file::TrackedFile,
     time,
 };
-use glob::glob;
+use glob::{glob_with, MatchOptions};
 use serde_json::Value;
 use sqlx::Row;
 use std::path::PathBuf;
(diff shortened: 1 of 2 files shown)


devenv-tasks/src/task_cache.rs
```diff
@@ -19,12 +19,19 @@ pub fn expand_glob_patterns(patterns: &[String]) -> Vec<String> {
...
-            glob(pattern)
-                .ok()
-                .into_iter()
-                .flatten()
-                .filter_map(|path| path.ok())
-                .filter_map(|path| path.to_str().map(String::from))
+            glob_with(
+                pattern,
+                MatchOptions {
+                    case_sensitive: true,
```
domenkozar, line 32: Can we add a comment to each of these options why they exists?

devenv-tasks/src/task_cache.rs
```diff
@@ -7,24 +7,38 @@ use devenv_cache_core::{
...
+use glob::{glob_with, MatchOptions};
 use serde_json::Value;
 use sqlx::Row;
 use std::path::PathBuf;
 use std::time::SystemTime;
 use tracing::{debug, warn};
 
 /// Expand glob patterns into actual file paths
+///
+/// The expansion uses strict matching rules:
```
domenkozar, line 19: @sandydoo this looks good to me, hidden files shouldn't be matched (maybe later on we should expose an option to disable it). WDYT?

shikanime: Honestly, I don't think it'll be necessary; the behaviour introduced by this PR matches how every other Unix tool behave.

PR #2293, opened 2025-11-18
Introduced new option: services.mongodb.replication

hencjo:
Now, one will hardly run production environments and replication using devenv. *However* when using MongoDB transactions their behaviour (read- and write concerns) varies depending on if the MongoDB instance is set up for replication or not.

This patch gives the user the option to configure the devenv mongodb as a one-node replica-set.

When replication is enabled, MongoDB wants a `keyFile` to authenticate other nodes in the cluster. This `keyFile` is created using openssl to make MongoDB happy.

### DIFF
--- a/src/modules/services/mongodb.nix
+++ b/src/modules/services/mongodb.nix
@@ -5,56 +5,81 @@ with lib;
 let
   cfg = config.services.mongodb;
 
-  setupScript = pkgs.writeShellScriptBin "setup-mongodb" ''
-    set -euo pipefail
-    # Abort if the data dir already exists
-    [[ ! -d "$MONGODBDATA" ]] || exit 0
-    mkdir -p "$MONGODBDATA"
-  '';
+  setupScript =
+    let
+      replication =
+        if cfg.replication.enable
+        then ''
+          # Set up replication (requires a keyfile)
+          ${pkgs.openssl}/bin/openssl rand -base64 765 > "$MONGODBDATA/keyfile"
+          chmod 400 "$MONGODBDATA/keyfile"''
+        else "";
+    in
+    pkgs.writeShellScriptBin "setup-mongodb" ''
+      set -euo pipefail
+      # Abort if the data dir already exists
+      [[ ! -d "$MONGODBDATA" ]] || exit 0
+      mkdir -p "$MONGODBDATA"
 
-  startScript = pkgs.writeShellScriptBin "start-mongodb" ''
-    set -euo pipefail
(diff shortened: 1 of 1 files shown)


domenkozar:
Nice addition :) I'm skeptical if we should be generating secrets the old way, we should find a way to declare a secretspec here and then the user can provide it. We probably need to extend secretspec to provision a path and generate the secret if requested. @sandydoo thoughts?

hencjo:
My thoughts, as a user, on this: I wouldn't want to specify a keyFile or specify any secrets in this case. 

The option really only enables a one-node replica-set and there's no support for adding nodes. KeyFile is only used for authenticating other nodes in the same replica-set, not for app connections or interactive users.

I'm probably missing some aspect though. Perhaps it's that we want a deterministic `.devenv/state/mongdb`-directory?

We could set the keyFile to a constant in mongodb.nix; but I'm worried that might be a security issue if users leave MongoDB listening and the keyFile is constant and known for every devenv.

domenkozar:
> My thoughts, as a user, on this: I wouldn't want to specify a keyFile or specify any secrets in this case.
> 
> The option really only enables a one-node replica-set and there's no support for adding nodes. KeyFile is only used for authenticating other nodes in the same replica-set, not for app connections or interactive users.
> 
> I'm probably missing some aspect though. Perhaps it's that we want a deterministic `.devenv/state/mongdb`-directory?
> 
> We could set the keyFile to a constant in mongodb.nix; but I'm worried that might be a security issue if users leave MongoDB listening and the keyFile is constant and known for every devenv.

I'm worried mostly because one day I hope that devenv will also allow running production environments, I hope that we'll handle all secrets via secretspec. Given that we're missing some features for now, I think this is a good compromise :)

domenkozar:
I've opened https://github.com/cachix/secretspec/issues/44 and referencing https://github.com/cachix/secretspec/issues/9

hencjo:
I didn't quite get which of the options you want me to go with :D : 

A) Keep MR as-is (generate keyFile with openssl).
B) Change to keyFile to constant value.
C) Pursue secret-spec.

Devenv is great because of your vision, so I will happily take the path you want on this.

domenkozar:
Let's keep it as it is for now, secretspec needs some work to support this kind of workflow.

domenkozar:
I've opened https://github.com/cachix/secretspec/pull/45 that implements `as_path=true` in secretspec, returning secret as a path that can be then used in Nix.

Issue #1748, opened 2025-02-24
Remove all hooks from .pre-commit-config.yaml if disabled


jashandeep-sohi:
While in the shell, I occasionally like to use the `pre-commit` CLI directly.
This works well, expect when you disable git-hooks.
In that case, `.pre-commit-config.yaml` goes out of sync, because currently it is only updated if at least one pre-commit hook is enabled https://github.com/cachix/devenv/blob/74832581642f2a59f0c2a5b74c6ef221080f3c39/src/modules/integrations/git-hooks.nix#L43

Would it be possible to always run https://github.com/cachix/devenv/blob/74832581642f2a59f0c2a5b74c6ef221080f3c39/src/modules/integrations/git-hooks.nix#L49 to make sure all disabled hooks are removed?

domenkozar:
Would love to get this fixed, happy to accept a PR!

domenkozar:
We should probably just remove the file if there are no hooks and the file is a symlink to the store

domenkozar:
This was fixed in https://github.com/cachix/devenv/pull/1751

PR #2295, opened 2025-11-19
git-hooks: support `git-hooks` being an optional input

sandydoo:
The `git-hooks` input is currently always added by the devenv CLI.
This was done out of development convenience (for ourselves, mostly), but as devenv grows, it makes sense to make this large integration optional.

Another tricky bit is that the integration is enabled whenever any of the hooks are enabled. There's also a global `enable`, but that's imported as part of the submodule. This is why the integration here is quite a bit more complicated than, for example, treefmt.

Future versions of the devenv CLI will remove the default `git-hooks` input.


domenkozar:
We can probably now handle https://github.com/cachix/devenv/issues/1748 by checking if the file symlinks to store and remove it if it's disabled?

EDIT: this is already fixed.

PR #2297, opened 2025-11-19
eval-cache: exclude volatile env vars from cache tracking

### LINKED ISSUE #2107: `devenv tasks run` enter the shell even when run from devenv shell
**Describe the bug**

When running:

```bash
devenv tasks run my-task
```

The command reload the devenv shell even if called from an active shell.

Please add a condition to prevent this.

En aparté, It might be interesting to add `initShell` in addition to `enterShell`:

- `initShell` would run tasks required to initialize any devenv shell (setting PATH, ...)
- `enterShell` would run tasks needed when entering an interactive shell (starting services, starship, ...).

**Version**

devenv 1.8.1 (x86_64-linux)

domenkozar:
Fixes #2107 where running `devenv tasks run` from inside a devenv shell would cause unnecessary rebuilds. Environment variables like TMPDIR and PWD change between shell contexts (e.g., TMPDIR is modified by enterShell), causing cache invalidation even though the project configuration hasn't changed.

This adds a VOLATILE_ENV_VARS constant with environment variables that represent execution context rather than configuration (PWD, TMPDIR, TMP, TEMP, TEMPDIR, OLDPWD, XDG_RUNTIME_DIR). These are now excluded from cache tracking by default, with a watch_env_var() method to opt-in when needed.

Untested.

### DIFF
--- a/devenv-eval-cache/src/command.rs
+++ b/devenv-eval-cache/src/command.rs
@@ -26,12 +26,30 @@ pub enum CommandError {
 
 type OnStderr = Box<dyn Fn(&InternalLog) + Send>;
 
+/// Environment variables that are volatile across execution contexts.
+/// These variables are excluded from eval cache tracking by default because they:
+/// - Change between shell contexts (e.g., TMPDIR is modified by enterShell)
+/// - Represent execution environment rather than project configuration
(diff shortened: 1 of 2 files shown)


domenkozar:
Not sure this fixes the issue, let's take it for redesign of the new eval caching.

PR #2304, opened 2025-11-24
git-hooks: change default package to pkgs.prek

domenkozar:
Add a changelog since it's a breaking change #2300 

Upstream: https://github.com/cachix/git-hooks.nix/pull/656

### DIFF
--- a/src/modules/integrations/claude.nix
+++ b/src/modules/integrations/claude.nix
@@ -49,14 +49,11 @@ let
         })
         hooks;
 
-  # Check if git-hooks task is defined (indicates git-hooks are enabled)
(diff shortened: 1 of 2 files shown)


compactcode:
Just learning about prek, looks like a nice upgrade. Would this require an update to the Claude Code integration as well?

https://github.com/cachix/devenv/blob/5b6cb65d7dec827badc6bbceeac934931beb7f4e/src/modules/integrations/claude.nix#L35

domenkozar:
> Just learning about prek, looks like a nice upgrade. Would this require an update to the Claude Code integration as well?
> 
> https://github.com/cachix/devenv/blob/5b6cb65d7dec827badc6bbceeac934931beb7f4e/src/modules/integrations/claude.nix#L35

Good catch, thanks, fixed!

PR #2321, opened 2025-12-02
docs: Recommend installing devenv from nixpkgs revision matching lock file

Yethal:


domenkozar:
That's not a good idea, because it will use devenv cli from development branch instead of a known release.

What's the motivation behind it?

Yethal:
I noticed much less cache misses in CI when running devenv test after I did this.
Although I do use regular nixpkgs instead of devenv-nixpkgs so perhaps this isn't as useful when running nixpkgs-devenv

domenkozar:
Regular nixpkgs build all devenvs so I'm not sure why it's not hitting cache but it should.

We shouldn't be using development versions of devenv cli.

PR #2331, opened 2025-12-07
Fix chmod permission error when load-exports owned by different user

tennox:
## Summary
- Use `install -m 755` instead of separate `echo` + `chmod` to atomically create the `load-exports` file with correct permissions
- Fixes permission error when the file was previously created by a different user (e.g., root)

## Problem
When `.devenv/load-exports` is owned by another user, `chmod +x` fails with:
```
chmod: changing permissions of '/path/.devenv/load-exports': Operation not permitted
```

## Solution
`install` internally unlinks and recreates the file, so it works even when the existing file is owned by another user. This pattern is already used elsewhere in the codebase (`src/modules/services/minio.nix`).

## Test plan
- [x] Verified `install -m 755` works on files owned by root in a non-sticky directory

### DIFF
--- a/src/modules/tasks.nix
+++ b/src/modules/tasks.nix
@@ -210,6 +210,8 @@ in
         description = "Runs when entering the shell";
         exec = ''
           mkdir -p "$DEVENV_DOTFILE" || { echo "Failed to create $DEVENV_DOTFILE"; exit 1; }
+          # Remove first in case file is owned by another user (chmod would fail otherwise)
+          rm -f "$DEVENV_DOTFILE/load-exports" 2>/dev/null || true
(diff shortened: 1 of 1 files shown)


tennox:
After some [darwin compatibility concerns](https://github.com/nix-community/nix-direnv/pull/647#issuecomment-3644341236) - I did [some tests](https://github.com/tennox/test-install-portability#install-stdin-portability-test-results)
But as we need dynamic content here, a HEREDOC is out - and **I went back to 'simple' approach of running `rm` before**, thus recreating the file. (there would've been the option to create temp file and pass that to install, but I thought that overcomplicates things.

PR #2333, opened 2025-12-08
android: use dynamic version lookup from nixpkgs

domenkozar:
Read platform-tools and emulator versions from nixpkgs' repo.json instead of hardcoding them. This prevents version mismatch errors when nixpkgs removes older Android SDK versions.

### DIFF
--- a/docs/src/reference/options.md
+++ b/docs/src/reference/options.md
@@ -225,7 +225,7 @@ boolean
 
 
 The version of the Android Emulator to install.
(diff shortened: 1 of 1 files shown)


k3yss:
Looks good to me! I have verified that we don't see a warning while building the shell, we might need to add logic to pin specific version for react-native and flutter to work out of box, as I remember both require older version to work properly. I will verify this

PR #1999, opened 2025-07-10
fix: correctly set UV_PYTHON to the wrapped python interpreter

mauricege:
**Underlying Issue**
Since #1959, UV_PYTHON is forced to devenv's python interpreter. However, it is the unwrapped interpreter. This of course breaks the native libs functionality.

**To reproduce**
Create a devenv.nix with `languages.python.uv.sync.enable = true`:
```nix
{ pkgs, config, inputs, ... }:
{
  languages.python = {
    enable = true;
    venv.enable = true;
    uv = {
      enable = true;
      sync.enable = true;
    };
  };
}
```
Check UV_PYTHON variable from shell:
```sh
echo $UV_PYTHON
> /nix/store/fqm9bqqlmaqqr02qbalm1bazp810qfiw-python3-3.12.9/bin/python3.12 <- This is an unwrapped interpreter
```

**Fix**
Just use the wrapped python bound to `package` in `python.nix`

PR #2336, opened 2025-12-10
python: configure uv to use the python interpreter from the profile

sandydoo:
Now that we correctly wrap python to support Nix-provide libraries, we should revisit setting `UV_PYTHON`.

Follow up to #1959, #1999, and #2024.


domenkozar:
We should add a test because it's very fragile around this code.

sandydoo:
Superseded by #2623

PR #2342, opened 2025-12-13
Refactor/use devenv activity for tasks

domenkozar:


devenv-tui/src/app.rs
```diff
@@ -142,20 +142,23 @@ impl TuiApp {
...
-        };
-
-        if lines_to_clear > 0 {
-            let mut stdout = io::stdout();
-            let _ = execute!(
-                stdout,
-                cursor::MoveToPreviousLine(lines_to_clear),
-                terminal::Clear(terminal::ClearType::FromCursorDown)
-            );
+        // Only clear the TUI output on user interrupt (Ctrl+C), not on normal completion.
```
domenkozar, line 145: @sandydoo this is probably not needed?

devenv-activity/src/stack.rs
```diff
@@ -51,6 +79,31 @@ pub(crate) fn get_current_stack() -> Vec<u64> {
...
+    F: std::future::Future<Output = T>,
+{
+    let mut stack = get_current_stack();
+    stack.push(parent_id);
+    ACTIVITY_STACK.scope(RefCell::new(stack), f).await
+}
+
 /// Emit a standalone message, associated with the current activity if one exists
 pub fn message(level: ActivityLevel, text: impl Into<String>) {
     message_with_details(level, text, None)
```
sandydoo, line 85: Does `in_activity` not work? That get's you tracing automatically.

devenv-tui/src/app.rs
sandydoo: Would need to test

devenv-tui/src/app.rs
```diff
@@ -264,21 +267,34 @@ fn MainView(mut hooks: Hooks) -> impl Into<AnyElement<'static>> {
...
-            View(width: terminal_width) {
-                #(vec![view(&model_guard, &ui).into()])
-            }
-        }
-    } else {
-        element!(View(width: terminal_width))
+    // For normal completion (is_done), trigger shutdown AFTER this frame renders.
+    // This ensures the final state is displayed before exit.
+    if is_done {
+        shutdown.shutdown();
```
sandydoo: Could this shutdown logic at the start of the render loop?

devenv-tui/src/lib.rs
```diff
@@ -26,11 +26,18 @@ pub use model_events::UiEvent;
 /// - Renders immediately on first notification (leading edge)
 /// - Sleeps to enforce FPS cap
 /// - Subsequent notifications during sleep are coalesced (model accumulates changes)
+/// - Also redraws periodically (every 100ms) to catch final state after events stop
```
sandydoo: This looks suspicious, like it's hiding another bug

devenv-activity/src/instrument.rs
```diff
@@ -24,9 +24,23 @@ use crate::stack::{ACTIVITY_STACK, get_current_stack};
...
+/// });
+/// ```
 pub trait ActivityInstrument: Future + Sized {
     /// Instrument this future with the given activity's context.
-    fn in_activity(self, activity: &Activity) -> impl Future<Output = Self::Output> {
+    ///
+    /// The `&Activity` is only used to extract the ID and span at call time;
+    /// the returned future captures only `Self`, not the `&Activity` reference.
+    /// This means when `Self` is `'static`, the returned future is also `'static`.
+    fn in_activity(self, activity: &Activity) -> impl Future<Output = Self::Output> + use<Self> {
```
sandydoo, line 43: Oh, nice!

PR #2347, opened 2025-12-15
android: add defaultText for version options

roberth:
Adds defaultText to platformTools.version and emulator.version options because devenv documentation should not make assumptions about the nixpkgs version, which is controlled by users. The list of available versions would be incorrect for users on different nixpkgs versions.

Also improves descriptions to explain how users can discover the available versions for their nixpkgs version.

### DIFF
--- a/docs/src/reference/options.md
+++ b/docs/src/reference/options.md
@@ -225,7 +225,8 @@ boolean
 
 
 The version of the Android Emulator to install.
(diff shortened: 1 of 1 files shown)


domenkozar:
@k3yss

k3yss:
@roberth This PR #2333 introduces an `availableVersions` function which can be used to dynamically generate the available versions instead of giving a generic response. Can you evaluate and possibly update your PR to use this instead?

roberth:
@k3yss That looks very useful, and I believe that's what it did, but it does not provide reliable information for a static documentation website like `devenv.sh` or `flake.parts`

k3yss:
@roberth I missed the context that we are going to do this for the documentation site. I believe this should be fine from the correctness point of view, but for the end-user it's a few more steps to get the version information. This looks to be the best option for us judging by the nature of nixpkgs.


Approved!

PR #2350, opened 2025-12-16
files: cleanup of dangling symlinks

theoborealis:
this PR fixes https://github.com/cachix/devenv/issues/2351

implementation is consistent with the current codebase, following the git-hooks state tracking approach


domenkozar:
Sorry I completely missed this one while working on #2351, but this is fixed now!

PR #2357, opened 2025-12-25
Add languages.ruby.docSupport option for in-terminal ri document accessing

hukaidong:
When developing with Ruby, it's common to check a function's documentation by running `ri SomeClass`. By default, the Nix Ruby package doesn't ship with `ri` documentation, but a configurable `docSupport` option exists in the package. Currently, users can enable this by overriding the Ruby package via `languages.ruby.package`:

```nix
languages.ruby = {
  enable = true;
  package = (
    (inputs.nixpkgs-ruby.lib.packageFromRubyVersionFile {
      file = ./.ruby-version;
      system = pkgs.stdenv.system;
    }).override
      { docSupport = true; }
  );

  bundler.enable = true;
  ...
};
```

However, this approach prevents users from using other devenv features like `languages.ruby.version` or `languages.ruby.versionFile`. This patch adds a `languages.ruby.docSupport` option, allowing users to enable `ri` documentation while still using the standard version configuration:

```nix
languages.ruby = {
  enable = true;
  versionFile = ./.ruby-version;
  docSupport = true;

  bundler.enable = true;
  ...
};
```


domenkozar:
cc @bobvanderlinden

hukaidong:
Want to check if everything looks well on this pr? Just make sure it will not be left and forgot

src/modules/languages/ruby.nix
```diff
@@ -57,6 +61,8 @@ in
         description = "The bundler package to use.";
       };
     };
+
+    docSupport = lib.mkEnableOption "enable documentation support for Ruby packages";
```
domenkozar: let's call it documentation.enable

hukaidong: updated and tested

PR #2369, opened 2025-12-30
mkcert: Allow configurable filenames

Yethal:
Default mkcert filenames change every time the certificate list is updated, for users who'd like to use it elsewhere (i.e in caddy config) it might be preferable to set the names explicitly

### DIFF
--- a/docs/src/reference/options.md
+++ b/docs/src/reference/options.md
@@ -992,6 +992,32 @@ null or string
 
 
 
+## certFile
+
+
+
+Custom certificate file name, uses mkcert default if unset
+
+
+
+*Type:*
+null or string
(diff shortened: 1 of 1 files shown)


src/modules/integrations/mkcert.nix
```diff
@@ -15,6 +24,26 @@ in
         "*.example.com"
       ];
     };
+    certificateOptions = lib.mkOption {
```
domenkozar: let's expose `certFile` and `keyFile` at the top-level without making the implementation about cli options in the options list.

Yethal: Done

Issue #2243, opened 2025-10-26
mysql process doesn’t stop with mprocs


sephii:
**Describe the bug**

Using mysql with the mprocs manager makes it hang on exit. This is caused by [this line](https://github.com/cachix/devenv/blob/7419c04fc798d5d5918413d4cb6c8629f9d4e8a3/src/modules/services/mysql.nix#L321) which creates a new process, and mprocs seems to be [unable to stop a process and its children](https://github.com/pvolok/mprocs/issues/76). Prefixing the aforementioned mysql line with `exec` fixes the issue.

**To reproduce**

```nix
{
  services.mysql.enable = true;
  process.manager.implementation = "mprocs";
}
```
Run `devenv up` and then type `q` to quit. mprocs will hang indefinitely.

**Version**
I’m using flakes.

PR #2372, opened 2026-01-01
Fix SIGTERM not working for MySQL and Redis services

pokonski:
Fixes https://github.com/cachix/devenv/issues/2243

Both Redis and MySQL exec scripts are missing `exec` prefix, so they spawn a separate process that none of the process manager can correctly shut down (tried mprocs, process-compose, hivemind, overmind)

### DIFF
--- a/src/modules/services/caddy.nix
+++ b/src/modules/services/caddy.nix
@@ -190,6 +190,6 @@ in
   };
 
   config = lib.mkIf cfg.enable {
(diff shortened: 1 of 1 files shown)


domenkozar:
Do we need this for all processes? Then it should be part of processes definition, so it's not repeated all over the place

pokonski:
That's highly likely though I can only vouch for the services I use in my projects (minio, postgres, redis, mysql, mailpit)

Minio, Mailpit and Postgres are prefixed with `exec` and shut down correctly. 

Some services have more custom scripts that can't be simply prefixed "magically", I think. Like this:

https://github.com/cachix/devenv/blob/main/src/modules/services/minio.nix#L36

pokonski:
Caddy was also affected, so pushed a fix for it too

domenkozar:
See https://github.com/cachix/devenv/pull/2375 for a more general fix

PR #2375, opened 2026-01-02
fix(processes): auto-add exec for proper SIGTERM handling

### LINKED ISSUE #2243: mysql process doesn’t stop with mprocs

domenkozar:
## Summary

Automatically add `exec` prefix for single-line process commands in the task script wrapper. This ensures SIGTERM reaches the actual process instead of the shell wrapper, fixing the issue where process managers (mprocs, process-compose, etc.) couldn't properly shut down services.

**How it works:**
- For single-line process commands: `exec` is automatically added
- For multi-line scripts: users can manually place `exec` before the main process command

**Services updated:**
- `postgres`, `keycloak`: removed manual `exec` (now auto-added)
- `httpbin`, `sqld`: converted to single-line, removed manual `exec`
- `mailpit`, `typesense`: refactored to use setup tasks for `mkdir`, process command is now single-line

This is a more general fix than #2372 - instead of adding `exec` to individual services, it handles all ~38 services automatically.

Fixes #2243
Supersedes #2372

## Test plan

- [ ] Verify `devenv up` starts processes correctly
- [ ] Verify SIGTERM properly shuts down processes (no hanging)
- [ ] Test with mprocs process manager

### DIFF
--- a/src/modules/tasks.nix
+++ b/src/modules/tasks.nix
@@ -43,11 +43,12 @@ let
                 if config.binary != null
                 then config.binary == "bash"
                 else config.package.meta.mainProgram or null == "bash";
-              # Only auto-add exec for single-line process commands.
+              # Only auto-add exec for single-line process commands that don't already have exec.
               # Multi-line scripts need manual exec placement before the main process.
               trimmedCommand = lib.strings.trim command;
               isSingleLine = !lib.hasInfix "\n" trimmedCommand;
-              addExec = config.type == "process" && isSingleLine;
+              alreadyHasExec = lib.hasPrefix "exec " trimmedCommand;
+              addExec = config.type == "process" && isSingleLine && !alreadyHasExec;
             in
             pkgs.writeScript name ''
               #!${binary}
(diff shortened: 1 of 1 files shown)


src/modules/tasks.nix
```diff
@@ -43,11 +43,16 @@ let
...
+              # Multi-line scripts need manual exec placement before the main process.
+              trimmedCommand = lib.strings.trim command;
+              isSingleLine = !lib.hasInfix "\n" trimmedCommand;
+              addExec = config.type == "process" && isSingleLine;
             in
             pkgs.writeScript name ''
               #!${binary}
               ${lib.optionalString (!isStatus && isBash) "set -e"}
-              ${command}
+              ${lib.optionalString addExec "exec "}${command}
```
thenonameguy, line 56: This should be made idempotent, in case custom `processes` already added this within the `command`. Otherwise this will cause `exec exec`.

thenonameguy: Additionally, I'd probably look for `exec` in multi-line strings and emit a warning if it's not detected. Sometimes long single-program bash scripts are split into newlines for their args.

domenkozar: good catch! see https://github.com/cachix/devenv/pull/2375/commits/900593731f052bfdbcdd9870344db5df2791d212

PR #2389, opened 2026-01-09
feat: add automatic port allocation for processes

domenkozar:
## Summary

Add automatic port allocation for processes, solving port conflicts when running multiple instances of the same devenv environment.

### The problem

When running multiple devenv instances of the same project (e.g., on different branches, in CI, or by multiple developers on a shared machine), hardcoded ports cause conflicts. The second instance fails to start because the port is already in use.

### The solution

- Add `processes.<name>.ports.<port>.allocate` option that specifies a base port
- Devenv automatically finds a free port starting from that base, incrementing until one is available
- The resolved port is accessible via `config.processes.<name>.ports.<port>.value`
- Ports are held during Nix evaluation to prevent race conditions, then released just before processes start so applications can bind to them

### Example

```nix
processes.server = {
  ports.http.allocate = 8080;
  exec = ''
    python -m http.server ${toString config.processes.server.ports.http.value}
  '';
};
```

If port 8080 is in use, devenv will automatically use 8081, 8082, etc.

### DIFF
--- a/devenv-nix-backend/src/nix_backend.rs
+++ b/devenv-nix-backend/src/nix_backend.rs
@@ -1012,9 +1012,10 @@ impl NixBackend for NixRustBackend {
 
             let args_nix = ser_nix::to_string(args).unwrap_or_else(|_| "{}".to_string());
             let cache_key_args = format!(
-                "{}:port_allocation={}",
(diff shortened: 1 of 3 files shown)


domenkozar:
Feedback to address: https://x.com/Shyim97/status/2009648611752095851

PR #2396, opened 2026-01-12
feat(claude): make hookType a string for forward compatibility

### LINKED ISSUE #2395: Make `claude.code.hooks.*.hookType` a string instead of enum
## Summary

Change the `hookType` option from a fixed enum to a string type for forward compatibility with Claude Code's hook system.

## Peoblem
The current `hookType` option uses a fixed enum:

```nix
type = lib.types.enum [ "PreToolUse" "PostToolUse" "Notification" "Stop" "SubagentStop" ];
```

This prevents users from configuring hooks for other valid Claude Code hook types like:
- SessionStart
- PermissionRequest
- UserPromptSubmit

Claude Code's hook system supports many hook types and may add more in the future. Using a string type makes the devenv integration forward-compatible and allows users to leverage the full hook API without requiring devenv updates for each new hook type.

gonzaloetjo:
Fixes #2395

## Summary
Add missing hook types to `hookType` enum:
- PostToolUseFailure
- UserPromptSubmit
- SessionStart
- SessionEnd
- SubagentStart
- PreCompact
- PermissionRequest

### DIFF
--- a/src/modules/integrations/claude.nix
+++ b/src/modules/integrations/claude.nix
@@ -89,9 +89,16 @@ let
     hooks = lib.filterAttrs (n: v: v != null) {
       PreToolUse = buildHooks "PreToolUse" (groupedHooks.PreToolUse or [ ]);
       PostToolUse = buildHooks "PostToolUse" postToolUseHooks;
+      PostToolUseFailure = buildHooks "PostToolUseFailure" (groupedHooks.PostToolUseFailure or [ ]);
       Notification = buildHooks "Notification" (groupedHooks.Notification or [ ]);
+      UserPromptSubmit = buildHooks "UserPromptSubmit" (groupedHooks.UserPromptSubmit or [ ]);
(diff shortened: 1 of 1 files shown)


sandydoo (commented):
I would've preferred to see the new hook types added to the enum and documented in the description. [Latest list here](https://code.claude.com/docs/en/hooks-guide#hook-events-overview).

I doubt this hook list is going to continue growing forever in a way that we can't keep up. If it proves annoying, I'd prefer an `either str` to at least keep the known options in the type.

gonzaloetjo:
Fixed it 

> I would've preferred to see the new hook types added to the enum and documented in the description. [Latest list here](https://code.claude.com/docs/en/hooks-guide#hook-events-overview).
> 
> I doubt this hook list is going to continue growing forever in a way that we can't keep up. If it proves annoying, I'd prefer an `either str` to at least keep the known options in the type.

PR #2423, opened 2026-01-22
feat(tasks): add negation pattern support for execIfModified

### LINKED ISSUE #2422: execIfModified globs traverse symlinks into node_modules, causing massive performance overhead
## Problem

The `execIfModified` glob patterns traverse into symlinked directories (like `node_modules`), causing devenv to hash hundreds of thousands of files instead of just source files.

### Example

```nix
tasks = {
  "lint:check" = {
    exec = "oxlint .";
    execIfModified = [ "**/*.ts" "**/*.tsx" ];
  };
};
```

This pattern matches **178,000+ files** (traversing into `node_modules`) instead of **~800 actual source files**.

### Impact

- **Cold run**: 3:40 → should be ~40s (5x slower)
- **Cached run**: 1:20 → should be ~19s (4x slower)

The actual linter (oxlint) only takes ~1 second. The 200+ seconds of overhead is entirely from devenv hashing files in `node_modules`.

## Root Cause

The `expand_glob_patterns()` function in `devenv-tasks/src/task_cache.rs` uses the basic `glob` crate which:
1. Follows symlinks by default
2. Doesn't support negation/exclusion patterns

```rust
pub fn expand_glob_patterns(patterns: &[String]) -> Vec<String> {
    patterns
        .iter()
        .flat_map(|pattern| {
            glob_with(pattern, MatchOptions { ... })
            // ... no filtering, no exclusions
        })
        .collect()
}
```

## Proposed Solution

Support negation patterns (gitignore-style) like `!**/node_modules/**`:

```nix
execIfModified = [ 
  "**/*.ts" 
  "!**/node_modules/**"  # Exclude node_modules
  "!**/dist/**"          # Exclude build output
];
```

### Implementation

Replace the `glob` crate with `globset` or `ignore` crate, which support:
- Negation patterns (`!pattern`)
- Better symlink handling
- Gitignore-style matching

The change would be localized to `expand_glob_patterns()`:
1. Separate patterns into include vs exclude (patterns starting with `!`)
2. Expand include patterns
3. Filter out paths matching exclude patterns

## Workaround

Currently we've disabled `execIfModified` entirely for lint tasks, since the linters themselves are fast (<1s) and the hashing overhead negates any caching benefit.

## Environment

- devenv version: latest (from source)
- OS: macOS
- Monorepo with ~20 packages, each with `node_modules` symlinks

schickling:
## Summary

Add support for negation patterns (prefixed with `!`) in `execIfModified` glob patterns. This allows users to exclude specific paths from file watching, which is **critical for performance** when working with symlinked directories like `node_modules`.

## Problem

When using patterns like `**/*.ts`, the glob expansion follows symlinks into `node_modules`, causing devenv to hash 100,000+ files instead of the ~800 actual source files. This results in **200+ seconds of overhead** on every task run.

See issue #2422 for details.

## Solution

This PR adds negation pattern support using the `globset` crate:

```nix
execIfModified = ["**/*.ts" "!**/node_modules/**"];
```

The implementation:
1. Separates positive patterns from negation patterns (prefixed with `!`)
2. Expands positive patterns using the existing `glob` crate
3. Filters results through a `GlobSet` built from negation patterns
4. Returns only paths that don't match any negation pattern

## Changes

- Add `globset` dependency for efficient pattern matching
- Add `build_negation_globset()` helper function
- Update `expand_glob_patterns()` to handle negation patterns
- Add comprehensive tests for negation pattern functionality

## Performance Impact

With this change, users can exclude `node_modules` and similar directories from file watching, reducing the number of files to hash from 178,000+ to ~800 in typical JS/TS projects. This provides a **5-10x speedup** for task execution.

Fixes #2422

### DIFF
--- a/devenv-tasks/src/task_cache.rs
+++ b/devenv-tasks/src/task_cache.rs
@@ -117,13 +117,10 @@ fn extract_base_dir(pattern: &str) -> &Path {
             }
         }
         None => {
-            // No separator found - use current directory if pattern starts with special char,
-            // otherwise the pattern itself might be a file/dir name
-            if first_special == 0 {
-                Path::new(".")
-            } else {
-                Path::new(prefix)
-            }
+            // No separator before the first wildcard means the wildcard is in the first segment.
+            // Walk from cwd so patterns like `src*/*.ts` can match `src1/...`, `src2/...`, etc.
+            let _ = prefix;
+            Path::new(".")
         }
     }
 }
@@ -1141,6 +1138,8 @@ mod tests {
         // Pattern starting with glob
         assert_eq!(extract_base_dir("**/*.ts"), Path::new("."));
         assert_eq!(extract_base_dir("*.ts"), Path::new("."));
+        assert_eq!(extract_base_dir("src*/*.ts"), Path::new("."));
+        assert_eq!(extract_base_dir("foo?.nix"), Path::new("."));
 
         // Pattern with no glob - returns directory part since no separator after the prefix
         assert_eq!(extract_base_dir("/foo/bar/file.ts"), Path::new("/foo/bar"));


devenv-tasks/Cargo.toml
```diff
@@ -25,6 +25,7 @@ blake3.workspace = true
 walkdir.workspace = true
 shell-escape.workspace = true
 glob = "0.3.3"
+globset = "0.4"
```
domenkozar: we should then replace glob with globset, not mix the two

devenv-tasks/src/task_cache.rs
```diff
@@ -7,40 +7,136 @@ use devenv_cache_core::{
...
+            }
+        }
+    }
+}
 
 /// Expand glob patterns into actual file paths
 ///
-/// The expansion uses strict matching rules:
-/// - `case_sensitive: true` – patterns are case-sensitive (e.g. `*.RS` will not match `main.rs`)
-/// - `require_literal_separator: true` – path separators must appear literally in the pattern
```
domenkozar, line 21: we probably need tests for these two, I have a feeling this change regressed those two rules now that we're doing our own walking.

devenv-tasks/src/task_cache.rs
```diff
@@ -7,40 +7,136 @@ use devenv_cache_core::{
...
+
+    // Collect unique base directories to walk
+    let mut base_dirs: Vec<&Path> = positive_patterns
+        .iter()
+        .map(|p| extract_base_dir(p))
+        .collect();
+    base_dirs.sort();
+    base_dirs.dedup();
+
+    // Walk each base directory and collect matching files
```
domenkozar: I can take over the PR, but we should be using existing logic for this, maybe something like https://docs.rs/ignore/latest/ignore/

schickling: Switched exec_if_modified expansion to use `ignore` (OverrideBuilder + WalkBuilder) per suggestion, with gitignore/hidden filtering disabled to preserve previous behavior.

Follow-up: after stripping `base_dir` we now anchor single-segment patterns (prefix `/`) to avoid `ignore` basename-at-any-depth semantics. This keeps `src/foo.txt` and `src/*.ts` from matching `src/sub/...` unless users explicitly opt into that with `**/`.

Added a regression unit test to cover the depth/anchoring case.

schickling-assistant:
Addressed feedback from review 3826569030 and rebased onto current `main`.

What changed:
- Attached ignore overrides directly to `WalkBuilder` via `.overrides(overrides)` so excluded dirs are pruned during traversal (not only filtered post hoc).
- Added a literal-path fast path in `expand_glob_patterns` for non-glob patterns, restoring efficient direct existence checks instead of broad directory walks.
- Split command-path invalidation from `exec_if_modified` pattern expansion in `TaskState`: command path is checked/updated separately so negation patterns cannot suppress command-change invalidation.
- Preserved old dotfile behavior (`test_exec_if_modified_dotfiles` still passes).
- Kept docs example as `!**/*.test.ts`.
- Replaced lossy path conversion with `to_str()` filtering.
- Removed Windows-specific backslash normalization from this Unix-targeted path handling.
- Added regression coverage: `test_exec_if_modified_command_change_with_negation`.

Validation:
- `cargo fmt --all --check`
- `CI=1 cargo test -p devenv-tasks` (85 passed)

_Acting on behalf of @schickling via Codex._

devenv-tasks/src/task_cache.rs
```diff
@@ -7,40 +7,203 @@ use devenv_cache_core::{
...
+                warn!("Failed to build ignore overrides: {}", e);
+                continue;
+            }
+        };
+        let overrides_for_match = overrides.clone();
+
+        let mut walk_builder = WalkBuilder::new(&base_dir);
+        walk_builder
+            .hidden(false)
+            .ignore(false)
```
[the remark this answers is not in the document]
schickling-assistant: Addressed in d186ccf2.

Changes:
- Added hidden-path guard in `expand_glob_patterns` so hidden entries are only accepted when matched by patterns that explicitly target hidden components (`.` in the matched path segment).
- Added regression test `test_expand_glob_patterns_requires_explicit_dot_for_hidden_files`:
  - `src/*.ts` matches `visible.ts` but not `.hidden.ts`
  - `src/.*.ts` matches `.hidden.ts`

Validation:
- `cargo fmt --all --check`
- `CI=1 cargo test -p devenv-tasks` (86 passed)

_Acting on behalf of @schickling via Codex._

devenv-tasks/src/task_cache.rs
```diff
@@ -7,40 +7,245 @@ use devenv_cache_core::{
...
+
+        return Path::new(".");
+    }
+
+    // Get the substring up to the first special character
+    let prefix = &pattern[..first_special];
+
+    // Find the last path separator in the prefix to get the directory
+    match prefix.rfind(std::path::MAIN_SEPARATOR) {
+        Some(last_sep) => Path::new(&pattern[..last_sep]),
```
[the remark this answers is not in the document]
schickling-assistant: Addressed in 6bf3204b.

Fix:
- `extract_base_dir` now preserves `/` when the first wildcard appears in the top-level absolute segment (e.g. `/*.ts`, `/nix*`) instead of returning an empty path.

Regression coverage:
- Extended `test_extract_base_dir` with:
  - `extract_base_dir("/*.ts") == Path::new("/")`
  - `extract_base_dir("/nix*") == Path::new("/")`

Validation:
- `cargo fmt --all --check`
- `CI=1 cargo test -p devenv-tasks` (86 passed)

_Acting on behalf of @schickling via Codex._

devenv-tasks/src/task_cache.rs
```diff
@@ -7,40 +7,251 @@ use devenv_cache_core::{
...
+                Path::new(&pattern[..last_sep])
+            }
+        }
+        None => {
+            // No separator found - use current directory if pattern starts with special char,
+            // otherwise the pattern itself might be a file/dir name
+            if first_special == 0 {
+                Path::new(".")
+            } else {
+                Path::new(prefix)
```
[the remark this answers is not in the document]
schickling-assistant: Addressed in 67ab202d.

Fix:
- In `extract_base_dir`, when there is no separator before the first wildcard, base dir now resolves to `.`.
- This preserves matches for first-segment wildcard patterns like `src*/*.ts` and `foo?.nix`.

Regression coverage:
- Extended `test_extract_base_dir` with:
  - `extract_base_dir("src*/*.ts") == Path::new(".")`
  - `extract_base_dir("foo?.nix") == Path::new(".")`

Validation:
- `cargo fmt --all --check`
- `CI=1 cargo test -p devenv-tasks` (86 passed)

_Acting on behalf of @schickling via Codex._

PR #2457, opened 2026-02-03
docs: add defaultText for ports value option

roberth:
The default depends on the allocate option which has no default, so docs generation fails when trying to evaluate it.


sandydoo:
Thanks (again)! I'll enable `warningsAreErrors` on our doc gen to catch these before they land in your projects.

PR #2468, opened 2026-02-07
feat(claude): add option to disable pre-commit on writes

roman:
### Summary

Adds a new `runPreCommitOnWrites` option to the Claude integration that controls whether pre-commit hooks run after every `Edit`, `MultiEdit`, or `Write` operation.

### Motivation

In monorepos with expensive pre-commit hooks, running them after every write operation significantly impacts workflow performance. This option allows users to disable this behavior while keeping `git-hooks` available for manual use.

### Changes

- Added `claude.runPreCommitOnWrites` option (boolean, default: `true`)
- Pre-commit hook now checks both `runPreCommitOnWrites` and `git-hooks.enable`

### Usage

```nix
{
  claude.enable = true;
  claude.runPreCommitOnWrites = false;  # disable per-write pre-commit
  git-hooks.enable = true;  # hooks still available for manual use
}
```

### Backwards Compatibility

Defaults to `true` to preserve existing behavior.

### DIFF
--- a/docs/src/reference/options.md
+++ b/docs/src/reference/options.md
@@ -1860,7 +1860,7 @@ string
 
 
 Automatically runs git-hooks after Claude edits files.
-Only active when ` git-hooks.enable ` is true.
+Enabled by default when ` git-hooks.enable ` is true.
 
 
 
@@ -1873,10 +1873,13 @@ submodule
 
 ```nix
 {
-  command = "true";
-  enable = false;
+  enable = config.git-hooks.enable;
   name = "Run git-hooks";
+  hookType = "PostToolUse";
+  matcher = "^(Edit|MultiEdit|Write)$";
+  command = "cd \"$DEVENV_ROOT\" && \${config.git-hooks.package.meta.mainProgram} run";
 }
+
 ```
 
 *Declared by:*


roman:
@domenkozar can you take a look when you have some cycles please?

sandydoo:
Sorry for letting this one languish for a month!

You're right that we need to expose options to control this, but I think there's a better design.

1. Extract the hooks submodule type into a variable we can reuse.

2. Make `hooks` a freeform type of hook submodules and add a `git-hooks-format`  option so that it shows up in the docs.
   ```nix
    hooks = lib.mkOption {
      type = lib.types.submodule {
        freeformType = lib.types.attrsOf hookSubmodule;
        options.git-hooks-format = lib.mkOption {
          type = hookSubmodule;
          default = {};
          description = ''
            Automatically runs git-hooks after Claude edits files.
            Only active when `git-hooks.enable` is true.
          '';
        };
      };
      default = {};
      description = "Hooks that run at different points in Claude Code's workflow.";
    };
   ```
  
3. Remove `preCommitHook` from `postToolUseHooks` and set the defaults:

    ```nix
    config = lib.mkMerge [
      (lib.mkIf (cfg.enable && config.git-hooks.enable) {
        claude.code.hooks.git-hooks-format = {
          enable = lib.mkDefault true;
          hookType = lib.mkDefault "PostToolUse";
          matcher = lib.mkDefault "^(Edit|MultiEdit|Write)$";
          command = lib.mkDefault ''
            cd "$DEVENV_ROOT" && ${config.git-hooks.package.meta.mainProgram} run
          '';
        };

        # ... rest of existing config (files, infoSections, etc.)
      })
    ];
    ```
  This part needs to be tested because `git-hooks` is an optional input. We might need to set some option defaults to make the module system happy.
  
This makes the `git-hooks` integration visible in the docs and fully configurable.

roman:
@sandydoo good feedback. 

I added the configuration that can be fully customized. I named it `git-hooks-run` rather than `git-hooks-format`, as git-hooks may be used more than just formatting files. 

Let me know what you think.

PR #2513, opened 2026-02-19
Update settings file path in claude.nix

sebb3:
The settingsPath is not used, use it in place of the hardcoded file to allow users to write to settings.local.json instead if conflicting with hand-written config in repo.

### DIFF
--- a/src/modules/integrations/claude.nix
+++ b/src/modules/integrations/claude.nix
@@ -593,7 +593,7 @@ in
 
   config = lib.mkIf cfg.enable {
     files = lib.mkMerge [
-      { "${settingsPath}".json = settingsContent; }
+      { "${cfg.settingsPath}".json = settingsContent; }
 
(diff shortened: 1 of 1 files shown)


src/modules/integrations/claude.nix
```diff
@@ -593,7 +593,7 @@ in
 
   config = lib.mkIf cfg.enable {
     files = lib.mkMerge [
-      { ".claude/settings.json".json = settingsContent; }
+      { "${settingsPath}".json = settingsContent; }
```
sandydoo: I think this needs to come from `cfg`?
```suggestion
      { "${cfg.settingsPath}".json = settingsContent; }
```

sebb3: my first contrib to this amazing piece of software, and I fail on the ONE word committed 😳

PR #2518, opened 2026-02-23
fix: process duplication when using before/after with process-compose

### LINKED ISSUE #2514: "devenv up" duplicates processes when defining a dependency between them
**Describe the bug**

When I define a dependency graph between `devenv` processes using `processes.<name>.before` or `processes.<name>.after`, the generated `process-compose.yaml` runs each process via `devenv-tasks run` with `--mode all`.

Because each process is started with `--mode all`, each “process” ends up running *all* devenv tasks, not just its own `devenv:processes:<name>` task. With two processes (`foo` and `bar`), `devenv up` runs both process tasks twice (and each process shows output from both tasks).

Intended behavior: each process should run exactly once, with the parent process running before the child, and each `process-compose` process should only execute its corresponding task (i.e., should not use `--mode all`).

**To reproduce**

Minimal config change (everything else is the default `devenv init` output; only `processes.*` lines were added):

```nix
processes.foo.exec = "echo 'foo (parent)'";
processes.foo.ready.exec = "echo beep";
processes.foo.before = ["devenv:processes:bar"];
processes.bar.exec = "echo 'bar (child)'";
```

Steps:

1) Start `devenv` in detached mode:

```sh
devenv up --detach
```

2) Observe that each process runs twice (excerpt from `.devenv/processes.log`):

```text
[foo   ] Running tasks     devenv:processes:foo
[bar   ] Running tasks     devenv:processes:bar
[foo   ] bar (child)
[foo   ] bar (child)
[foo   ] foo (parent)
[foo   ] foo (parent)
[bar   ] bar (child)
[bar   ] bar (child)
[bar   ] foo (parent)
[bar   ] foo (parent)
```

3) Fetch the generated `process-compose` YAML:

```sh
process-compose project state
```

This prints a Nix store path like:

```text
File Names:
  - /nix/store/...-process-compose.yaml
```

4) Inspect that file and note `--mode all` is used for each process:

```sh
cat /nix/store/...-process-compose.yaml
```

Relevant excerpt (both `foo` and `bar` look like this):

```yaml
processes:
  bar:
    command: exec .../bin/devenv-tasks run --task-file .../tasks.json --mode all ... devenv:processes:bar
  foo:
    command: exec .../bin/devenv-tasks run --task-file .../tasks.json --mode all ... devenv:processes:foo
```

Expected:

- `process-compose` should run each process task once (parent before child).
- The generated command for each `process-compose` process should *not* use `--mode all` (it should only execute that process’s corresponding `devenv:processes:<name>` task).

Actual:

- `--mode all` is used, so each process runs all tasks; with two processes it results in duplicated execution/output.

Cleanup:

```sh
devenv processes stop
```

**Version**

```text
devenv 1.11.2 (aarch64-linux)
process-compose v1.87.0
nix (Nix) 2.31.3
Linux nixos 6.19.0 aarch64 GNU/Linux
```

domenkozar:
## Summary

Fixes #2514

- When processes use `before`/`after` dependencies with process-compose, each process-compose entry runs `devenv-tasks run --mode all devenv:processes:<name>`. `RunMode::All` traverses the full dependency graph, so each process ends up running other dependent processes too — causing duplication.
- Adds `--ignore-process-deps` CLI flag to `devenv-tasks` which excludes non-root process-type tasks from the scheduled subgraph. Non-process deps (migrations, setup scripts) still run normally.
- Process ordering is now handled by process-compose's native `depends_on`:
  - `after` deps are translated to `depends_on` in the `processType` submodule config
  - `before` deps are inverted to `depends_on` on target processes in `process-compose.nix`

## Test plan

- [x] `cargo build` passes
- [x] `cargo test` passes
- [ ] Manual test with the reproducer from #2514:
  ```nix
  processes.foo.exec = "echo 'foo (parent)'";
  processes.foo.ready.exec = "echo beep";
  processes.foo.before = ["devenv:processes:bar"];
  processes.bar.exec = "echo 'bar (child)'";
  ```
  Verify each process runs exactly once and in correct order.

### DIFF
--- a/docs/src/reference/options.md
+++ b/docs/src/reference/options.md
@@ -22441,7 +22441,8 @@ boolean
 
 Tasks that must be ready before this process starts.
 Use task names like “devenv:processes:postgres” or “myapp:setup”.
-Supports @ready (default) and @complete suffixes.
+Supports @started, @ready (default for processes), and @completed suffixes for process dependencies.
+Supports @started, @succeeded (default for tasks), and @completed suffixes for task dependencies.
 
 
 
@@ -22460,7 +22461,7 @@ list of string
 ```
 [
   "devenv:processes:postgres"
-  "myapp:migrations@complete"
+  "myapp:migrations@succeeded"
 ]
 ```
 
@@ -32774,10 +32775,12 @@ Here’s a helpful mnemonic to remember: This task runs *after* these tasks.
 
 You can append a suffix to control dependency behavior:
 
- - ` task ` or ` task@ready ` - wait for task to succeed (default, fails if dependency fails)
- - ` task@complete ` - wait for task to finish (soft dependency, does NOT propagate failure)
+ - ` task@started ` - wait for task to begin execution
+ - ` task ` or ` task@ready ` - wait for task to be ready/healthy (default for processes, processes only)
+ - ` task@succeeded ` - wait for task to exit successfully (default for tasks, tasks only)
+ - ` task@completed ` - wait for task to finish, regardless of exit code (soft dependency)
 
-Example: ` after = [ "pnpm:install@complete" ]; ` allows this task to run
+Example: ` after = [ "pnpm:install@completed" ]; ` allows this task to run
 even if pnpm:install fails.
 
 
@@ -32805,8 +32808,10 @@ Here’s a helpful mnemonic to remember: This task runs *before* these tasks.
 
 You can append a suffix to control dependency behavior:
 
- - ` task ` or ` task@ready ` - the dependent waits for this task to succeed (default)
- - ` task@complete ` - the dependent waits for this task to finish (soft dependency)
+ - ` task@started ` - the dependent waits for this task to begin execution
+ - ` task ` or ` task@ready ` - the dependent waits for this task to be ready/healthy (default for processes, processes only)
+ - ` task@succeeded ` - the dependent waits for this task to exit successfully (default for tasks, tasks only)
+ - ` task@completed ` - the dependent waits for this task to finish (soft dependency)
 
 
 


sebb3:
Ive just finished testing my actual setup with this fix now (running 2.0 main on darwin), and to summarize I believe it solves most of the issues.
My dependency chain is:
Mysql service -> migrations + seed task -> 15+ web services (processes)
+ A few other infra services (kafka, redis, etc.) 

The problem I had before the fix was that the 15 services at the end of the graph triggered a chain reaction that spawned multiple mysqld processes, that competed for the INNODB files = never-ending restart mess..

Now it seems like the task is waiting until the database comes online and doesnt spread its plague to mysqld at least. I'm still struggling with making all web services wait for ONE instance of the task to finish, they spawn one task each, but it's not continuing down the dependency chain at least.

I can work with this! Just need to come up with some locking mechanism (flock or something). Or do you have another recommendation for this scenario? I can see there's quite a lot of options on both the processes and tasks, but I'm struggling with making it work, so for now I'll just hack something together.

What is the current status of the native process-manager? I cant make it work at all (TUI freezes on start), so I guess I'll resort to process-compose until it's more stable...

Let me know if I can provide more information that could be of use!

domenkozar:
> Ive just finished testing my actual setup with this fix now (running 2.0 main on darwin), and to summarize I believe it solves most of the issues. My dependency chain is: Mysql service -> migrations + seed task -> 15+ web services (processes)
> 
> * A few other infra services (kafka, redis, etc.)
> 
> The problem I had before the fix was that the 15 services at the end of the graph triggered a chain reaction that spawned multiple mysqld processes, that competed for the INNODB files = never-ending restart mess..
> 
> Now it seems like the task is waiting until the database comes online and doesnt spread its plague to mysqld at least. I'm still struggling with making all web services wait for ONE instance of the task to finish, they spawn one task each, but it's not continuing down the dependency chain at least.
> 
> I can work with this! Just need to come up with some locking mechanism (flock or something). Or do you have another recommendation for this scenario? I can see there's quite a lot of options on both the processes and tasks, but I'm struggling with making it work, so for now I'll just hack something together.
> 
> What is the current status of the native process-manager? I cant make it work at all (TUI freezes on start), so I guess I'll resort to process-compose until it's more stable...
> 
> Let me know if I can provide more information that could be of use!

We've fixed one case where TUI would hang on macOS recently, did you try recent `devenv` binary from `main?

For getting inter-process dependencies to work you need native process manager :)

PR #2533, opened 2026-03-01
fix(fs): await watcher readiness instead of yield_now

domenkozar:
## Summary

- Uses watchexec's new `fs_ready` signal (from https://github.com/watchexec/watchexec/pull/1024) to wait for OS watches to be registered before returning from `FileWatcher::new()`
- Fixes a race condition where file changes immediately after watcher construction were missed because inotify/FSEvents watches weren't set up yet
- Adds `WatcherHandle::watch_async()` for runtime path additions that need the same readiness guarantee

## Test plan

- [x] All 7 `fs::tests` pass consistently across 5 runs
- [x] `test_multiple_files` and `test_handle_adds_path_at_runtime` (previously flaky) now pass reliably

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1525,6 +1525,7 @@ dependencies = [
  "tempfile",
  "tokio",
  "tracing",
+ "tracing-subscriber",
  "watchexec",
  "watchexec-filterer-globset",
 ]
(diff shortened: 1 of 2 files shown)


sandydoo:
Oh, nice! We should remove the old `watch` and take it's place then?

domenkozar:
pushed only the first commit to main, the rest isn't ready/helpful

Issue #2538, opened 2026-03-03
JavaScript failed on NPM because of nodejs-slim nixpkgs changes


shikanime:
**Describe the bug**

`enableNpm` option have been removed upstream https://github.com/NixOS/nixpkgs/commit/7459fe949f2905620de01ee4a16455d974f4b391

**To reproduce**

```nix
{
  languages.javascript.npm.enable = true;
}
```

```
❯ direnv reload
direnv: loading ~/[redacted]/.envrc
direnv: using flake . --accept-flake-config --no-pure-eval
direnv: nix-direnv: cache invalidated: files newer than cache:
nix-direnv: /Users/shikanimedeva/[redacted]/.envrc
/Users/shikanimedeva/[redacted]/flake.nix
/Users/shikanimedeva/[redacted]/flake.lock
error:
       … while calling the 'derivationStrict' builtin
         at <nix/derivation-internal.nix>:37:12:
           36|
           37|   strict = derivationStrict drvAttrs;
             |            ^
           38|

       … while evaluating derivation 'devenv-shell'
         whose name attribute is located at /nix/store/ky7q5vm0kgv6qbq5gfzmv1ki79qr4q0z-source/pkgs/stdenv/generic/make-derivation.nix:536:13

       … while evaluating attribute 'DEVENV_PROFILE' of derivation 'devenv-shell'

       … while evaluating the option `perSystem.aarch64-darwin.devenv.shells.default.packages':

       … while evaluating definitions from `/nix/store/zgbjmrhgvd8a42nyi1jck6v98gzvvql0-source/src/modules/languages/javascript.nix':

       … while evaluating the option `perSystem.aarch64-darwin.devenv.shells.default.languages.javascript.npm.package':

       (stack trace truncated; use '--show-trace' to show the full, detailed trace)

       error: function 'anonymous lambda' called with unexpected argument 'enableNpm'
       at /nix/store/ky7q5vm0kgv6qbq5gfzmv1ki79qr4q0z-source/pkgs/development/web/nodejs/v24.nix:1:1:
            1| {
             | ^
            2|   lib,
direnv: nix-direnv: Evaluating current devShell failed. Falling back to previous environment!
```

**Version**

```
devenv 1.11.2 (aarch64-darwin)
```

PR #2540, opened 2026-03-03
fix(javascript): enableNpm option have been removed from nixpkgs

shikanime:
Fixes: https://github.com/cachix/devenv/issues/2538
Signed-off-by: William Phetsinorath <william.phetsinorath@shikanime.studio>
Change-Id: I3f64d53a63ef26dd285c3575276f10686a6a6964

### DIFF
--- a/src/modules/languages/javascript.nix
+++ b/src/modules/languages/javascript.nix
@@ -260,9 +260,9 @@ in
       enable = lib.mkEnableOption "install npm";
       package = lib.mkOption {
         type = lib.types.package;
-        default = cfg.package.override {
-          enableNpm = true;
-        };
+        default =
+          # Nixpkgs releases 26.05 no longer expose enableNpm option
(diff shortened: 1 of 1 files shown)


domenkozar:
we need to check if it supports the parameter and still call it if it does to support old nixpkgs

shikanime:
@domenkozar done 👍

shikanime:
Thinking about it, npm should not even exist anymore at this point.

sandydoo:
Fixed in https://github.com/cachix/devenv/commit/6da2445b6000864bbce5f3bc8f51ec413f0b81dd.

@shikanime, sorry for sniping this one from you, but I needed a quickfix after bumping devenv-nixpkgs.

One issue with the PR here is that default args don't show up in the `args` you get from `.override`. This is common footgun we've encounteredd across our modules.

PR #2382, opened 2026-01-06
feat(languages): add lsp.enable and lsp.package options to all langua…

domenkozar:
…ge modules

Adds configurable LSP support to all language modules with:
- `lsp.enable` option (defaults to true) to control LSP package installation
- `lsp.package` option to customize which LSP server package to use

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2546, opened 2026-03-04
fix(rust): respect toolchain/channel for lsp server

eyJhb:
With the changes made in #2382 , rust-analyzer would always just get pulled from pkgs, which causes a mismatch between the cargo version and the rust-analyzer version.

These changes are untested, but I would like some input from @domenkozar if possible. In reality I'm more tempted to just remove the lsp options, since they don't make any sense in this case, as they are included by default :D


domenkozar:
I was going back and forth on this one, I wanted all languages to the the option defined even if it repeats so that it's not confusing not having the option. Alternative would be not to define those, leading to confusing why it doesn't exist.

PR #2547, opened 2026-03-05
Update options.md

jerivl:
Added missing description in list of git hooks.

### DIFF
--- a/docs/src/reference/options.md
+++ b/docs/src/reference/options.md
@@ -4649,6 +4649,8 @@ Prevent committing secret keys into your source code
 
 **` rome `**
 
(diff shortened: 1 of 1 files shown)


jerivl:
Also, `rome`  is also missing a description, but it seems the tool that is being sunset by its devs in favor of `biome`. I'm unsure what the policy is on depreciation is here, so I've included a description, but feel free to cherry-pick from any of these changes

domenkozar:
these are auto-generated files, you need to find option where it's defined in Nix and fix it there

sandydoo:
These docs are generated from the options defined in https://github.com/cachix/git-hooks.nix. @jerivl, you should open a PR against that repo. Thanks!

PR #2557, opened 2026-03-06
fix languages.rust.channel: include makeWrapper in mkAggregated import

### LINKED ISSUE #2556: Private API change in rust-overlay breaks devenv languages.rust.channel option
`languages.rust.channel = "nightly"` will break because of https://github.com/oxalica/rust-overlay/pull/251 .

The bug comes from here: https://github.com/cachix/devenv/blob/main/src/modules/languages/rust.nix#L318

PaulGrandperrin:
Added makeWrapper to the imported components in mkAggregated.

fixes #2556

### DIFF
--- a/src/modules/languages/rust.nix
+++ b/src/modules/languages/rust.nix
@@ -319,7 +319,7 @@ in
         # Import the mkAggregated function.
         # This symlinkJoins and patches the individual components.
         mkAggregated = import (rust-overlay + "/lib/mk-aggregated.nix") {
-          inherit (pkgs) lib stdenv symlinkJoin bash curl;
+          inherit (pkgs) lib stdenv symlinkJoin bash curl makeWrapper;
           inherit (pkgs.buildPackages) rustc;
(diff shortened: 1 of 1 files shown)


domenkozar:
We are going to need to add this optionally for older versions:

```
 │        … while evaluating the option `packages':
  │
  │        … while evaluating definitions from `/nix/store/93gzcrk85ssq2pfgkkx2bkiqx2f8ar2k-source/src/modules/languages/rust.nix':
  │
  │        … while evaluating the option `languages.rust.toolchainPackage':
  │
  │        (stack trace truncated; use '--show-trace' to show the full, detailed trace)
  │
  │
  │        at /nix/store/q9cssql5c8fbnvp2lb21gg0xa4hfvm5p-source/lib/mk-aggregated.nix:1:1:
  │             1| {
  │              | ^
  │             2|   lib,
  │
  │        error: function 'anonymous lambda' called with unexpected argument 'makeWrapper'
```

PR #2571, opened 2026-03-07
Emit typed exec_if_modified OTEL attrs

schickling:
This instruments the real `exec_if_modified` cache path in `devenv-tasks` so status spans can report pattern counts, matched-file counts, command-path checks, and evaluation time. The stats are collected in `devenv-tasks/src/task_cache.rs` and the span emission lives in `devenv-tasks/src/task_state.rs`. The implementation uses `otel-span emit` with an explicit OTLP JSON payload so numeric values land as typed `intValue` fields instead of `stringValue` attrs. I validated the pre-fix failure mode locally and switched the upstream implementation to the typed payload path to address it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1745,6 +1745,7 @@ dependencies = [
  "tokio-test",
  "tokio-util",
  "tracing",
+ "which 8.0.0",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


sandydoo:
The hand-rolled OTEL implementation is not suitable for upstreaming, but we add tracing to the actual function in #2703.

PR #2580, opened 2026-03-08
fix: use NIX_STACK_SIZE for TUI devenv and REPL threads

f64u:
`run_with_tui` spawns threads with `std::thread::spawn` (default 8MB stack) instead of using `NIX_STACK_SIZE` (64MB). Since `block_on` runs the future on the calling thread, Nix evaluation happens on these undersized threads.

Projects with deeply recursive Nix expressions (e.g. crate2nix `appliedCargoNix`) hit a stack overflow in TUI mode but not in non-TUI mode, where the main thread already uses `NIX_STACK_SIZE`.

Both the devenv TUI thread and the REPL thread now use `std::thread::Builder::new().stack_size(NIX_STACK_SIZE)`.

**Workaround**: `RUST_MIN_STACK=16777216 devenv shell`

### DIFF
--- a/devenv/src/main.rs
+++ b/devenv/src/main.rs
@@ -267,28 +267,32 @@ async fn run_with_tui(cli: Cli) -> Result<()> {
 
     // Devenv on background thread (own runtime with GC-registered workers)
     let shutdown_clone = shutdown.clone();
-    let devenv_thread = std::thread::spawn(move || {
-        build_gc_runtime().block_on(async {
-            // Don't race with shutdown - let run_devenv handle shutdown via cancellation token
-            // This ensures process cleanup happens before the future is dropped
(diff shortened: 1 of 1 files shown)


domenkozar:
Did you try devenv 2.0.3? That fixes it.

f64u:
I'm on 2.0.3. The fix in https://github.com/cachix/devenv/commit/78481366 applied NIX_STACK_SIZE to the main thread and tokio
  workers, but missed two std::thread::spawn call sites in run_with_tui:

  - https://github.com/cachix/devenv/blob/v2.0.3/devenv/src/main.rs#L270
  - https://github.com/cachix/devenv/blob/v2.0.3/devenv/src/main.rs#L369

  Both still use the default 8MB stack.

PR #2582, opened 2026-03-08
feat(rust): use developer-configured toolchain for `import`

muni-corn:
I noticed that `devenv` was using some stable version of Rust for building packages with `import`. I'm guessing it was the `nixpkgs` version.

This PR changes the behavior of `import` so that it uses the toolchain configured by developers via `languages.rust`.


domenkozar:
Is there a chance we can make the change without all the formatting changes? I currently fail to see what changed because the whole file is modified.

muni-corn:
of course! it's done.

PR #2594, opened 2026-03-10
fix: bump nix to include temproot fix for fetchToStore

domenkozar:
## Summary

- Bumps the nix input to `cachix/nix#fix/temproot-fetch-to-store` which adds a client-side `addTempRoot()` call for the non-cached code path in `fetchToStore`
- Fixes sporadic `path '/nix/store/...' is not valid` errors during evaluation on CI

## Root cause

When `fetchToStore` returns a newly added store path (non-cached), the only temp root is held by the daemon worker that handled the `addToStore` call. If the daemon connection is recycled during a long evaluation, that temp root is lost and auto-GC can collect the path before `derivationStrict` writes the `.drv` file referencing it.

The cached code path already had `addTempRoot()` (line 40 of `fetch-to-store.cc`), but the non-cached path was missing it. This specifically affects plain file paths like patches and source files that appear as opaque context elements in `inputSrcs`.

## Test plan

- [ ] CI examples that previously failed with "is not valid" errors should pass
- [ ] Once verified, merge `fix/temproot-fetch-to-store` into `devenv-2.32` in cachix/nix and switch this input URL back

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -142,16 +142,16 @@
         "nixpkgs-regression": []
       },
       "locked": {
-        "lastModified": 1772748357,
-        "narHash": "sha256-vtf03lfgQKNkPH9FdXdboBDS5DtFkXB8xRw5EBpuDas=",
+        "lastModified": 1773158757,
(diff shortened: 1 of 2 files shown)


domenkozar:
done in main

PR #2620, opened 2026-03-13
fix: kill child processes on exit

### LINKED ISSUE #2619: Child processes (postgres, redis, etc.) not killed on exit
## Problem

When exiting devenv (either via Ctrl+C or normal exit), child processes like postgres and redis are left running. This happens with both the process-compose backend and the native process manager.

After exiting, these services remain bound to their ports, which can cause conflicts on the next `devenv up` or leave unexpected resource usage.

## Steps to reproduce

1. Configure a `devenv.nix` with postgres and redis services enabled
2. Run `devenv up`
3. Exit with Ctrl+C (or close the terminal)
4. Check for orphaned processes:

```bash
ps aux | grep postgres
ps aux | grep redis
```

Both services are still running.

## Expected behavior

All child processes spawned by devenv should be terminated when devenv exits, regardless of the process manager backend used.

## Environment

- devenv 2.0.4 (v2.0.4 tag)
- macOS (aarch64-darwin)

## Notes

This is related to but distinct from #2608 (per-process shutdown signals). Even if the correct signal is sent, the child processes are not being cleaned up at all on exit in some cases.

sadjow:
## Summary

Fixes child processes (postgres, redis, etc.) being left running after `devenv up` exits or `devenv processes down` is called.

Two root causes addressed:

**1. Native process manager: wrapper script didn't `exec` the actual command**

The bash wrapper script ran the service command as a child process instead of `exec`'ing into it. When SIGTERM was sent, only bash was killed while the actual service (and its children) became orphaned.

Fix: use `exec` so the service process replaces bash and becomes the session leader, receiving signals directly.

**2. Process-compose backend: SIGTERM sent only to PID, not process group**

The `stop()` method sent SIGTERM to the process-compose PID only, not its process group. If process-compose didn't forward the signal, child services survived.

Fix: send SIGTERM to the process group first (falls back to PID if the group doesn't exist).

Closes #2619

## Test plan

- [x] Added `test_stop_kills_child_processes` integration test that spawns a parent process with a background child and verifies both are killed on stop
- [x] All existing process lifecycle tests pass (10/10)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Bug Fixes
 
-- Fixed child processes (postgres, redis, etc.) being left running after `devenv up` exits or `devenv processes down` is called. The wrapper script now uses `exec` so the service process receives signals directly, and the process-compose backend sends SIGTERM to the process group instead of just the PID ([#2619](https://github.com/cachix/devenv/issues/2619)).
+- Fixed child processes (postgres, redis, etc.) being left running after `devenv up` exits or `devenv processes down` is called. The native manager wrapper now forwards TERM/INT signals to the child process group, and the process-compose backend creates a proper process group for signaling ([#2619](https://github.com/cachix/devenv/issues/2619)).
 - Fixed secretspec prompting for secrets in non-interactive contexts like direnv.
 - Fixed `devenv search` showing truncated package names (e.g. `pkgs.` instead of `pkgs.ncdu`).
 - Fixed TUI hanging when the backend encounters an error in the PTY shell path (e.g. Nix evaluation failure).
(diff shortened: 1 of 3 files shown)


domenkozar:
Could you also update CHANGELOG.md?

sadjow:
@domenkozar Sure let me add it. Also, let me do a quick double-check. if it works locally.

sadjow:
@domenkozar the fix is QA'ed locally and working well!

sadjow:
@domenkozar CHANGELOG updated.

sadjow:
QA: Tested locally with a project running postgres, redis, mosquitto, sidekiq, and several other services. All child processes were properly terminated after exit. No orphaned processes remained.

PR #2635, opened 2026-03-18
feat: add a lot more tui functionality

samchouse:
This PR changes the native process manager to use PTY's instead of log streaming which allows more complex UIs (qr codes, colors, etc) to be rendered properly. It properly shuts processes down instead of them surviving. Adds support for providing input into a process, such as interacting with Expo. Fixes flickering issues.

needs https://github.com/cachix/iocraft/pull/1

**Disclaimer:** this PR was entirely done by AI but I've been using it locally without issue

<details>
<summary>More in depth summary by AI:</summary>

## Summary

- Move process rendering in the TUI closer to real terminal behavior instead of treating everything as plain appended logs
- Add direct keyboard interaction with selected processes from inside the TUI
- Improve how process output is shown in both collapsed previews and expanded views
- Make stop/restart flows more explicit and reusable for manually managed processes
- Tighten PTY lifecycle handling so interactive processes behave more reliably during resize, restart, and shutdown

### TUI process rendering

- Process activities now maintain terminal state so output that relies on ANSI control sequences, cursor movement, and screen updates is rendered more faithfully
- Richer output handling is used across process views instead of only relying on line-buffered logs, which makes active services feel less like static log streams
- Process status banners and lifecycle updates are integrated into the rendered output more cleanly so state transitions are easier to follow

### Interactive process input

- A selected process can now be focused for input from the TUI with `F12`
- Once focused, key presses are forwarded to the underlying PTY instead of being handled only as TUI navigation
- Input forwarding includes normal character input plus control, alt, navigation, and editing keys so terminal-driven tools can be interacted with more naturally
- The same focus/input model is carried into the expanded process view so interaction is consistent between overview and detail modes

### Expanded and collapsed process views

- The expanded process view now works against terminal-backed process state instead of only a plain-text log representation
- Expanded view behavior is improved for live output by keeping a follow mode and adapting better when content is still actively changing
- Expanded view selection/copy behavior is updated to work with the richer rendered content model
- Collapsed previews now better represent ANSI-heavy process output, including styled content, while still stripping destructive control behavior that would break the layout
- Preview rendering is width-aware, making the condensed process view more readable and less noisy

### Process lifecycle and controls

- Process lifecycle handling is expanded with a reusable `Stopped` state instead of treating manually stopped processes the same as never-started ones
- Stopping a process now moves through an explicit `Stopping` state before settling into `Stopped`, which makes the UI reflect that transition more clearly
- Restart flows now account for processes that are active, never started, or manually stopped, so TUI controls behave more predictably across states
- Manual stop/restart actions also preserve and reuse process activity state more cleanly instead of forcing everything through a fresh registration path

### PTY and process supervision changes

- PTY-backed process execution is used more directly, which gives the TUI better control over process I/O and terminal behavior
- PTY instances are resized when the terminal changes size so interactive programs can adapt properly to the current viewport
- Raw process output is captured separately so terminal state can be updated without losing the existing log stream behavior used elsewhere in the UI
- Stop/restart handling is tightened so PTY-backed processes are torn down and recreated more deliberately during file-change restarts and manual control actions
- Shutdown behavior is improved so child processes are cleaned up more consistently rather than being left behind by partial teardown behavior

### UI polish and behavior

- Main-view scrolling behavior is more deliberate, including page/home/end navigation support around the activity list
- Input focus and selected activity state are tracked separately, which helps distinguish between navigating the TUI and interacting with a live process
- Exit/final-render behavior is cleaned up so the TUI leaves the terminal in a better state and presents a cleaner final snapshot when appropriate
- tmux gets a lower frame-rate cap to reduce unnecessary redraw pressure in environments where higher refresh rates are less useful

</details>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1929,6 +1929,7 @@ version = "2.0.7"
 dependencies = [
  "async-priority-channel",
  "miette",
+ "nix 0.31.2",
  "notify",
  "reqwest 0.13.2",
  "tempfile",
@@ -1987,6 +1988,7 @@ version = "2.0.7"
 dependencies = [
  "async-trait",
  "caps",
+ "crossterm",
  "devenv-activity",
  "devenv-event-sources",
  "futures",
@@ -2117,6 +2119,7 @@ version = "2.0.7"
 dependencies = [
  "anyhow",
  "async-trait",
+ "avt",
  "base64",
  "chrono",
  "clap",
--- a/Cargo.nix
+++ b/Cargo.nix
@@ -6710,6 +6710,11 @@ rec {
             features = [ "fancy" ];
           }
           {
+            name = "nix";
+            packageId = "nix 0.31.2";
+            features = [ "fs" "process" "signal" ];
+          }
+          {
             name = "notify";
             packageId = "notify";
             optional = true;
@@ -6933,6 +6938,10 @@ rec {
             name = "caps";
             packageId = "caps";
             target = { target, features }: ("linux" == target."os" or null);
+          }
+          {
+            name = "crossterm";
+            packageId = "crossterm";
           }
           {
             name = "devenv-activity";
@@ -7445,6 +7454,10 @@ rec {
           {
             name = "async-trait";
             packageId = "async-trait";
+          }
+          {
+            name = "avt";
+            packageId = "avt";
           }
           {
             name = "base64";
--- a/devenv-activity/src/activity.rs
+++ b/devenv-activity/src/activity.rs
@@ -423,6 +423,17 @@ impl Activity {
         }
     }
 
+    pub fn raw_log(&self, data: Vec<u8>) {
+        let _guard = self.span.enter();
+        if matches!(self.activity_type, ActivityType::Process) {
+            send_activity_event(ActivityEvent::Process(Process::RawLog {
+                id: self.id,
+                data,
+                timestamp: Timestamp::now(),
+            }));
+        }
+    }
+
     /// Set process status (for Process activities only)
     pub fn set_status(&self, status: ProcessStatus) {
         let _guard = self.span.enter();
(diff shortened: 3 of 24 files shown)


domenkozar:
This is very welcome! High-level overview:

 - Processes now run in PTYs instead of piped log streams, so rich terminal output (colors, QR codes,
   ANSI UIs) renders correctly
  - Users can focus a process with F12 and send keyboard input to it (useful for interactive tools like
  Expo)
  - Expanded and collapsed process views are completely reworked to render from virtual terminal state
  rather than plain text logs
  - Process lifecycle gets explicit Stopped/Stopping states with cleaner shutdown that no longer leaves
  orphaned child processes
  
  I think it might be worth splitting this up. A decision we need to make if we really want PTY to be the default, because that complicates writing logs, which are now mixed with terminal codes.

We definitely should allow it at least as an option, like `processes.XXX.is_pty`

samchouse:
Sure, I can work on adding the option to use PTYs, but I think opt out is the better way to do it since it renders like the process expects it to. The previous one would lose a bunch of logs for mean (QR code in Expo, among other things) which I don't think is a great default experience.

If you let me know how you wanna split the PR into others I can also work on that.

domenkozar:
@sandydoo thoughts on making pty the default for processes?

samchouse:
@domenkozar I added the option and set it to default for now. The PR also no longer needs my fork of iocraft and has been updated to main.

domenkozar:
@samchouse are we still missing some changes from you? PTY at least

samchouse:
It took too long to merge and I don't want to take another go at it with all the merge conflicts now. Feel free to take what you can from it though, it was working well on the old version.

domenkozar:
Yeah, I understand. Thanks for your efforts, I'll take a look at PTY since I want to support that (probably as an option not by default).

PR #2639, opened 2026-03-19
Implement support for allowUnsupportedSystem

tarc:
This PR adds a configuration option `allowUnsupportedSystem` to fix issues like this one from the devenv-nixpkgs repo: https://github.com/cachix/devenv-nixpkgs/issues/21

I'm not sure if this is the right approach.


domenkozar:
Thanks! Needs to be added to reference docs for devenv.yaml and noted it's added in 2.0.5

then update CHANGELOG.md

PR #2643, opened 2026-03-20
fix: invalidate shell eval cache when devenv config changes

hencjo:
### Human bug report

My changes in devenv.nix did not reflect in the build environment.
This happened both in `direnv activation. and `devenv shell`.

Changes to these attributes in `devenv.nix` didn't have effect on my shell:
* `languages` (like adding gradle; and it not ending up on PATH).
* ENVs not apperaing in shell.
* scripts not appearing on path.
* packages not being available.

This PR fixes this, but I have no idea if it's a desired approach. Please complete it with taste :D.

### Everything that follows is AI-generated

This was caused by stale shell evaluation cache entries being reused after project configuration changed.

The key symptom was that different commands disagreed about the current configuration:
- `devenv info` showed the updated packages and environment
- `devenv print-dev-env` could still return an older cached shell environment

Because `devenv shell` and `direnv` activation both consume the shell environment output, they inherited the stale result as well. That made the issue look like a broad `languages` problem, but the language modules were not the root cause. The same failure mode affected top-level `packages`, `env`, and generated `scripts`.

The underlying issue was incomplete cache invalidation for shell environment evaluation. Changes to core project configuration files such as `devenv.nix` could leave an existing shell cache entry looking valid when it was no longer correct.

### Fix approach

This PR fixes the problem by explicitly tracking core project configuration files as shell eval-cache inputs.

The shell cache now watches the project files that directly define the environment, including `devenv.nix`, `devenv.yaml`, `devenv.lock`, and local override files when they exist.

This keeps the change focused on cache correctness instead of adding special handling for individual features like `languages`, `env`, or `scripts`.

### Why this approach

The failure was not specific to any one module. It was a generic shell-environment caching bug. Fixing cache input tracking addresses the common cause and keeps `devenv info`, `devenv print-dev-env`, `devenv shell`, and `direnv` activation consistent with each other.

### Verification

The change is covered by focused tests for:
- the shared shell cache key versioning
- explicit tracking of core project config files

This is intended to be a low-maintenance fix: it hardens the cache boundary directly, rather than adding feature-specific behavior or a large end-to-end regression test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug Fixes
 
+- Fixed stale shell eval cache entries causing `devenv shell` and `direnv` activation to ignore changes to `devenv.nix` config such as `languages`, `env`, `packages`, and `scripts` ([#2643](https://github.com/cachix/devenv/pull/2643)).
 - Fixed services failing to start with E2BIG when the shell environment is large by moving env/cwd/stdio setup from the bash wrapper script to the spawned process directly ([#2638](https://github.com/cachix/devenv/issues/2638)).
 - Fixed processes failing to start when exported bash functions are present in the environment ([#2587](https://github.com/cachix/devenv/issues/2587)).
 - Fixed eval cache storing inconsistent port allocations across different cached attributes ([#2631](https://github.com/cachix/devenv/issues/2631)).


devenv-core/src/nix_backend.rs
```diff
@@ -12,6 +12,24 @@ use std::path::{Path, PathBuf};
 use crate::config::Input;
 use crate::nix_args::NixArgs;
 
+/// Version suffix for eval-cache keys.
+///
+/// Bump this when the cache key or tracked inputs change in a way that should
+/// invalidate previously stored rows.
+pub const EVAL_CACHE_KEY_VERSION: &str = "v2";
```
domenkozar: I don't think we need this, we already include cli version in the nix_args

domenkozar:
Looks good, could you also add an entry to `CHANGELOG.md`?

hencjo:
Thank you. Will do!

Issue #2488, opened 2026-02-11
Support for directory-based auto-activation (direnv shell hook equivalent)


schickling:
## Problem

A key feature of direnv is its shell hook (`eval "$(direnv hook fish)"` etc.) which automatically activates/deactivates environments when `cd`-ing into/out of project directories. This is essential for seamless developer experience — the environment just works without manually running any command.

As devenv moves toward replacing direnv (#2061), an equivalent mechanism is needed: automatic environment activation/deactivation triggered by directory changes.

## Expected behavior

Devenv should provide a shell hook (for bash, zsh, and fish) that:
1. Detects when the user enters a directory with a `devenv.nix` / `devenv.yaml`
2. Automatically activates the devenv environment (ideally with background evaluation, as noted in #2061)
3. Deactivates/restores the previous environment when leaving the directory
4. Works with nested projects (inner project takes precedence)

This is the main remaining blocker for fully replacing direnv with devenv.

## Related

- #2061 — Replace direnv with Rust
- #2177 — Activate profiles automatically based on folder location

---
*This issue was created on behalf of @schickling*

domenkozar:
To implement this we need something like `directory-auto-activation = true` so we know which devenvs should just enter in case you have a monorepo.

PR #2644, opened 2026-03-20
feat: add --with-direnv to devenv init as opt-in; restore init/.envrc

AmentumStudio:
Adding back (as opt-in either, via flag `--with-devenv` or env variable `DEVENV_DIRENV_USE=true`) support for `devenv init` creating `.envrc`.

### DIFF
--- a/devenv/init/.envrc
+++ b/devenv/init/.envrc
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+export DIRENV_WARN_TIMEOUT=20s
+
+eval "$(devenv direnvrc)"
+
+# `use devenv` supports the same options as the `devenv shell` command.
+#
+# To silence all output, use `--quiet`.
+#
+# Example usage: use devenv --quiet --impure --option services.postgres.enable:bool true
+use devenv
--- a/devenv/src/cli.rs
+++ b/devenv/src/cli.rs
@@ -459,7 +459,21 @@ impl Cli {
 #[derive(Subcommand, Clone)]
 pub enum Commands {
     #[command(about = "Scaffold devenv.yaml, devenv.nix, and .gitignore.")]
-    Init { target: Option<PathBuf> },
+    Init {
+        #[arg(
+            long,
+            num_args=0..=1,
+            value_name = "ENABLE",
+            default_value = "false",
+            default_missing_value = "true",
+            env = "DEVENV_DIRENV_INIT",
+            help = "For use with direnv, additionally scaffold .envrc.",
+            long_help = "For use with direnv, additionally scaffold .envrc.\n\nhttps://devenv.sh/integrations/direnv/",
+
+        )]
+        with_direnv: Option<bool>,
+        target: Option<PathBuf>,
+    },
 
     #[command(about = "Generate devenv.yaml and devenv.nix using AI")]
     Generate,
--- a/devenv/src/devenv/mod.rs
+++ b/devenv/src/devenv/mod.rs
@@ -49,6 +49,7 @@ const REQUIRED_FILES: [(&str, &str); 3] = [
     ("devenv.yaml", "devenv.yaml"),
     ("gitignore", ".gitignore"), // source name -> target name
 ];
+const OPTIONAL_DIRENV_REQUIRED_FILES: [(&str, &str); 1] = [(".envrc", ".envrc")];
 const EXISTING_REQUIRED_FILES: [&str; 1] = [".gitignore"];
 const PROJECT_DIR: Dir = include_dir!("$CARGO_MANIFEST_DIR/init");
 pub static DIRENVRC: Lazy<String> = Lazy::new(|| {
@@ -559,7 +560,7 @@ impl Devenv {
         ))
     }
 
-    pub fn init(&self, target: &Option<PathBuf>) -> Result<()> {
+    pub fn init(&self, target: &Option<PathBuf>, with_direnv: &Option<bool>) -> Result<()> {
         let target = target.clone().unwrap_or_else(|| {
             std::fs::canonicalize(".").expect("Failed to get current directory")
         });
@@ -569,7 +570,14 @@ impl Devenv {
             std::fs::create_dir_all(&target).expect("Failed to create target directory");
         }
 
-        for (source_name, target_name) in REQUIRED_FILES {
+        for (source_name, target_name) in
+            REQUIRED_FILES
+                .iter()
+                .chain(match with_direnv.is_some_and(|x| x) {
+                    true => OPTIONAL_DIRENV_REQUIRED_FILES.iter(),
+                    false => [].iter(),
+                })
+        {
             info!(devenv.is_user_message = true, "Creating {}", target_name);
 
             let path = PROJECT_DIR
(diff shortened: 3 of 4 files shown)


AmentumStudio:
Created as draft to get your approval on overall approach.
When all is fine, will add docs / changelog / maybe a test or two.

devenv/init/.envrc
```diff
@@ -0,0 +1,12 @@
...
+export DIRENV_WARN_TIMEOUT=20s
+
+eval "$(devenv direnvrc)"
+
+# `use devenv` supports the same options as the `devenv shell` command.
+#
+# To silence all output, use `--quiet`.
+#
+# Example usage: use devenv --quiet --impure --option services.postgres.enable:bool true
+use devenv
```
AmentumStudio, line 12: This file is verbatim the same as the one removed in https://github.com/cachix/devenv/commit/087763b94aba6d69cf6731ae2aa5bd927eefc73a . This PR's goal is to (for those who want) restore the "old" behaviour of `devenv init`, so found it only fitting to just restore the old file.

devenv/src/devenv/mod.rs
```diff
@@ -560,7 +561,14 @@ impl Devenv {
             std::fs::create_dir_all(&target).expect("Failed to create target directory");
         }
 
-        for (source_name, target_name) in REQUIRED_FILES {
+        for (source_name, target_name) in
+            REQUIRED_FILES
+                .iter()
+                .chain(match with_direnv.is_some_and(|x| x) {
+                    true => OPTIONAL_DIRENV_REQUIRED_FILES.iter(),
+                    false => [].iter(),
```
AmentumStudio, line 578: I expect this to be a no-op in Release builds, but haven't checked binary.

devenv/src/cli.rs
```diff
@@ -458,7 +458,21 @@ impl Cli {
...
     #[command(about = "Scaffold devenv.yaml, devenv.nix, and .gitignore.")]
-    Init { target: Option<PathBuf> },
+    Init {
+        #[arg(
+            long,
+            num_args=0..=1,
+            value_name = "ENABLE",
+            default_value = "false",
+            default_missing_value = "true",
+            env = "DEVENV_DIRENV_USE",
```
AmentumStudio: I explicitely wanted for the user to be able to do all of:

| Command  | `.envrc` created? |
| - | - |
| `devenv init` | `false` |
| `devenv init --with-direnv=false` | `false` |
| `devenv init --with-direnv=true` | `true` |
| `devenv init --with-direnv` | `true` |

so that they don't necessarily _have_ to know what the default is, and don't be caught unawares if/when that default changes (i.e. if they never want `.envrc`, just provide `--with-direnv=false` in your script).

Ofc everything that comes with `clap`'s environment support is also there, so they can as well do `DEVENV_DIRENV_USE=false` or `DEVENV_DIRENV_USE=true` in their `~/.bashrc`.

Please do say, whether this env variable is named to your liking, was not sure what your convention is.

devenv/src/devenv/mod.rs
AmentumStudio: Eh I got too curious, and at least according to https://rust.godbolt.org/z/q411vKbzM
the only real difference is additional jump (`lea`) and the fact that we now store some additional strings in `.data` section.

devenv/src/cli.rs
domenkozar: It should be called `DEVENV_DIRENV_INIT` because it's about creating the file, not using it.

domenkozar:
@AmentumStudio could I ask you why do you prefer to use direnv? Maybe we should fix that instead :)

devenv/src/cli.rs
AmentumStudio: Makes sense. Corrected.

AmentumStudio:
> @AmentumStudio could I ask you why do you prefer to use direnv? Maybe we should fix that instead :)

As I've written on cachix discord, biggest blockers on using "new" `devenv shell` are:
- [ ] https://github.com/cachix/devenv/issues/36
- [ ] https://github.com/cachix/devenv/issues/2488

I understood your [reply](https://discord.com/channels/1036369714731036712/1036369758024650792/1484165526388473926) as encourangement for this PR.

If instead you've meant
> we should add that [directory-based auto-activation in `devenv shell`]!
> I'm happy to review a PR [which introduces directory-based auto-activation in `devenv shell`]

then apologies for confusion, but I might not have enough resources and determination to implement all that.

I appreciate your work on "new" `devenv shell` (not sure how the proper name for it is, "native" `devenv shell`?), but as it is right now, it feels too bleeding edge for mundane day-to-day work.

AmentumStudio:
@domenkozar I'd just want to know whether to continue working on this PR
if this is not something you want merged, I will just close

rob-mur:
> @AmentumStudio could I ask you why do you prefer to use direnv? Maybe we should fix that instead :)

Just to comment in my two cents on why I prefer using direnv - it broadly comes down to two use cases. Feel free to ignore them/take them into account as you prefer :)

## CI

I prefer not to litter all of my ci command with devenv shell... and I notice that if I have built a container from my shell environment to use as a CI cache image, then devenv shell forces a recalculation/rebuild of the environment (which wastes time in CI) vs activating the direnv being almost instant and then for the entire rest of the CI pipeline I can just use my environment exactly the same as locally, no devenv commands required.

## Shell exposure

When developing locally I have a lot of nice-ities I have built into my zsh setup that I am used to using, devenv shell (at least at the time I last checked) breaks this as it isn't supported - I'm also not sure if even when it were to be supported that it would have exactly the same behaviour. There's a bit of a worry about being locked into a TUI rather than something that is imply 'just the shell with the environment carefully controlled' if it's clear what I mean?

sandydoo:
> I notice that if I have built a container from my shell environment to use as a CI cache image, then devenv shell forces a recalculation/rebuild of the environment (which wastes time in CI) vs activating the direnv being almost instant 

@rob-mur this is worth opening an issue for

domenkozar:
> > @AmentumStudio could I ask you why do you prefer to use direnv? Maybe we should fix that instead :)
> 
> As I've written on cachix discord, biggest blockers on using "new" `devenv shell` are:
> 
> * [ ]  [zsh support #36](https://github.com/cachix/devenv/issues/36)
> * [ ]  [Support for directory-based auto-activation (direnv shell hook equivalent) #2488](https://github.com/cachix/devenv/issues/2488)
> 
> I understood your [reply](https://discord.com/channels/1036369714731036712/1036369758024650792/1484165526388473926) as encourangement for this PR.
> 
> If instead you've meant
> 
> > we should add that [directory-based auto-activation in `devenv shell`]!
> > I'm happy to review a PR [which introduces directory-based auto-activation in `devenv shell`]
> 
> then apologies for confusion, but I might not have enough resources and determination to implement all that.
> 
> I appreciate your work on "new" `devenv shell` (not sure how the proper name for it is, "native" `devenv shell`?), but as it is right now, it feels too bleeding edge for mundane day-to-day work.

https://github.com/cachix/devenv/issues/2488 is fixed on main

#36 I have a prototype ready, will ask for testing soon :)

domenkozar:
I'm closing this for now, I'm happy to improve https://devenv.sh/integrations/direnv/ to better explain how to use direnv if you prefer so

PR #2645, opened 2026-03-20
Fix transient strict-port conflicts on restart and improve shutdown visibility

sadjow:
## Summary

This improves the native process manager shutdown flow and makes strict port reservation more reliable during immediate restarts.

On macOS, `devenv up` could fail right after `Ctrl-C` with an error like:

`Port 6380 is already in use. Use --strict-ports=false to auto-allocate an available port.` even when the previous process was in the middle of shutting down and no real listener remained. This change addresses that case and makes shutdown progress more visible while processes are terminating.

## Changes

- wait for declared process ports to be released before marking a process as stopped
- improve shutdown output so users can see that termination is still in progress
- stop processes concurrently instead of serially during shutdown
- retry strict exact-port reservation when `EADDRINUSE` appears transient and no owning PID is visible

## Why

PR #2640 fixed an IPv6-related false positive, but local QA showed a remaining restart failure on the IPv4/macOS path during the `Ctrl-C` -> `devenv up` loop (even after minutes). This change covers that remaining transient conflict window.

## QA

Tested locally with the forked binary against project with multiple processes and services.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Fixed eval cache storing inconsistent port allocations across different cached attributes ([#2631](https://github.com/cachix/devenv/issues/2631)).
 - Fixed stale eval cache invalidation for `devenv up` process config changes caused by overlapping evaluations clearing each other's file dependency observers ([#2632](https://github.com/cachix/devenv/pull/2632)).
 - Fixed false-positive `--strict-ports` failures on macOS when IPv6 loopback binding returns a benign error after a clean shutdown and cache invalidation ([#2640](https://github.com/cachix/devenv/pull/2640)).
+- Fixed transient `--strict-ports` failures on immediate restart when a port briefly reports `EADDRINUSE` after shutdown despite having no visible owning process.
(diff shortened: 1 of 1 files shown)


domenkozar:
This is getting unessarily complex, can you destill down to what the fix it?

sadjow:
@domenkozar the issue is that devenv shutdown is not releasing the process port postgres and redis. And there's no visibility around this. After we quit the devenv, it immediately disappeared from the screen (there's no feedback on what is happening). And even after some minutes, the port is still allocating redis's port.

image

For instance, the Redis one is not easily releasing the ports we need to await and perhaps add some feedback in the UI, see image. This PR adds this feature to display that we are releasing Redis ports and checking if the ports were released. You can try it out. But perhaps we can narrow down this PR to just await the ports to get released? Is this happening on your side? The native process manager is not stable at all regarding this port allocation/release.

image

sadjow:
The fix part of this PR is to await the ports to get released and improve the flow.

The "feature" part is to improve some logging/visibility. around this . (Redis awaiting port below, while postgres already released)
image

Redis terminated)
image

sadjow:
> This is getting unessarily complex, can you destill down to what the fix it?

@domenkozar I will be separate into two PRs (the fix one) and the proposal for the logs/output in another.

PR #2647, opened 2026-03-20
fix: await port release on process stop to prevent strict-port restart failures

sadjow:
## Summary
Strict-port restarts fail with "port already in use" when the kernel hasn't released the socket yet after process shutdown. This adds a retry loop that waits for ownerless ports to become available, and shows a "stopping" status in the TUI during the graceful shutdown window.

## Why
After stopping a process the port can stay in `EADDRINUSE` for a few seconds even though no process owns it anymore. In strict mode, the immediate restart attempt sees this and gives up. The fix distinguishes ownerless transient conflicts from real ones and retries accordingly.

## Testing
- `cargo nextest run -p devenv-activity -p devenv-processes -p devenv-tui`
- `nix build .#devenv`
- Manual QA with native manager and Redis churn/restart

### DIFF
--- a/devenv-activity/src/events.rs
+++ b/devenv-activity/src/events.rs
@@ -351,6 +351,8 @@ pub enum ProcessStatus {
     Ready,
     /// Stop + start cycle in progress.
     Restarting,
+    /// Graceful shutdown in progress (SIGTERM sent, waiting for exit and port release).
+    Stopping,
     /// Exited.
     Stopped,
 }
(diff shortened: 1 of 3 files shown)


sadjow:
@domenkozar this PR contains the focused fix. I'm testing this locally, and after several quits and devenv ups, no port conflicts.

sadjow:
@domenkozar this one makes the port release more reliable; it is not only about waiting and displaying that the process has stopped. But ensure to not have conflicts when starting again. Restarting the process manager is an everyday task, and we should not have friction. This improves the DX, where we are still required to manually kill the processes before starting devenv again.

The image below shows that now native process manager is now awaiting to show that the processes are stopped.

Also, before this fix the redis process was taking up to 15 seconds to stop and now it stops almost immediatelly.

image

sadjow:
@domenkozar update the PR title and description. I wanted also to work on the stop specific process feature on top of the new "stopping" state.

sadjow:
Please feel free to let me know any thoughts.

PR #2656, opened 2026-03-23
Add Ctrl+X to stop individual processes from the TUI

sadjow:
## Summary

Adds the ability to stop a specific process from the TUI by pressing Ctrl+X when a process is selected. The process stays visible with a "stopped" status and can be restarted with Ctrl+R.

- Introduces `stop_and_keep()` on the native process manager, which gracefully stops the process (SIGTERM + port release) but transitions it to `NotStarted` instead of removing it, preserving the Activity so it remains in the TUI
- Adds `ProcessCommand::Stop` variant and handles it in the command listener
- Adds `ProcessStatus::is_stoppable()` so the keybinding and help bar only appear for processes that are actually running (not Waiting or already Stopping)
- Notifies the task system after stopping so dependency tracking stays consistent

## Test plan

- [x] Start `devenv up` with a process, select it with arrow keys, press Ctrl+X
- [x] Verify TUI shows "stopped" status and process count decreases
- [x] Press Ctrl+R to restart, verify process comes back
- [x] Verify Ctrl+X hint only shows for running processes (not for waiting/stopped)
- [x] Verify Ctrl+C full shutdown still works correctly

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Improvements
 
+- Added Ctrl+X keybinding to stop individual processes from the TUI while keeping them visible and restartable.
 - Added `nixpkgs.rocmSupport` option to enable ROCm support in nixpkgs configuration.
 
 ## 2.0.6 (2026-03-22)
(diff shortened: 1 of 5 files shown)


domenkozar:
let's use ctrl-x since that will conflict less with terminals (ctrl-s is `XON/XOFF`)

PR #2658, opened 2026-03-24
Auto-reload shell environment instead of requiring Ctrl-Alt-R

### LINKED ISSUE #2595: Unable to trigger shell reload using Ctrl+Alt-R combination on MacOS
**Describe the bug**

Any keypress ordering of the combination just ends up triggering `reverse-i-search` via `Ctrl+R`.

**To reproduce**

1. Start a devenv shell
2. Make change to devenv configuration
3. Wait for evaluation to complete
4. Attempt to trigger reload

**Version**

`devenv 2.0.3+2105b1a (aarch64-darwin)`

domenkozar:
## Summary

- Shell environment now auto-reloads at the next prompt via `PROMPT_COMMAND` instead of requiring the user to press Ctrl-Alt-R
- Status bar shows "reloaded" after the environment is applied, instead of prompting with a keybind hint
- Ctrl-Alt-R keybind still works for immediate reload, just no longer advertised

Fixes #2595.

## Test plan

- [ ] Run `devenv shell`, edit `devenv.nix`, verify environment reloads automatically at next prompt
- [ ] Verify status bar shows "building..." then "ready" then "reloaded" transitions
- [ ] Verify Ctrl-Alt-R still works for immediate reload
- [ ] `cargo nextest run -p devenv-shell` passes (131 tests, snapshot updated)


AmentumStudio:
Changing `.name` in `devenv.nix` triggers rebuild, but one needs to exit shell and enter shell again to see changes.
(I see `ready`, but never `reloaded`).

EDIT: This happens with https://github.com/mcdonc/devenv-zsh turned on (it actually does use `zsh` with `devenv shell` v2 now).

AmentumStudio:
Adding random package:
`packages = [ pkgs.gost ];`
Triggers rebuild, there is `ready`, never `reloaded`, `which gost` results in `gost not found`.
I need to exit shell and run `nix run github:cachix/devenv/auto-reload-shell shell` for `which gost` to return the path.

EDIT: This happens with https://github.com/mcdonc/devenv-zsh turned on (it actually does use `zsh` with `devenv shell` v2 now).

AmentumStudio:
If I don't use `devenv-zsh` (and use `bash` as result), I need to press `ENTER` once there appears `ready` in the bottom of TUI, then it changes to `reloaded`.
But once it is rebuilt and `relodaded`, changing anything else in `devenv.nix` does not trigger any rebuilds.
It behaves as if it only listens for first change, then stops listenging.

euphemism:
Is it possible to support both auto-reload and manual triggering? i.e. introducing a config option for the manual keybind?

domenkozar:
> Is it possible to support both auto-reload and manual triggering? i.e. introducing a config option for the manual keybind?

You'd prefer manual reload?

euphemism:
> > Is it possible to support both auto-reload and manual triggering? i.e. introducing a config option for the manual keybind?
> 
> You'd prefer manual reload?

In some scenarios, possibly. My main issue is that as currently implemented I simply can't use the feature.

domenkozar:
> Changing `.name` in `devenv.nix` triggers rebuild, but one needs to exit shell and enter shell again to see changes. (I see `ready`, but never `reloaded`).
> 
> EDIT: This happens with https://github.com/mcdonc/devenv-zsh turned on (it actually does use `zsh` with `devenv shell` v2 now).

It should reload if you press enter. Can you try again?

domenkozar:
I've tested and this works now, addressing the issue of manually reloading the shell. We can still introduce possibility to bind it to a key later on, as this matches how direnv worked.

PR #2660, opened 2026-03-24
Fix process watch.paths not triggering restarts

### LINKED ISSUE #2657: devenv processes not restarting on file change
**Describe the bug**
The devenv `watch` feature for processes is not triggering a restart when a file changes. I believe this might be due to some weird interaction with being installed through `home-manager`.

home-manager config: https://github.com/rob0rt/nix-config/blob/main/home.nix#L18

example `devenv.nix` file:
```nix

{
  packages = [ pkgs.tailwindcss_4 ];

  languages.rust = {
    enable = true;
    channel = "stable";
  };

  dotenv.enable = true;

  processes = {
    tailwind = {
      exec = "${pkgs.tailwindcss_4}/bin/tailwindcss -i ./assets/app.css -o ./static/styles.css --watch";
      cwd = "./braid";
    };

    api = {
      exec = "cargo run --package braid";
      watch = {
        paths = [ ./braid/src ./braid/Cargo.toml ./braid/templates ];
      };
      after = [ "devenv:processes:tailwind@started" ];
    };
  };
}
```

I've observed this behavior on x86_64 linux and darwin (MacOS).

Maybe a smoking gun, the line at the bottom of dieenv shell shows  `✓ devenv ready in 4ms | watching 6 files` - should that be more?

**To reproduce**
* Install devenv with home-manager
* run devenv shell
* run devenv up
* modify any file in processes.api.watch.paths
* nothing happens

**Version**
devenv 2.0.6 (x86_64-linux)

domenkozar:
## Summary

- Fixed process `watch.paths` not triggering restarts because Nix paths were serialized as Nix store paths instead of source directory paths
- Removed dead code from native.nix (`nativeConfigJson`, `processConfig`, redundant type definitions) that was superseded by the task system

The root cause: `watch.paths` has type `types.listOf types.path`. When serialized to JSON via `pkgs.formats.json`, `builtins.toJSON` converts Nix path values to store paths (e.g., `/nix/store/abc-src`). The file watcher then monitors immutable store paths, so user edits are never detected.

The fix applies `map toString` before serialization to preserve the original absolute source paths. This is the same pattern that was already used in `nativeConfigJson` (which turned out to be dead code, since the native process manager reads configs from `task.config` via the task system, not from `nativeConfigJson`).

Fixes #2657

## Test plan

- [ ] Verify `devenv up` with `watch.paths` configured detects file changes and restarts the process
- [ ] Verify existing process tests still pass (`cargo nextest run -p devenv-processes` passes all 114 tests)
- [ ] Verify Nix evaluation succeeds after removing dead `nativeConfigJson` option


domenkozar:
Superseded by #2715, which includes the same fix plus a type change in tasks.nix and an integration test.

PR #2661, opened 2026-03-24
Add task messages for post-init shell output

### LINKED ISSUE #2500: Feature: post-drain hook for displaying messages after shell initialization
## Problem

`enterShell` code that prints to stdout/stderr (e.g. `echo`, `printf`) is consumed by the PTY drain phase (`drain_pty_to_vt` in `devenv-shell/src/task_runner.rs`) and never shown to the user in the interactive session. This makes it impossible for `enterShell` scripts to display informational messages after shell initialization.

Workarounds like `PROMPT_COMMAND` or writing to `/dev/tty` are fragile and get swallowed by the VT drain as well.

Related: #2407 covers task *output* not being visible in the TUI. This issue is specifically about displaying messages to the user *after* the TUI exits and the interactive shell starts.

## Use Case

We set up OpenTelemetry tracing during `enterShell` and want to display a clickable trace URL (via OSC 8 hyperlink) when the user enters `devenv shell`. The trace URL is only known at shell-init time, so it can't be baked into a static config.

## Proposed Solutions

Any of these would work (in rough order of simplicity):

### Option A: Environment variable message channel

After PTY drain completes, devenv checks an env var like `__DEVENV_POST_INIT_MESSAGE` and prints its contents to the terminal:

```bash
# in enterShell
export __DEVENV_POST_INIT_MESSAGE="Trace: https://jaeger.example.com/trace/abc123"
```

### Option B: File-based message channel

After PTY drain completes, devenv checks for a file like `$DEVENV_DOTFILE/post-init-message`, prints its contents, and deletes it:

```bash
# in enterShell
echo "Trace: https://jaeger.example.com/trace/abc123" > "$DEVENV_DOTFILE/post-init-message"
```

### Option C: New ShellCommand variant

Add a `ShowMessage(String)` variant to `ShellCommand` (similar to how `PrintWatchedFiles` works via the status line system), allowing Rust-side code to emit messages after drain.

---

> [!NOTE]
> Filed on behalf of @schickling via Claude Code

domenkozar:
## Summary

- Tasks can now display messages when entering the shell by writing `{"devenv":{"messages":["..."]}}` to `$DEVENV_TASK_OUTPUT_FILE`
- Follows the same pattern as task exports (`devenv.env`): collected from task outputs, stored on the Devenv struct, injected as `printf` statements into the shell script
- Messages are excluded from `direnv export` to avoid corrupting env var output

Closes #2500

## Test plan

- [ ] Create a task with `before = [ "devenv:enterShell" ]` that writes messages to `$DEVENV_TASK_OUTPUT_FILE` and verify they appear after `devenv shell` init
- [ ] Verify `devenv shell -- echo test` (non PTY path) also prints messages
- [ ] Verify `direnv export` does not include message output
- [ ] Verify cached/skipped tasks still produce their messages

### DIFF
--- a/devenv/src/devenv/mod.rs
+++ b/devenv/src/devenv/mod.rs
@@ -232,6 +232,10 @@ pub struct Devenv {
     // run_enter_shell_tasks(). Injected into the bash script by prepare_shell()
     // so they take effect AFTER the Nix shell env is applied.
     task_exports: std::sync::Mutex<BTreeMap<String, String>>,
+
+    // Task messages to display when entering the shell, set by
+    // run_enter_shell_tasks(). Injected as echo statements into the bash script.
+    task_messages: std::sync::Mutex<Vec<String>>,
 }
 
 /// Sanitize profile name to be filesystem-safe
@@ -415,6 +419,7 @@ impl Devenv {
             native_process_manager: Arc::new(OnceCell::new()),
             shutdown: options.shutdown,
             task_exports: std::sync::Mutex::new(BTreeMap::new()),
+            task_messages: std::sync::Mutex::new(Vec::new()),
         }
     }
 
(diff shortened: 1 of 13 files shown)


devenv/src/devenv/mod.rs
```diff
@@ -1162,6 +1178,18 @@ impl Devenv {
         result
     }
 
+    /// Format task messages as bash `printf` statements.
+    pub fn format_task_messages_bash(messages: &[String]) -> String {
+        use std::borrow::Cow;
+        let mut result = String::with_capacity(messages.len() * 40);
+        for msg in messages {
+            result.push_str("printf '%s\\n' ");
```
domenkozar: this should be part of shell dialect

PR #2678, opened 2026-03-28
switching nix profile add to install

Triscal:
After installing nix (I installed via lix) running `nix profile add` 

errors out with: 
```
error: 'add' is not a recognised command
Try 'nix --help' for more information.
```

But `install` works. See https://nix.dev/manual/nix/2.26/command-ref/new-cli/nix3-profile-install

I didn't check other pages, just https://devenv.sh/getting-started/

### DIFF
--- a/docs/src/getting-started.md
+++ b/docs/src/getting-started.md
@@ -40,7 +40,7 @@
     === "Nix profiles (requires experimental flags)"
 
         ```
-        nix profile add nixpkgs#bashInteractive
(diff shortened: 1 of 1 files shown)


euphemism:
This is not the correct fix. @Triscal needs to update their Nix installation. The `install` command is deprecated in favor of `add`.

```
% nix profile install --help
warning: 'install' is a deprecated alias for 'add'
```

https://github.com/NixOS/nix/pull/13224

PR #2705, opened 2026-04-07
Add mouse support to TUI: click to select, drag to copy, scroll wheel

domenkozar:
Enable mouse capture for process views (up/processes up). Supports double click to expand an activity, drag to select and copy activity names via OSC 52, and scroll wheel navigation.

It has a bug of calculating y axis wrong.

### DIFF
--- a/devenv-tui/src/app.rs
+++ b/devenv-tui/src/app.rs
@@ -3,6 +3,8 @@ use crate::{
     model::{ActivityModel, RenderContext, UiState, ViewMode},
     view::{ActivityHeights, SUMMARY_BAR_HEIGHT, ScrollState, view},
 };
+use base64::Engine;
+use crossterm::event::MouseButton;
 use crossterm::{
     cursor, event, execute,
     style::{Color, ResetColor, SetForegroundColor},
@@ -10,10 +12,12 @@ use crossterm::{
 };
 use devenv_activity::{ActivityEvent, ActivityLevel};
 use devenv_processes::ProcessCommand;
+use iocraft::MouseEventKind;
 use iocraft::prelude::*;
 use std::io::{self, Write};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, OnceLock, RwLock};
+use std::time::Instant;
 use tokio::sync::{Notify, mpsc};
 use tokio_shutdown::Shutdown;
 use tracing::debug;
@@ -86,6 +90,7 @@ pub struct TuiApp {
     shutdown: Arc<Shutdown>,
     command_tx: Option<mpsc::Sender<ProcessCommand>>,
     shutdown_on_backend_done: bool,
+    mouse_capture: bool,
 }
 
 impl TuiApp {
@@ -100,6 +105,7 @@ impl TuiApp {
             shutdown,
             command_tx: None,
             shutdown_on_backend_done: true,
+            mouse_capture: false,
         }
     }
 
@@ -147,6 +153,12 @@ impl TuiApp {
         self
     }
 
+    /// Enable mouse capture for click to select and double click to expand.
+    pub fn with_mouse_capture(mut self) -> Self {
+        self.mouse_capture = true;
+        self
+    }
+
     /// Run the TUI application until the backend completes.
     ///
     /// The `backend_done` notify signals when the backend has completed its
@@ -160,6 +172,7 @@ impl TuiApp {
         let shutdown = self.shutdown;
         let command_tx = self.command_tx;
         let shutdown_on_backend_done = self.shutdown_on_backend_done;
+        let mouse_capture = self.mouse_capture;
 
         let exit_flag = ExitFlag::new();
 
@@ -260,6 +273,7 @@ impl TuiApp {
                 command_tx.clone(),
                 &mut pre_expand_height,
                 exit_flag.clone(),
+                mouse_capture,
             )
             .await;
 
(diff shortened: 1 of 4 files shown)


devenv-tui/src/app.rs
```diff
@@ -576,89 +674,249 @@ fn MainView(mut hooks: Hooks) -> impl Into<AnyElement<'static>> {
...
+                                            .as_ref()
+                                            .is_some_and(|(prev_id, prev_time)| {
+                                                *prev_id == activity_id
+                                                    && prev_time.elapsed().as_millis() < 400
+                                            });
+
+                                        if is_double_click {
+                                            *last_click.write() = None;
+                                            if let Ok(mut ui) = ui_state.write() {
+                                                ui.selected_activity = Some(activity_id);
```
sadjow, line 766: This branch never switches to `ViewMode::ExpandedLogs`, so a double-click only selects the activity instead of expanding it. It also means the first click never selects anything, because `selected_activity` is only written in the double-click path.

devenv-tui/src/app.rs
```diff
@@ -576,89 +674,249 @@ fn MainView(mut hooks: Hooks) -> impl Into<AnyElement<'static>> {
...
+                                            }
+                                        } else {
+                                            *last_click.write() =
+                                                Some((activity_id, Instant::now()));
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                        MouseEventKind::ScrollDown | MouseEventKind::ScrollUp => {
```
sadjow: I think this wheel handler duplicates `ScrollView`'s own mouse scrolling. `view()` already wraps the activity list in `ScrollView`, and iocraft handles `ScrollUp`/`ScrollDown` there, so each wheel tick over the list should scroll twice.

devenv-tui/src/app.rs
```diff
@@ -493,6 +510,79 @@ fn activity_height(heights: &std::collections::HashMap<u64, i32>, id: u64) -> i3
...
+        .sum();
+    let activity_area_height = if is_scrolling {
+        terminal_height.saturating_sub(SUMMARY_BAR_HEIGHT) as i32
+    } else {
+        total_content
+    };
+    let canvas_height = activity_area_height + SUMMARY_BAR_HEIGHT as i32;
+    // Canvas starts at initial_cursor_row, but if it would overflow the
+    // terminal, the terminal scrolls so canvas_bottom = terminal_height - 1.
+    let canvas_top = (initial_cursor_row as i32).min(terminal_height as i32 - canvas_height);
```
sadjow: I don't think terminal-level `canvas_top` math is valid here. `use_terminal_events` rewrites `FullscreenMouse` coordinates into the component's local coordinate system before `MainView` sees them, so subtracting `initial_cursor_row` / `canvas_top` again shifts hit testing and drag selection.

devenv-tui/src/app.rs
```diff
@@ -576,89 +644,248 @@ fn MainView(mut hooks: Hooks) -> impl Into<AnyElement<'static>> {
...
+                            if let Some(start) = *mouse_down_row.read() {
+                                let end = mouse_event.row;
+                                let Ok(model) = activity_model.read() else {
+                                    return;
+                                };
+                                let display = model.get_display_activities();
+                                let heights = activity_heights.read();
+                                let is_scrolling = *scroll_view_active.read();
+                                let scroll_offset = if is_scrolling {
+                                    scroll_handle.read().scroll_offset()
```
sadjow, line 676: Because this `ScrollView` uses `auto_scroll: true`, an overflowing activity list initially displays the bottom rows while `ScrollViewHandle::scroll_offset()` still returns its stored offset, which starts at 0. For example, with 30 one-line activities in a 10-line viewport, rows 21–30 are visible but clicking the first visible row is hit-tested as row 1; selection then jumps to that wrong activity, and drag-copy uses the wrong names. Could we base hit-testing on the effective displayed offset—including the pinned-to-bottom state—and add a test that clicks and drags before any manual scrolling?

devenv-tui/src/app.rs
```diff
@@ -565,9 +627,15 @@ fn MainView(mut hooks: Hooks) -> impl Into<AnyElement<'static>> {
...
-    hooks.use_terminal_events({
+    let mouse_capture = *hooks.use_context::<bool>();
+    // Mouse state for click/drag tracking
+    let mut last_click: Ref<Option<(u64, Instant)>> = hooks.use_ref_default();
+    let mut mouse_down_row: Ref<Option<u16>> = hooks.use_ref_default();
+    let mut mouse_dragged: Ref<bool> = hooks.use_ref_default();
+    let mut mouse_drag_end_row: Ref<Option<u16>> = hooks.use_ref_default();
+
+    // Handle keyboard and mouse events - only UI state updates, no activity model writes
+    hooks.use_local_terminal_events({
```
sadjow, line 638: `use_local_terminal_events` translates the mouse position relative to `MainView` inside iocraft's canvas, but this view is rendered inline. Crossterm's row also includes the canvas's physical starting row in the terminal, and that offset is not removed here. If a short TUI starts at terminal row 12, clicking its first row can arrive as row 12 and be shifted or rejected as outside the component. Could we include the inline canvas origin in this conversion—possibly in iocraft—and test a non-zero terminal start row?

devenv-tui/src/app.rs
```diff
@@ -576,89 +644,248 @@ fn MainView(mut hooks: Hooks) -> impl Into<AnyElement<'static>> {
...
+                                    let r1 = r1.max(0) + scroll_offset;
+                                    let r2 = r2.min(activity_area_height - 1) + scroll_offset;
+                                    let ids = collect_activity_ids(&display, &heights, r1, r2);
+                                    if let Ok(mut ui) = ui_state.write() {
+                                        ui.drag_selected = ids;
+                                    }
+                                }
+                            }
+                        }
+                        MouseEventKind::Up(MouseButton::Left) => {
```
sadjow, line 700: `use_local_terminal_events` only forwards mouse events while the pointer is inside `MainView`. If a drag starts on an activity and the user releases above or below the inline TUI, this `Up` branch never runs: nothing is copied, and `drag_selected` remains highlighted until another press clears it. Could we continue receiving the release once a drag has started—such as through a global listener with explicit bounds and clamping—and test releasing outside the component?

PR #2714, opened 2026-04-08
feat: add version constraint field to devenv.yaml

### LINKED ISSUE #2391: Pin devenv version itself for reproducibility
After spending hours debugging a silent Caddy reverse proxy failure in my Django project with process-compose, I discovered the issue was caused by a devenv version change (1.10 → 1.11). This explained why the project worked fine on older NixOS generations.

I verified the fix exists in a recent commit by pinning devenv:

```bash
nix shell github:cachix/devenv/bcf87e4b117c7a5b78f3f57c6ff8d782a5f02a00 -c devenv up
```

However, I can't find a documented way to pin the devenv version itself within devenv.yaml inputs. The devenv version comes from the NixOS system rather than the project configuration (as expected, unfortunately).

Is there a recommended way to pin devenv version for reproducibility?

If not currently supported, I'd suggest either:
- Adding support for pinning devenv itself in devenv.yaml inputs
- Creating a Nix shell wrapper (ideally without requiring flakes) that pins the devenv version

Currently, without pinning devenv itself, environments seem non-reproducible across different devenv versions installed in the systems, which breaks one of the core promises of the Nix ecosystem.

Bonus question: should systems use devenv from nixpkgs, or is it better to add a flake input pointing to the master branch of this repo?

Thank your for this amazing project ❤️

domenkozar:
## Summary

- Add `version` field to `devenv.yaml` that checks the running devenv version against a constraint (e.g. `>=2.1`, `<3.0.0`, `2.0.7`)
- Errors early (before Nix evaluation) if the constraint is not satisfied
- Supports `X.Y` and `X.Y.Z` version formats with `>=`, `<=`, `>`, `<`, `=` operators or bare version for exact match

Closes #2391.

## Test plan

- [x] Unit tests for all operators, two-component versions, exact match, no constraint, invalid inputs
- [ ] Manual: add `version: ">=99.0.0"` to a devenv.yaml, confirm error
- [ ] Manual: add `version: ">=2.0.0"` to a devenv.yaml, confirm it passes

### DIFF
--- a/devenv-core/src/config.rs
+++ b/devenv-core/src/config.rs
@@ -2,6 +2,7 @@ use miette::{IntoDiagnostic, Result, WrapErr, bail};
 use pathdiff;
 use schemars::{JsonSchema, schema_for};
 use schematic::ConfigLoader;
+use semver::{Version, VersionReq};
 use serde::{Deserialize, Serialize};
 use std::{
     collections::{BTreeMap, HashMap, HashSet},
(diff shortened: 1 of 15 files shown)


devenv-core/src/config.rs
```diff
@@ -246,6 +246,10 @@ pub struct Nixpkgs {
 #[config(rename_all = "camelCase", allow_unknown_fields)]
 #[serde(rename_all = "snake_case")]
 pub struct Config {
+    /// Version constraint for devenv (e.g. ">=2.0.0", "2.0.7")
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    #[setting(merge = schematic::merge::replace)]
+    pub version: Option<String>,
```
sandydoo: Are we sure about calling this `version`?

domenkozar: I'm not sure, but couldn't think of anything better.

domenkozar: renamed to `require_version`

PR #2718, opened 2026-04-08
Zsh/fish/nushell support

### LINKED ISSUE #36: zsh support

### LINKED ISSUE #2487: devenv shell: support fish as interactive shell
## Problem

`devenv shell` currently only supports bash as the interactive shell. For users whose default shell is fish, this means `devenv shell` drops them into bash instead of fish, losing their fish configuration, functions, completions, and prompt.

## Context

Issue #36 tracks zsh support, and recent work (by @sandydoo and @mcdonc) has made progress on alternative shell support. Fish is another commonly used shell that would benefit from the same treatment.

As devenv moves toward replacing direnv (#2061), native fish support in `devenv shell` becomes a prerequisite for fish users to migrate away from the `use devenv` direnv integration.

## Expected behavior

`devenv shell` should detect the user's preferred shell (via `$SHELL` or similar) and launch fish when appropriate, with the devenv environment correctly exported.

---
*This issue was created on behalf of @schickling*

### LINKED ISSUE #2592: Support nushell
Add built-in support for launching a `nushell` with `devenv shell`.

Currently, this has to be done with `devenv shell nu`

### LINKED ISSUE #2593: Support alternative shells out of the box
This is a tracking issue for adding built-in support for alternative (to bash) shells.

Vote for your favourite shell by upvoting the corresponding issue!

- [ ] `fish` #2487
- [ ] `nushell` #2592 
- [ ] `zsh` #36

## Current state of affairs

`devenv shell` always launches a bash shell.
Using alternative shells requires using a third-party integration or manually launching the shell, each with its own pros and cons.

#### `direnv`

`direnv` evaluates the shell using bash and then layers the environment on top of the system shell.
This is currently the best experience if you want a custom shell.

Cons:
- Uses whichever shell the system uses
- Leaks system environment variables, making the environment less reproducible.

#### `devenv shell <cmd>`

`devenv shell zsh` launches a bash shell, evaluates the environment there and then `exec`s into `zsh`. 

Cons:
- The load order will be incorrect. If your rcfile/profile touches `PATH` or other variables, then the system profile will take precedence.
- In devenv 2.0, commands passed to `devenv shell` are `exec`'d without launching an auto-reloading shell, so this approach is currently incompatible with that feature.

## System shell

Another feature we could implement is the ability to layer the shell environment on top of the existing shell, much like `direnv` does. This makes the environment less reproducible, but is sometimes convenient, particularly on macOS.

domenkozar:
Fixes #2592 
Fixes #36 
Fixes #2487
Fixes #2593


mcdonc:
Hoorah!

PR #2733, opened 2026-04-11
languages.javascript: make Node.js optional

arch-fan:
## Summary

Currently, enabling `languages.javascript` implicitly adds Node.js, even in setups that only want to use Bun.

This PR adds a new option:

`languages.javascript.nodejs.enable`

which defaults to `true` for backward compatibility. When set to `false`, Node.js is no longer added to the environment.

This makes Bun-only JavaScript environments possible while keeping existing configurations unchanged.

## Validation

- `languages.javascript.enable = true;` still includes Node.js
- `languages.javascript = { enable = true; bun.enable = true; nodejs.enable = false; };` works without Node.js
- Enabling npm/pnpm/yarn/corepack/lsp without Node.js now fails with a clear assertion

### DIFF
--- a/src/modules/languages/javascript.nix
+++ b/src/modules/languages/javascript.nix
@@ -236,6 +236,10 @@ in
       description = "The Node.js package to use.";
     };
 
+    nodejs = {
+      enable = lib.mkEnableOption "Node.js runtime" // { default = true; };
+    };
+
     corepack = {
       enable = lib.mkEnableOption "wrappers for npm, pnpm and Yarn via Node.js Corepack";
     };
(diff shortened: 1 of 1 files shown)


src/modules/languages/javascript.nix
```diff
@@ -348,5 +349,44 @@ in
         ''
       ]
     );
+
+    assertions = [
+      {
+        assertion = (!cfg.npm.enable) || cfg.nodejs.enable;
```
sandydoo, line 355: nitpick: nix has a nice logical implication operator for this exact pattern `->`

sandydoo (changes requested):
Thanks! 

Technically, there's no reason why one couldn't use another package manager with `bun`, but I think it makes sense for the module system to steer people towards "best practices".

sandydoo:
Another option would be to add a `runtime` enum. Then `bun.enable` would be deprecated and migrated to `runtime`. We could then bring back `deno` as an option and deprecate the separate `deno` language.

arch-fan:
I will not remove Bun out of JavaScript.

I like the runtime enum, but it exists the case where you would like to have two runtimes (bun & node for example)

Maybe a list of runtimes with default node (but we should be able to exclude node also, that would be uncomfortable)

Maybe "runtime" enum and "extraRuntimes"? Also for defaulting the main runtime for some scripts like "language.javascript.install" and making proprietary scripts for each runtime

sandydoo:
> I like the runtime enum, but it exists the case where you would like to have two runtimes (bun & node for example)

There's a point at which we need to limit the features. `languages` are opinionated defaults. Extra runtimes add complexity that we can't encode reliably for everyone. If you just need the binary in PATH, then it can be added directly to `packages`.

arch-fan:
> Extra runtimes add complexity that we can't encode reliably for everyone.

Well, that's true, it will be just a shortcut for some nix expressions which will be searching your runtimes.

> If you just need the binary in PATH, then it can be added directly to `packages`.

Yep, true.

I’m trying to think of cases where extra runtimes would make sense. Any ideas? Just adding the runtime to packages should be enough.

And does devenv expose lib? Maybe it would be useful a lib.enabledJavaScriptRuntimes or even other kind of helpers (this is another topic, I know)

sandydoo:
Going to merge as is. We can figure out alternatives later.

PR #2739, opened 2026-04-13
feat(tui): Add human-facing process URLs.

hencjo:
# Summary

Add first-class human-facing process URLs.

This adds:
- `processes.<name>.urls.<label>`
- `devenv processes endpoints`
- inline clickable URLs in the `devenv up` TUI
- seeded URLs for browser-facing built-in services
- urls defined for some processes (mailhog, mailpit, minio, adminer, rabbitmq)

# Description

## Overview

This PR adds a first-class way to describe and surface browser-facing process endpoints.

It builds on existing process and port support by letting processes declare human-facing URLs that can reference dynamically allocated ports, and then exposing those URLs consistently in the CLI and TUI.

2026-04-17-143100_822x147_scrot

### Process URLs

Processes can now declare human-facing URLs:

```nix
processes.<name>.urls.<label> = {
  scheme = "http";
  host = "127.0.0.1";
  port = config.processes.<name>.ports.<port>.value;
  path = "/";
};
```

These URLs can reference dynamically allocated ports, so the displayed URL matches the resolved runtime port.

### CLI

Added:

```bash
devenv processes endpoints
```

Example output:

```text
rabbitmq/admin   http://127.0.0.1:15672/
mailpit/ui       http://127.0.0.1:8025/
```

This is useful both for user-defined processes and for services started with `devenv up -d`.

### TUI

`devenv up` now shows process URLs inline as clickable links.

### Built-in services

Added human-facing URLs for:
- RabbitMQ (`admin`)
- Mailpit (`ui`)
- MailHog (`ui`)
- MinIO (`console`)
- Adminer (`ui`)

Wildcard bind addresses such as `0.0.0.0` and `::` are normalized to `127.0.0.1` for display.

## Tests

Added coverage for:
- process URL config/plumbing
- native manager endpoint rendering
- CLI parsing and formatting
- TUI URL rendering
- integration test for resolved dynamic runtime URLs

## Docs

Updated `docs/src/processes.md` with:
- `processes.<name>.urls`
- dynamic port composition


hencjo:
I realize this feature is unsolicited and I'm happy to take any directions on how to adapt it or simply skip it.

If it's found to be helpful I'll add the URLs of existing services in next PR.

venkatd:
Hi I wanted to reference the portless project for inspiration.

They have a section on why portless makes sense:
https://portless.sh/why

I never actually set this up because I didn't want to deal with installing a javascript package on top of devenv and then figuring out how to integrate the two.

Let me give you an example from our process. Suppose I have 3-4 running worktrees with agents developing different things in our application.

Each worktree has its own sevices running. In our case it is postgres, garage for s3, frontend react server (we have two because we have two separate apps), backend Elixir server, one backend microservice.

Thankfully devenv makes it easy with `devenv up` and processes to spin these up.

But then we have to allocate ports for (# of worktrees) * (# of services per worktree) and that gets out of hand. Because I can't easily keep track of "what is the frontend for this worktree? how can I give it a try?" And the agents make mistakes on this too. I run into port conflict and stuff like that.

Example snippet from agent mistake:
> The canonical stack is healthy now: Rosetta is serving /status from the single committed worktree on port 3003. One older Rosetta process is still listening on 3024 under a different preexisting manager; I’m identifying its owner before stopping it so we do not kill an unrelated checkout.

That's the main case!

I don't have ideas on the best API but their configuration section might give you some ideas here: https://portless.sh/configuration

One idea would be:
- Make it easy for a devenv processes config to know the branch info (worktree, etc)
- Allow configuration of a custom `.local` url per service based for example to get something like `clients.myapp--signatures-v2.local` where `signatures-v2` is the branch name


I'm open to other solutions on this. Just sharing the scenario and one possible solution!

venkatd:
I wanted to also follow up with a simple use case

On some of my devenvs we have development tooling.

For example a tracing server like jaeger.
I will pick arbitrary ports and hardcode that.

But then it's hard to keep track of what service is on what port because I have to map port number to service name in my head.

So even without worktrees, it's hard because I'm working with ports instead of human readable names!

venkatd:
Here is what I might propose for the API:

```
processes.frontend = {
  exec = "npm run dev";
  ports.http = {
    allocate = 5173; # omit if we want a dynamically allocated port
    hostname = "clients.myapp--{branch}.localhost"; # can be dynamic if we want
  };
};
```

This would make it so that if I navigate to `http://clients.myapp--login-v2.localhost`, it would resolve to whatever the allocated port is.

I put this as a sibling property to allocate because it seems like they are related. Both are about how to make the service available.

This would mean that we could have:
- a bunch of worktrees with ephemeral environments running dev envs w/ a conflict-free convention on how to access each 
- OS-wide services with named `.localhost` domains

Later on, this could be extended to allowing non `.localhost` domains if we want.

hencjo:
@venkatd this seems like a much nicer way to approach the end goal. And as portless is included in 2.3 I say we abandon this. Thanks for this!

PR #2746, opened 2026-04-18
feat(clickhouse):  adding user creation support

szucsitg:
Current Clickhouse implementation doesn't allow to create a user or modify any behavior.
Additionally I run into issues when try to run migration scripts due to lacking keeper integration and configuration for it.
Also macro support wasn't enabled.

### DIFF
--- a/src/modules/services/clickhouse.nix
+++ b/src/modules/services/clickhouse.nix
@@ -62,19 +62,19 @@ in
       description = "Which timezone to use for ClickHouse.";
     };
 
-    enableMacros = lib.mkOption {
+    macros.enable = lib.mkOption {
       type = types.bool;
       default = false;
       description = "Whether to enable macros in ClickHouse.";
     };
 
-    enableRemoteServers = lib.mkOption {
+    remoteServers.enable = lib.mkOption {
       type = types.bool;
       default = false;
       description = "Whether to enable remote_servers in ClickHouse.";
     };
 
-    enableKeeper = lib.mkOption {
+    keeper.enable = lib.mkOption {
       type = types.bool;
       default = false;
       description = "Whether to enable keeper_server in ClickHouse.";
@@ -131,13 +131,13 @@ in
         local_directory:
           path: ${config.env.DEVENV_STATE}/clickhouse/access
 
-      ${lib.optionalString cfg.enableMacros ''
+      ${lib.optionalString cfg.macros.enable ''
         macros:
           shard: 1
           replica: localhost
       ''}
 
-      ${lib.optionalString cfg.enableRemoteServers ''
+      ${lib.optionalString cfg.remoteServers.enable ''
         remote_servers:
           default:
             shard:
@@ -146,7 +146,7 @@ in
                 port: ${toString allocatedPort}
       ''}
 
-      ${lib.optionalString cfg.enableKeeper ''
+      ${lib.optionalString cfg.keeper.enable ''
         keeper_server:
           tcp_port: ${toString allocatedKeeperPort}
           server_id: 1
--- a/src/modules/services/clickhouse.nix
+++ b/src/modules/services/clickhouse.nix
@@ -1,4 +1,9 @@
-{ pkgs, lib, config, ... }:
+{
+  pkgs,
+  lib,
+  config,
+  ...
+}:
 
 let
   cfg = config.services.clickhouse;
@@ -7,8 +12,13 @@ let
   # Port allocation
   basePort = cfg.port;
   baseHttpPort = cfg.httpPort;
+  baseKeeperPort = cfg.keeperPort;
+  baseRaftPort = cfg.raftPort;
   allocatedPort = config.processes.clickhouse-server.ports.main.value;
   allocatedHttpPort = config.processes.clickhouse-server.ports.http.value;
+  allocatedKeeperPort = config.processes.clickhouse-server.ports.keeper.value;
+  allocatedRaftPort = config.processes.clickhouse-server.ports.raft.value;
+
   format = pkgs.formats.yaml { };
 in
 {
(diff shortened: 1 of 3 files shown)


src/modules/services/clickhouse.nix
```diff
@@ -47,20 +70,56 @@ in
         console: 1
       tcp_port: ${toString allocatedPort}
       http_port: ${toString allocatedHttpPort}
+      mysql_port: 0
+      postgresql_port: 0
```
sandydoo: What are these for?

src/modules/services/clickhouse.nix
```diff
@@ -47,20 +70,56 @@ in
...
+      keeper_server:
+        tcp_port: 9181
+        server_id: 1
+        log_storage_path: ${config.env.DEVENV_STATE}/clickhouse/coordination/log
+        snapshot_storage_path: ${config.env.DEVENV_STATE}/clickhouse/coordination/snapshots
+        raft_configuration:
+          server:
+            id: 1
+            hostname: localhost
+            port: 9234
```
sandydoo: Any reason to enable this by default?

src/modules/services/clickhouse.nix
```diff
@@ -47,20 +70,56 @@ in
...
     '';
+
     processes.clickhouse-server = {
       ports.main.allocate = basePort;
       ports.http.allocate = baseHttpPort;
-      exec = "exec clickhouse-server --config-file=${pkgs.writeText "clickhouse-config.yaml" cfg.config}";
+      exec = ''
+        mkdir -p ${config.env.DEVENV_STATE}/clickhouse/{server/config.d,server/users.d,tmp,user_files,format_schemas,access,caches,coordination/snapshots,coordination/log}
```
sandydoo, line 165: Are all of these required? Does clickhouse fail to start otherwise?

src/modules/services/clickhouse.nix
```diff
@@ -47,20 +70,56 @@ in
...
+        install -m 644 ${cfg.package}/etc/clickhouse-server/users.xml ${config.env.DEVENV_STATE}/clickhouse/server/users.xml
+        install -m 644 ${cfg.package}/etc/clickhouse-server/config.xml ${config.env.DEVENV_STATE}/clickhouse/server/config.xml
+        install -m 644 ${pkgs.writeText "clickhouse-config.yaml" cfg.config} ${config.env.DEVENV_STATE}/clickhouse/server/config.d/clickhouse-config.yaml
+        install -m 644 ${format.generate "users.yaml" cfg.usersConfig} ${config.env.DEVENV_STATE}/clickhouse/server/users.d/users.yaml
+        exec clickhouse-server --config-file=${config.env.DEVENV_STATE}/clickhouse/server/config.xml
+      '';
```
sandydoo, line 170: This should be a task that runs before the process. Something like:

```nix
tasks."devenv:clickhouse:setup" = {
    description = "Setup ClickHouse";
    exec = ''
      mkdir -p "${serverDir}"/{config.d,users.d}
      install -m 644 ${cfg.package}/etc/clickhouse-server/users.xml  "${serverDir}/users.xml"
      install -m 644 ${cfg.package}/etc/clickhouse-server/config.xml "${serverDir}/config.xml"
      install -m 644 ${generatedConfig} "${serverDir}/config.d/clickhouse-config.yaml"
      install -m 644 ${generatedUsers}  "${serverDir}/users.d/users.yaml"
    '';
    before = [ "devenv:processes:clickhouse-server" ];
  };
```

sandydoo (changes requested):

src/modules/services/clickhouse.nix
sandydoo: I just read your comment about migrations. Could you explain in detail what the issue was and how this solves it?

If we add stuff like this, we should document why it was added.

szucsitg:
Thanks for your comments, I’ll get back to you with some changes based on your feedback. I also found a few minor issues in the last few days while I was testing my changes.

src/modules/services/clickhouse.nix
szucsitg: i wanted to explicitly disabled it in my case, but looking at deeper into the config (clickhouse docs are sometimes hard to follow), it's not necessary

src/modules/services/clickhouse.nix
szucsitg: I'm not a clickhouse expert, but macros needed for replicated tables, remote server is needed distributed table engine, and raft is just a prerequisite to run in a cluster. i think the tool's migration scripts just assumes production quality deployment scenario, but i'll make it optional and default opt-out

src/modules/services/clickhouse.nix
szucsitg: i've interated through a coouple of scenarios, and it might not fail on all of them, but fails on a lot of them. and went with the fact it's easier to have an empty dir pre-created

sandydoo (changes requested):
Looking good! My only request is to rename the options.

For example, instead of `keeperEnable`, we prefer `keeper.enable`. Same for the rest.
I'll make sure to document this somewhere for the future.

szucsitg:
@sandydoo i made the changes

src/modules/services/clickhouse.nix
```diff
@@ -33,10 +44,68 @@ in
...
+      type = types.port;
+      description = "Which port to run clickhouse keeper service on.";
+      default = 9181;
+    };
+
+    raftPort = lib.mkOption {
+      type = types.port;
+      description = "Which http port to use clickhouse keeper for raft consensus.";
+      default = 9234;
+    };
```
sandydoo: `keeperPort` -> `keeper.port`.

`raftPort` -> `keeper.raft.port`? Does nesting under keeper make sense? Not familiar with this thing.

sandydoo (changes requested):

src/modules/services/clickhouse.nix
szucsitg: fixed  and squashed commits

sandydoo:
Awesome! Thanks!

PR #2748, opened 2026-04-19
eval-cache: skip cache write when input files change during eval

### LINKED ISSUE #2745: devenv often doesn’t update tasks when definitions change
**Describe the bug**
tasks can stay stale when devenv.nix changes during build 
- if `devenv.nix` changes while a `devenv tasks list` run is still evaluating, the completed run can write stale task state
- the next `devenv tasks list` can still print the old tasks and miss the new one

**To reproduce**
https://github.com/vibecoder43/devenv_repro see `mre/repro.sh`
**Version**
2.0.6

domenkozar:
## Summary

- Detect when a tracked input (e.g. `devenv.nix`) is rewritten while Nix is still evaluating, and skip persisting the cache entry in that case.
- Capture `eval_started_at` before evaluation; in `ops_to_inputs`, flag any tracked file whose untruncated mtime exceeds it. On `race_detected`, the eval result is still returned but nothing stale gets stored — next run re-evaluates against current disk state.
- Fixes #2745.

## Test plan

- [x] `cargo nextest run -p devenv-eval-cache` — 43 tests pass, including 3 new race-detection tests covering the read path, the stable-file case, and `extra_watch_paths`.
- [x] `cargo clippy -p devenv-eval-cache --tests` — clean.
- [ ] Reproduce locally: start `devenv shell`, edit `devenv.nix` while evaluation is in flight, and confirm the next run re-evaluates instead of serving a stale result.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bug Fixes
 
 - Fixed `devenv shell` and `devenv build` failing with `path '...drv' is required, but there is no substituter that can build it` after the cached derivation was garbage-collected. The stale eval-cache entry is now invalidated when its referenced store paths no longer exist, forcing a re-evaluation that re-materializes the `.drv` on disk.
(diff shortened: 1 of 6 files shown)


domenkozar:
We have to be careful here, as we want the opposite for reload. If devenv shell changes stuff during the build, then it shouldn''t reload itself as it could create a loop.

PR #2754, opened 2026-04-21
fix(reload): remove startup hang and track state across reload cycles

ndam-hexagon:
## Summary                         

This originated because we observed extremely long devshell load times when running `devenv shell` (after the shell finished building).
Running `devenv shell --no-reload` lead to significantly faster load times.

Investigation of the root cause and implementation of the fixes done with Claude Code. Validated that with the patches in this PR, `devenv shell` does not hang.

## Changes                                                                                                                                                         
                                                                                                                                                                                              
  - Batch watcher path registration via new `WatcherHandle::watch_many` — single lock, one pathset update, one fs-readiness wait. Replaces per-path `watcher.watch` calls in reload cache hydration that caused long hangs before `enterShell` on large cached input sets.                                                                                                            
  - Track path state (file/dir/missing) across reload cycles, hash watched directories by sorted listings, and queue deferred changes while a build is in flight. Reconcile drift after the
  rewatch gap so changes during watch refresh still trigger a follow-up rebuild.     

## Validation

Ran the following commands successfully:
```
devenv shell -- cargo fmt --check
devenv shell -- cargo clippy --all-targets -- -D warnings                                                                                                                                   
devenv shell -- cargo nextest run -p devenv-reload -p devenv-event-sources -p devenv-cache-core -p devenv
devenv shell -- cargo nextest run --features devenv/test-all
devenv shell -- devenv-run-tests run tests
```

> [!note]
>
> Pre-existing warnings surfaced with `-D warnings` ignored

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Bug Fixes
 
 - Fixed `devenv shell` and `devenv build` failing with `path '...drv' is required, but there is no substituter that can build it` after the cached derivation was garbage-collected. The stale eval-cache entry is now invalidated when its referenced store paths no longer exist, forcing a re-evaluation that re-materializes the `.drv` on disk.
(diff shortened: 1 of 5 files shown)


domenkozar:
Nice! Could you update CHANGELOG.md?

ndam-hexagon:
> Nice! Could you update CHANGELOG.md?

Done.

PR #2757, opened 2026-04-23
nix_backend: set umask around nix ops

### LINKED ISSUE #2751: CircleCI does not work after 2.0
Since 2.0 (and still happening in 2.6) I can't use devenv in circle-ci, it always fails with:
```
Error:   × Evaluation error: Failed to realize shell derivation: error: Cannot build
  │ '/nix/store/7almlgrmqg6amj8l66qi3z3w5gy338pj-devenv-shell.drv'.
  │        Reason: 1 dependency failed.
  │        Output paths:
  │          /nix/store/w3cl96a0422c3d4fxh7q9virzny266jr-devenv-shell
```

The error is not bound to the content of the devenv, even `{}:{}` fails with the same error.

I'm using this circleCI config: 

```yaml
version: 2.1
jobs:
  test:
    # resource_class: arm.medium
    machine:
      image: ubuntu-2404:current
    steps:
      - run:
          command: |
              sh <(curl -L https://nixos.org/nix/install)
      - checkout
      - run:
          name: Install devenv
          command: |
            . /home/circleci/.nix-profile/etc/profile.d/nix.sh
            nix-env -iA devenv -f https://github.com/NixOS/nixpkgs/tarball/nixpkgs-unstable
            export SKIP=no-commit-to-branch
            devenv test -v

      - store_test_results:
          path: test-results

      - store_artifacts:
          path: htmlcov
workflows:
  version: 2
  test:
    jobs:
      - test
```

hopefully it still works if I pi a nixpkgs before 2.0, but it's not nice and I fear it will start failing at some point if nix part of devenv stops playing nice with the old binary.

sandydoo:
Nix expects outputs to be created with group permissions set to read. In single-user installs, it's our job to ensure this happens.

A global umask would affect the files we write.
This scoped guard is still global, but since the FFI is sync, it should be safe.

Fixes #2751.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug Fixes
 
+- Fixed "suspicious ownership" errors when using Nix installed in single-user mode. Nix requires read-only group permissions on outputs ([#2751](https://github.com/cachix/devenv/issues/2751)).
 - Fixed the root `devenv.nix` being imported twice when `devenv.yaml` imported a sub-project that also had its own `devenv.yaml` ([#2755](https://github.com/cachix/devenv/issues/2755)).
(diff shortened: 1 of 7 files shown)


domenkozar (approved):
needs a changelog

PR #2777, opened 2026-04-30
Bugfix: Fix dynamic process port allocation reusing stale cached port values.

hencjo:
This appears to have been introduced by the backend refactor in `c5078654`.

## What changed

- Cache keys now include the current port allocator mode.
- `devenv up`, `devenv test`, and task runs enable/seed port allocation before `setup_cachix()` can evaluate user config.
- Build-cache hits replay saved port resources before reusing cached outputs; failed replay invalidates port-dependent cache and re-evaluates.
- Native manager replay is strict and uses the native ports API for seeding; process-compose keeps its existing in-use replay behavior.

## Test

Adds `tests/process-port-allocation-two-repos`:

- creates two independent projects with the same `ports.http.allocate` base port
- warms repo2 cache on the base port
- starts repo1 on the base port
- starts repo2 and verifies it re-evaluates to a different bound port

The test records the port only after the process successfully binds, so it catches real collisions.

## Note (inherent race)

There's a (preexisting) inherent race-condition when launching two `devenv up`. The port is reserved and released on process launch. But that means it's up for grabs by a different `devenv up` until the process binds. 

I have been unable to find a non-invasive way to fix this and I'm not sure there is a solution (without shared state between devenv up instances). However, it would be really nice if we could figure it out.


hencjo:
(Need to check something before this is ready for review.)

hencjo:
The test doesn't catch the regression. I'm reworking it.

PR #2781, opened 2026-05-03
feat: add services.garage

ap-1:
Adds a `services.garage` module that runs a single-node [Garage](https://garagehq.deuxfleurs.fr/) instance under devenv's process supervisor.

- `src/modules/services/garage.nix` generates a `garage.toml`, runs `garage server`, declares an HTTP readiness probe against the admin API, and configures the cluster on first start via a task gated on the readiness probe.
- `tests/garage/` enables the service with a single bucket plus a `.test.sh` that asserts the admin API answers `/v1/health`.

Garage rejects S3 traffic until at least one node has a role assigned, so `tasks."devenv:garage:configure"` runs `layout assign` + `layout apply` after the readiness probe clears, then idempotently creates any buckets the user declared in `services.garage.buckets`. This mirrors the mysql / rustfs pattern of using a separate task gated by `processes.X.before`.


domenkozar:
See https://github.com/cachix/devenv/issues/2786

Issue #2205, opened 2025-10-07
Allow importing devenv.yaml from inputs


domenkozar:
Now that #14 has landed, we're missing the ability to import devenv.yaml from inputs.

domenkozar:
@sandydoo has ideas how to fix this one using two-pass locking logic.

PR #2783, opened 2026-05-04
feat: add devenv.preFlight for pre-init env injection

ap-1:
A few of devenv's own subsystems (cachix push auth, substituter netrc) read environment variables at process startup, before any module-declared `env.X` is exported into the resulting shell. Values fetched at module-evaluation time can't reach those subsystems.

For organizations that store `CACHIX_AUTH_TOKEN` in a secret manager (Vault, OpenBao, 1Password, secretspec), this means there's no way to feed the token into devenv-core's cachix init without each developer running `cachix authtoken <token>` manually first. The cachix daemon spawns with whatever env devenv was launched with, which doesn't include anything modules declare.

This PR adds a new `devenv.preFlight` option (`attrsOf submodule { command = str; }`). Each command is run, in alphabetical key order, before subsystem initialization. Lines of `KEY=value` written to stdout are merged into devenv's process environment, so the subsequent `env::var()` reads in `devenv-core` (cachix, etc.) see them.

```nix
devenv.preFlight.cachix-auth.command = ''
  echo "CACHIX_AUTH_TOKEN=$(${pkgs.openbao}/bin/bao kv get \
    -address=https://vault.example.com \
    -field=CACHIX_AUTH_TOKEN secret/cachix)"
'';
```

### DIFF
--- a/devenv/src/devenv/mod.rs
+++ b/devenv/src/devenv/mod.rs
@@ -1,6 +1,7 @@
 mod cachix;
 mod container;
 mod gc;
+mod preflight;
 mod search;
 
 use super::{processes, tasks, util};
@@ -663,6 +664,7 @@ impl Devenv {
                 let Some(cnix) = self.cnix() else {
                     return Ok(None);
                 };
+                preflight::run(cnix).await?;
                 cachix::CachixIntegration::init(
                     cnix,
                     &self.cachix_manager,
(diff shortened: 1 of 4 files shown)


domenkozar:
Did you see we recently added support fetching CACHIX_AUTH_TOKEN from secretspec?

domenkozar:
Discussed on Discord: There's now CACHIX_AUTH_TOKEN integration using secretspec.dev

We do need to fix https://github.com/cachix/devenv/issues/2205

PR #2785, opened 2026-05-04
Revert "switching nix profile add to install"

euphemism:
This reverts commit 7754bb69390738dc0464592b6fd3ec93174c3a03.

Context:

This is not the correct fix. @Triscal needs to update their Nix installation. The `install` command is deprecated in favor of `add`.

```
% nix profile install --help
warning: 'install' is a deprecated alias for 'add'
```

https://github.com/NixOS/nix/pull/13224

_Originally posted by @euphemism in https://github.com/cachix/devenv/issues/2678#issuecomment-4374294006_

### DIFF
--- a/docs/src/getting-started.md
+++ b/docs/src/getting-started.md
@@ -40,7 +40,7 @@
     === "Nix profiles (requires experimental flags)"
 
         ```
-        nix profile install nixpkgs#bashInteractive
+        nix profile add nixpkgs#bashInteractive
         ```
 
 === "Windows (WSL2)"
@@ -68,7 +68,7 @@
 === "Nix profiles (requires experimental flags)"
 
     ```
-    nix profile install nixpkgs#devenv
+    nix profile add nixpkgs#devenv
     ```
 
 === "NixOS/nix-darwin"


domenkozar:
We reverted that because not all Nix installations support add yet.

euphemism:
Got it.

Triscal:
Thanks @euphemism! 

This has caused me to go down a rabbit hole to figure out how I missed https://github.com/NixOS/nix/pull/13224. 

I use [lix](https://lix.systems) as my nix variant and they [have not yet added `add`](
https://git.lix.systems/lix-project/lix/issues/1139). So while my original PR was not [verbose or well written](https://github.com/cachix/devenv/pull/2678) it was accurate that the latest version of my 'nix' doesn't have `add`. [Lix not having `add` has caused other issues](https://github.com/nix-community/home-manager/issues/8786) so maybe they will add it eventually. 

Nix also has multiple released versions: 
- https://nix.dev/reference/nix-manual 
- https://github.com/NixOS/nix/tags (as of today the following versions were released: 2.34.7, 2.33.6, 2.32.8, 2.31.5, 2.30.5, 2.29.4, 2.28.7)

At least based off the manual the versions that have `nix profile add` [2.30.x](https://nix.dev/manual/nix/2.30/) and higher. So two of the releases from today don't have it available in 'official' nix ([2.29](https://nix.dev/manual/nix/2.29/command-ref/new-cli/nix3-profile-install.html) and [2.28](https://nix.dev/manual/nix/2.28/command-ref/new-cli/nix3-profile-install.html) which is also 'the current stable release' according to https://nix.dev/reference/nix-manual

I'm not going to pretend that I know the full depths of reasons why nix has multiple supported releases, but I do still agree that the `nix profile install` command is the right command to successfully install devenv on the most number of systems. Maybe an adjustment could be made to say to use `nix profile add` and if that is not available then use `nix profile install` but considering `nix profile install` works I don't think it is necessary.

PR #2789, opened 2026-05-06
docs: fix timeout vs probe_timeout

gotcha:
Confusion between timeout and probe_timeout.

(No idea how auto generation of options.md was triggered.)


gotcha:
avoid auto generated commit

PR #2806, opened 2026-05-08
reload: fix infinite reload loop and spurious rebuilds from .devenv/

ndam-hexagon:
## Problem

`devenv shell` self-triggered hot-reload in an infinite loop after the first reload, and `lib.fileset.fromSource ./.` (or any other `readDir` on a parent of `.devenv/`) caused spurious cache invalidations and rebuilds.

Both came from the same root cause: devenv's own dotfile churn was leaking into Nix's tracked-input set, so every evaluation invalidated the previous one's cache.

## Fix

Exclude `.devenv/` from the reload watch set and the eval cache's tracked inputs, keeping `.devenv/state/` (`$DEVENV_STATE`) so files users persist there still trigger reload.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Fixed `devenv --version` and `devenv -V` failing with `'devenv' requires a subcommand but one was not provided`. The flags now print the version and exit, matching the behavior of `devenv --help` ([#2791](https://github.com/cachix/devenv/issues/2791)).
 - Fixed `devenv hook fish` not activating when starting a new fish shell directly inside a project directory. The initial activation now runs on the first `fish_prompt` event instead of inline during `source`, so the spawned `devenv shell` inherits the real terminal as stdin instead of the closed pipe from `devenv hook fish | source` ([#2798](https://github.com/cachix/devenv/issues/2798)).
+- Fixed `devenv shell` self-triggering hot-reload in an infinite loop after the first reload. The eval cache (`.devenv/profiles/<profile>/nix-eval-cache.db`) and its SQLite WAL/SHM sidecars were being added to the reload watch set; every Nix evaluation rewrote the WAL and the watcher saw it as a content change. Devenv's own state directory is now excluded both from the reload watch set and from the eval cache's tracked input paths.
 
 ### Improvements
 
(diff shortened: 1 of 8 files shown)


domenkozar:
Maybe we should skip individual devenv files? I'm worried that someone might store something into DEVENV_STATE and expect the reload to work.

domenkozar (approved):

schickling-assistant:
Independent confirmation that this is still needed, plus two adjacent findings this PR does not cover.

On **devenv 2.2.2+b8030c5** (the current stable release) and on `main` (`30058c52`), devenv still registers files it generates inside its own state directory as eval-cache inputs, and then rewrites them on the next invocation. Minimal repro (empty `devenv.nix`; every input below is devenv's own):

https://github.com/schickling-assistant/devenv-eval-cache-repros/tree/main/repro-4

```
== 1. Every path the eval cache will re-validate on the next invocation
  dir=0 recursive=0  /home/<user>/.config/nixpkgs/overlays        <-- OUTSIDE THE PROJECT
  dir=0 recursive=0  /home/<user>/.config/nixpkgs/overlays.nix    <-- OUTSIDE THE PROJECT
  dir=0 recursive=0  .devenv/bootstrap/bootstrapLib.nix           <-- DEVENV'S OWN STATE DIR
  dir=0 recursive=0  .devenv/bootstrap/default.nix                <-- DEVENV'S OWN STATE DIR
  dir=0 recursive=0  .devenv/bootstrap/resolve-lock.nix           <-- DEVENV'S OWN STATE DIR
  env inputs: DEVENV_CONTAINER, HOME, NIXPKGS_ALLOW_*, SECRETSPEC_SECRETS

== 3. devenv rewrites them, so its own state is load-bearing cache input
  appended '# MARKER-FROM-REPRO' to .devenv/bootstrap/default.nix
    marker present before next devenv run: 1
    marker present after  next devenv run: 0   (silently regenerated)
```

Excluding `.devenv/` from the tracked-input set (this PR) fixes the
`.devenv/bootstrap/*.nix` rows above, so I'd love to see it land — it's been
approved since May and is now showing as conflicted. Two things it looks like
it would **not** cover, in case they're in scope:

1. **`$HOME/.config/nixpkgs/overlays{,.nix}` are tracked as absolute inputs
   outside the project**, and **`HOME` is an env input**. Together they make a
   populated `.devenv` non-portable: a cache warmed by one user / container /
   CI image misses on `HOME` alone. If the paths derived from `HOME` are already
   tracked individually, `HOME` looks like it belongs in `excluded_envs`
   (`devenv-nix-backend/src/backend.rs:367`) alongside `NIXPKGS_CONFIG`.

2. **Latent mass-invalidation.** The cache is valid today only because devenv
   regenerates byte-identical bootstrap files. Any change to the bootstrap
   generator — i.e. any devenv version bump — changes those bytes and
   invalidates every cached attribute in every project, on top of whatever else
   the upgrade invalidates. Excluding the state dir removes this coupling too.

The one-line shape, if it helps, matches this PR's approach —
`excluded_paths` at `devenv-nix-backend/src/backend.rs:368` currently holds a
single hardcoded entry (`self.nixpkgs_config_path`); adding the `.devenv`
dotfile prefix there, with the filter already at
`devenv-eval-cache/src/ffi_cache.rs:156-164`, covers the state-dir half.

Versions: devenv 2.2.2+b8030c5 (x86_64-linux) and `main` `30058c52`; nix
Determinate 3.17.3 / 2.33.3; Linux 6.18.43 x86_64.


<details>

| field | value |
| --- | --- |
| `agent_identity` | dev3.direct.omp.f689zvm8 |
| `session` | dev3.f689zvm8 |
| `agent_persona` | generalist |
| `agent_supervisor` | unavailable |
| `agent_tool` | OMP |
| `agent_tool_version` | 18.1.2 |
| `agent_runtime` | OMP 18.1.2 |
| `tooling_profile` | dotfiles@0e52305-dirty |
</details>

PR #2809, opened 2026-05-08
Export devenv integrations for NixOS and home-manager

### LINKED ISSUE #2808: Expose hook files in nixpkgs for declarative sourcing
The directions for the new auto-activation feature https://devenv.sh/auto-activation/ indicate users should add the `devenv hook | source` to their configuration file.

This certainly works, but in more declarative setups like home-manager or nixos the extra invocation of devenv to pipe to source is unnecessary.

## Request

Could the hook shell scripts be exposed in the package's share like the current completion files are now, so that they can be sourced directly without needing to run `devenv hook fish | source`?

## Some related context

Home-Manager exposes a function (`lib.hm.shell.mkFishIntegrationOption` which is a wrapper around [`mkShellIntegrationOption`](https://github.com/nix-community/home-manager/blob/release-25.11/modules/lib/shell.nix#L5) which is an example of this declarative style.

Of note, the [direnv version](https://github.com/nix-community/home-manager/blob/release-25.11/modules/programs/direnv.nix#L78-L96) of this actually just adds [`pkg hook fish | source`](https://github.com/nix-community/home-manager/blob/release-25.11/modules/programs/direnv.nix#L145-L151), so there's precedent for the `| source` method, though I'd prefer to just use the file directly unless it's interpolated some way that is necessary by devenv.

sandydoo:
- Exports all of the shell hooks in the Nix build under `$out/share/devenv/shell-integration/<shell>/`.
- Exports NixOS and home-manager modules that configure the auto-loading `devenv shell` hooks for each shell.

Fixes #2808.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@
 
 ### Improvements
 
-- Added `nixosModules.default` and `homeManagerModules.default` flake outputs that expose `programs.devenv.enable` and per-shell integration toggles. The package now also installs hook scripts under `$out/share/devenv/shell-integration/<shell>/hook.<ext>` so they can be sourced directly.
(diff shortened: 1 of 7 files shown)


sandydoo:
The home manager module has already been upstreamed.

domenkozar:
We should also upstream NixOS one?

PR #2815, opened 2026-05-12
hook: do not exit outer shell on cd-out when DEVENV_ROOT is set externally

### LINKED ISSUE #2805: Terminal window closes when cd-ing out of project directory when using "devenv allow" or "direnv allow"
**Describe the bug**

When I am in a directory with devenv env and have enabled "devenv allow" or "direnv allow" I can't `cd` out of the project directory anymore (inside still works). Then I `cd` somewhere else the terminal window/tab closes. I tested with konsole and ghostty.

I use fish. Zoxide cd integration is turned off. 

home-manager config:

```nix
    programs = {
      fish = {
        enable = true;
        shellInit = ''
          devenv hook fish | source
        '';
      };
    };
```

When I remove the `devenv hook fish | source` integration the issue is gone.

**To reproduce**

- `cd` into project with devenv env, where you have enabled "devenv allow" or "direnv allow"
- `cd ..` or `cd /etc`
- See that terminal closes

**Version**

devenv 2.1.0 (x86_64-linux)

michaelglass:
## Summary

- The shell hooks (`devenv hook fish`/`bash`/`zsh`) call `exit` when the user cd's out of a project, intended to drop them out of the hook-spawned inner shell. The "am I in an inner shell?" check used `DEVENV_ROOT` alone — but direnv (and similar tools) export `DEVENV_ROOT` into the user's outer shell. With that env var set, cd-out called `exit` in the outer shell too, closing the terminal.
- Gates the exit on `_DEVENV_HOOK_DIR` (set only on shells the hook spawned) in addition to `DEVENV_ROOT`. The fish hook already sets `_DEVENV_HOOK_DIR` as the cd target; the posix hook now sets it on the spawn line as well.
- Fixes #2805.

an agent wrote the code (and most of this post) but a human being (me) has confirmed it fixes the issue.

## Test plan

- [x] New integration test `devenv/tests/hook_outer_shell_survives.rs` — bash and fish variants, each covering outer-survives + inner-still-exits
- [x] Verified red against unpatched hooks (both outer-shell tests fail), green with the fix
- [x] Reproduced the original bug interactively in fish before the fix, confirmed gone after
- [x] `cargo test -p devenv commands::hook::` — existing hook unit tests still pass
- [x] `cargo fmt --check` clean
- [ ] CI green


sandydoo:
@domenkozar this change causes our `devenv hook` integration to launch a second nested shell when you launch a shell manually.

Issue #2655, opened 2026-03-23
Container `devenv/devenv` doubled in size in v2


Allsimon:
**Describe the bug**
`ghcr.io/cachix/devenv/devenv:v1.11.2` weighs 444MB
`ghcr.io/cachix/devenv/devenv:v2.0.2` weighs 980MB

It looks like most of it comes from the `llvm-21.1.8-lib` layer (526 MB).

AmentumStudio:
> **Describe the bug** `ghcr.io/cachix/devenv/devenv:v1.11.2` weighs 444MB `ghcr.io/cachix/devenv/devenv:v2.0.2` weighs 980MB
> 
> It looks like most of it comes from the `llvm-21.1.8-lib` layer (526 MB).

A (slightly naive) quick look at this:
```sh
( \
  set -eux; \
  oci_optimise_devenv() { \
    VERSION_DEVENV=${1}; \
    docker run \
      -it \
      --rm \
      -v /var/run/docker.sock:/var/run/docker.sock \
      -- \
      dslim/slim \
        build \
        --target ghcr.io/cachix/devenv/devenv:v"${VERSION_DEVENV}" \
        --tag devenv:v"${VERSION_DEVENV}"-slim \
        --http-probe=false \
        --exec "devenv init; devenv update; devenv test;" \
    ; \
  }; \
  oci_optimise_devenv 1.11.2; \
  oci_optimise_devenv 2.0.2; \
)
```
Results in:
1. `devenv:v1.11.2-slim` image, that is `96MB`.
1. `devenv:v2.0.2-slim` image, that is `271MB`.

In my experience, no nix-based solutions to create OCI images is _even close_ to being space-efficient.

Why is this "slightly naive"? <https://github.com/slimtoolkit/slim> does the slimming by analyzing what parts of container are actually accessed during runtime, and shedding stuff that is not. For this to work, you either have to expose your API via http (and use the http probe), or you need to pass `--exec` with everything you'd like the thing to do.

So the ideal place to put this would be as a runner for E2E tests. Whatever is tested, will be available inside the container. If tests have good coverage, then you're pretty much sure everything will work, and the size will be smaller.

After changing the `--exec` in my snipper above 
from
`--exec "devenv init; devenv update; devenv test;"`
to
`--exec "devenv init; devenv update; devenv test; devenv shell< <(echo 'exit 0;'); devenv build; devenv tasks list;"`,
[...]

PR #2838, opened 2026-05-16
feat: disable LSP in devenv docker image

Allsimon:
The changes gate the lsp feature behind #[cfg(feature = "lsp")] in Rust and build a devenv-for-container variant without nixd in the PATH wrapper, keeping nixd (and its 540 Mo LLVM dep) out of the image closure entirely.

see #2655


Allsimon:
This is probably a bad idea... Nixd itself isn't that large, the problem comes to their packaging of LLVM. I'll try to fix it upstream

PR #2839, opened 2026-05-18
hook: do not spawn nested devenv shell when DEVENV_ROOT is already set

michaelglass:
## Summary

- Follow-up to #2815 (reported by @sandydoo): running `devenv shell` manually spawned a second nested shell. With `DEVENV_ROOT` set but no `_DEVENV_HOOK_DIR`, the hook fell through to activation.
- after this PR: 
  - If DEVENV_ROOT` set → hook does nothing.
  - if `_DEVENV_HOOK_DIR` _additionally_ set → `exit` on cd-out so the parent shell can follow.

## Test plan

- [x] Bash + fish regression tests shim `devenv` on PATH and assert no `hook-should-activate` / `shell` call when `DEVENV_ROOT` is set and PWD matches
- [x] Verified red against current main, green with the fix
- [x] Existing #2805 outer-survives and inner-still-exits tests still pass
- [ ] CI green


michaelglass:
sorry I realize I missed the nushell hook.  Will make one more PR.

PR #2844, opened 2026-05-19
hook: extend nu hook to match bash/zsh/fish behavior, add zsh + nu tests

michaelglass:
follow up to #2839, extending it to nushell. cc @sandydoo 

Extracted the nu's  closure body into a named `def --env _devenv_hook []` function. Nu's `env_change.PWD` hook only fires in interactive sessions, so tests can't trigger it via `cd` in script mode — calling the named function directly makes the logic testable.

Tests:
- bash/zsh/fish iterate over a shared template (their syntax is close enough that one body covers all three). Zsh was previously covered only transitively via hook.posix.sh; now exercised explicitly.
- nu lives in its own three tests sharing a `run_nu` helper. Folding it into the posix loop required more abstraction than it saved.


devenv/tests/hook_outer_shell_survives.rs
```diff
@@ -1,60 +1,76 @@
...
     [
-        ("bash", bash_src, bash_path_override as fn(&Path) -> String),
-        ("fish", fish_src, fish_path_override as fn(&Path) -> String),
+        (
+            "bash",
+            format!(r#"eval "$({bin} hook bash)""#),
+            posix_path_override as fn(&Path) -> String,
+        ),
+        (
+            "zsh",
```
michaelglass, line 51: note: this is kind of orthogonal to the PR. Just adds zsh tests as well.

PR #2846, opened 2026-05-19
bug: improve garage service and startup + tests

### LINKED ISSUE #2840: bug: garage service does not execute `configureScript` with `process-compose`
**Describe the bug**

The configureScript in https://github.com/cachix/devenv/blob/main/src/modules/services/garage.nix#L192
is not executed with `process-compose` due to that beeing a `task` etc.

@ap-1, @domenkozar  : Does that `configureScript` get executed automatically?? with devenv's process managers?

@ap-1: The tests `.test.sh` also does not test if the bucket is there etc, so its a bit too simple IMO.

**Version**

`main`

gabyx:
Closes [#2840](https://github.com/cachix/devenv/issues/2840)

### DIFF
--- a/src/modules/services/garage.nix
+++ b/src/modules/services/garage.nix
@@ -224,6 +224,27 @@ in
         Additional `garage.toml` snippet appended to the generated config.
       '';
     };
+
+    ui = {
+      enable = lib.mkEnableOption "Enable a simple web UI.";
+
+      start = lib.mkOption {
+        type = types.bool;
+        default = false;
+        description = ''
+          If the service is by default started or must be manually started.
+        '';
+      };
+
+      port = lib.mkOption {
+        type = types.port;
+        default = 3919;
+        example = 3919;
+        description = ''
+          On which port the UI should run.
+        '';
+      };
+    };
   };
 
   config = lib.mkIf cfg.enable {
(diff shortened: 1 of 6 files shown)


src/modules/services/garage.nix
```diff
@@ -166,11 +171,10 @@ in
       ports.rpc.allocate = 3901;
       exec = "exec ${cfg.package}/bin/garage -c ${configFile} server";
       ready.exec = ''
-        ${pkgs.curl}/bin/curl -sf \
-          -H "Authorization: Bearer ${cfg.adminToken}" \
-          "http://${adminHost}:${toString allocatedAdminPort}/v1/health"
+        ${pkgs.curl}/bin/curl -f -H 'Authorization: Bearer ${cfg.adminToken}' 'http://${adminHost}:${toString allocatedAdminPort}/v1/health'
       '';
-      before = [ "devenv:garage:configure" ];
```
gabyx, line 173: This task is never executed. Also the tests do not test if the buckets are actually created.

gabyx:
@sandydoo: Thanks for merging this. Improves Garage, and adds the missing bucket test.

gabyx:
@sandydoo : this needs further improvements. I am working on it. since the whole setup is not good.

also `lib.writeShellScriptBin` should ahve been used for user friendliness with shellcheck etc etc...

gabyx:
Is good to be merged, IMO.

PR #2848, opened 2026-05-19
eval-cache: skip storing results whose inputs changed mid-evaluation

### LINKED ISSUE #2745: devenv often doesn’t update tasks when definitions change

domenkozar:
## Summary

Fixes #2745: `devenv tasks list` (and similar commands) could persistently return stale results when `devenv.nix` was edited while a previous evaluation was still in flight, requiring users to delete `.devenv/nix-eval-cache.db*` to recover.

## Root cause

In `devenv-eval-cache::caching_eval::CachedEval::eval{,_typed}`, the post-eval snapshot is built by `FileInputDesc::new`, which reads the **current** on-disk mtime and content hash for every tracked file. If the file was rewritten between when Nix read it and when we snapshotted it, the stored row paired:

- `result` = evaluation of the **old** contents
- `input_hash` = hash of the **new** contents

The next lookup validated against the new on-disk hash, matched, returned the stale result indefinitely.

## Fix

Capture `eval_start: SystemTime` immediately before running the evaluator. After snapshotting, re-stat each tracked input with sub-second precision; if any mtime is strictly newer than `eval_start`, refuse to store. The check is intentionally conservative: false positives only cost an extra re-eval next time; false negatives let the bug recur.

Deliberately **not** added: a speculative `invalidate(key)` on the skip path. It provides no correctness benefit (any inherited stale row is already rejected by validation on the next lookup, since its stored hash no longer matches disk) and creates a multi-process race where a concurrent process B writing a valid entry for the same key would have its row erased by A's invalidate.

## Test plan

- [x] Regression test (`test_input_modified_during_eval_skips_cache`) exercises `eval()` end to end with a tracked file whose mtime is bumped during the closure, asserting no row is stored.
- [x] Two unit tests pin `any_input_modified_after` on both sides of the threshold using explicit `set_modified` timestamps (no sleeps).
- [x] `cargo nextest run -p devenv-eval-cache` — 43/43 pass.
- [x] `cargo build` — full workspace builds clean.
- [x] `cargo clippy -p devenv-eval-cache --tests` — clean (only a pre-existing warning in `devenv-activity`, unrelated).
- [ ] Manually verify against the reporter's `mre/repro.sh` from https://github.com/vibecoder43/devenv_repro with `DEVENV_BIN=$(cargo build && echo target/debug/devenv) DELAY=3.0 ./mre/repro.sh` — expects "SUMMARY: second tasks list contained demo:show".

## Limitations

Pre-existing stale entries from buggy versions are not cleaned up retroactively — they live until the file's mtime/hash changes from the stored values, at which point the existing validation invalidates them. Users already affected by #2745 may still need a one-time `.devenv/nix-eval-cache.db*` wipe.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Fixed `devenv up` corrupting a running daemon's PID file and socket when started in the foreground, leaving the daemon unmanageable. Foreground `up` now rejects with "Processes already running" when a daemon is active.
 - Fixed shell hook spawning a nested `devenv shell` when `devenv shell` was entered manually. Follow-up to [#2815](https://github.com/cachix/devenv/pull/2815).
 - Fixed "zoxide: infinite loop detected" when using `zoxide init --cmd=cd fish` and `cd`-ing into a devenv project. The fish hook now defers spawning `devenv shell` to the next prompt instead of spawning inline inside the PWD event handler, so in-progress shell state never leaks into the devenv shell ([#2841](https://github.com/cachix/devenv/issues/2841)).
+- Fixed stale task and option results that lingered when `devenv.nix` was edited while a command was already evaluating. The eval cache no longer stores results whose tracked input files were modified mid-evaluation, so the next run no longer needs `.devenv/nix-eval-cache.db*` to be wiped to see the new definitions ([#2745](https://github.com/cachix/devenv/issues/2745)).
 
 ### Improvements
 
--- a/devenv-eval-cache/src/caching_eval.rs
+++ b/devenv-eval-cache/src/caching_eval.rs
@@ -597,6 +597,39 @@ impl CachedEval {
         }
     }
 
+    /// Persist a cache miss, refusing to store if any tracked input file was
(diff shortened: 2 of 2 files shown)


domenkozar:
Verified locally against the reporter's reproducer (https://github.com/vibecoder43/devenv_repro) using `target/debug/devenv` from this branch:

- `DELAY=3.0`: `SUMMARY: second tasks list contained demo:show` ✓
- `DELAY=5.0`: `SUMMARY: second tasks list contained demo:show` ✓

Both delays previously failed on the reporter's `HEAD 5a8d9c9` (the commit immediately before this branch).

PR #2849, opened 2026-05-19
Always surface the Nix eval error, never a stale warning

### LINKED ISSUE #2820: unhelpful error message on syntax error
**Describe the bug**
When the `devenv.nix` file contains a syntax error, `devenv shell` fails with the misleading error message
```
Failed to get attribute 'config.cachix.enable' from devenv configuration: warning: Ignoring the client-specified setting 'system', because it is a restricted setting and you are not a trusted user
```

**To reproduce**
```
> mkdir test; cd test
> devenv init
> echo '{ pkgs, lib, config, inputs, ... }:
{
  # https://devenv.sh/basics/
  env.GREET = "devenv" # MISSING SEMICOLON HERE
}' > devenv.nix
> devenv shell
✓ Validating lock 4 files                                                                                                                                        470ms
✓ Configuring shell                                                                                                                                               25ms
  × Failed to get attribute 'config.cachix.enable' from devenv configuration: warning: Ignoring the client-specified setting 'system', because it is a restricted setting and you are not a trusted user
```

**Version**
```
devenv 2.1.0 (x86_64-linux)
```

domenkozar:
## Summary

Fixes #2820: when Nix emitted a warning at error verbosity (e.g. `Ignoring the client-specified setting 'system'` for untrusted users), devenv selected that log line as the error message and dropped the real error returned by eval. Syntax errors and actionable suggestions like `devenv inputs add …` were hidden behind the warning.

Root cause (thanks @sandydoo): the warning's color changed, `looks_like_warning` stopped matching, and the fallback treated the warning line as the main error.

## What changed

- The error returned by eval is now always rendered, in full and in Nix's natural order: `--show-trace` frames first, the actionable `error: …` paragraph last, at the bottom of the terminal where the cursor lands.
- Dropped the brittle log-line heuristics entirely: `select_raw_error`, `looks_like_warning`, and `peek_pre_repl_errors` are gone. There is no path left where a captured log line replaces the eval error, and no color matching that can rot.
- The only remaining text processing is dedenting the uniform left margin the C bindings add.

An earlier revision of this PR reordered the output to put the actionable paragraph above the trace; that was reverted per review since it forced scrolling up past the trace to find it.

## Tests

- Unit tests on `format_eval_error`: eval error always rendered (warning-shadow regression), natural trace-then-error order preserved, bare syntax error, no-`error:`-line fallback, dedenting.
- New `tests/missing-input-error/` integration test: the `devenv inputs add` suggestion is printed below the trace frames, and the headline never carries the stale warning.
- `tests/syntax-error/.test.sh` gains an assertion that the warning does not shadow the syntax error.

## Test plan

- [x] `cargo nextest run -p devenv-nix-backend error::`
- [x] `devenv-run-tests run --only syntax-error --only missing-input-error tests`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug Fixes
 
+- Fixed misleading evaluation errors where evaluation warnings replaced the real error message, hiding syntax errors and suggestions like `devenv inputs add …` ([#2820](https://github.com/cachix/devenv/issues/2820)).
 - Fixed `devenv shell` failing with `Failed to parse output store path … is not in the Nix store` when using a relocated/chroot Nix store (e.g. `NIX_REMOTE='local?root=…'` or a bind-mounted store). The realized shell path was translated to its physical location and then fed back into store-path parsing, which only accepts the logical `/nix/store/…` form; the logical path is now recovered before creating the GC root ([#2499](https://github.com/cachix/devenv/issues/2499)).
 - Fixed the TUI crashing on activity names or store paths containing multi-byte characters (e.g. non-ASCII package names or evaluation paths) when shortened for narrow terminals.
 - Fixed the TUI crashing with an arithmetic overflow when a download or task reported progress beyond its expected total.
--- a/devenv-nix-backend/src/error.rs
+++ b/devenv-nix-backend/src/error.rs
@@ -1,31 +1,26 @@
 //! Helpers for shaping Nix evaluation errors into miette diagnostics.
+//!
+//! These work on the Nix C bindings' already-rendered error text. There is no
+//! structured error type exposed by `nix-bindings-rust` to key off; the
+//! bindings stringify the C++ exception's `what()` into a single message and
+//! return it as `anyhow::Error`.
+//!
+//! The eval-returned error is always rendered in full and in Nix's natural
+//! order: `--show-trace` frames first, the actionable `error: …` paragraph
+//! last. That keeps the actionable message at the bottom of the terminal
+//! where the cursor lands, instead of requiring the user to scroll up past
+//! the trace to find it.
 
-/// Returns true if the message looks like a Nix warning (e.g. starts with
-/// `warning:`). Some Nix warnings (such as restricted-settings notices) get
-/// emitted through the FFI logger at the error verbosity, which would otherwise
-/// shadow the actual evaluation error we want to surface.
-pub(crate) fn looks_like_warning(msg: &str) -> bool {
-    let trimmed = msg.trim_start();
-    trimmed.starts_with("warning:") || trimmed.starts_with("\u{1b}[33;1mwarning:")
-}
-
-/// Pick the most useful raw error text to display.
+/// Shape a raw Nix error string into a `MietteDiagnostic` for rendering.
 ///
-/// Prefer the most recent entry from `nix_errors` that is not a warning —
-/// when Nix logs a full tree-style trace through the FFI logger, we want to
-/// surface that. If no real error was logged (e.g. a syntax error where the
-/// diagnostic only arrives via the FFI return value), fall back to
-/// `ffi_fallback`.
-pub(crate) fn select_raw_error<'a>(
-    nix_errors: &'a [String],
-    ffi_fallback: impl FnOnce() -> String,
-) -> std::borrow::Cow<'a, str> {
-    nix_errors
-        .iter()
-        .rev()
-        .find(|msg| !looks_like_warning(msg))
-        .map(|s| std::borrow::Cow::Borrowed(s.as_str()))
-        .unwrap_or_else(|| std::borrow::Cow::Owned(ffi_fallback()))
+/// Pure function over the FFI return text so it can be unit-tested without
+/// running an actual evaluation. The caller wraps the result with
+/// `Report::from(..)` to get a `miette::Error`.
+///
+/// The raw text is rendered as-is (modulo dedenting) — no reordering, no
+/// filtering — so whatever eval returned is shown in full.
+pub(crate) fn format_eval_error(raw: &str, context: &str) -> miette::MietteDiagnostic {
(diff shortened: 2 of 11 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,9 @@
...
+- Surfaced the actionable `error: …` paragraph from Nix evaluation diagnostics as the headline instead of burying it under ~100 lines of `--show-trace` frames. For example, the "use `devenv inputs add …`" suggestion that follows from a missing input in `devenv.yaml` now appears next to the "Failed to …" line; the full trace moves to the `help:` section ([#2820](https://github.com/cachix/devenv/issues/2820)).
+- Fixed import path validation rejecting valid input-relative imports on macOS, where `TMPDIR` resolves via the `/var → /private/var` symlink (`Import path '<name>/...' resolves outside the git repository`). `validate_within_root` now canonicalizes the longest existing ancestor of the path before comparing against the root.
+- Fixed `devenv inputs add` from a subdirectory writing to a stray `devenv.yaml` in the subdir instead of the enclosing project. It now walks up to find `devenv.nix` the same way `devenv shell` does, so the input is added where the rest of devenv reads it.
```
sandydoo: Looks like some other project dir fixes got bundled in

devenv/src/main.rs
```diff
@@ -208,6 +213,27 @@ impl TestDirs {
     }
 }
 
+/// For commands dispatched before `resolve()` (e.g. `inputs add`): if invoked
+/// from a subdirectory of a project, chdir up to the directory containing
+/// `devenv.nix` so the command operates on the enclosing project, matching
+/// where `devenv shell` would run. Best-effort — a no-op if discovery or the
+/// chdir fails.
+fn enter_discovered_project_root() {
```
sandydoo: Worth adding some traces here for the failure modes.

I'd also think about whether we might want to resolve the project root earlier in the flow. Maybe `devenv allow` also wants the project root? And do we even want to chdir at all?

devenv-nix-backend/src/error.rs
```diff
@@ -1,31 +1,104 @@
...
+/// Match rule: an `error:` line counts as a paragraph boundary only when it
+/// is preceded by a blank line (or sits at the start of input). Once there is
+/// leading trace/noise before the first matching `error:` line, that first
+/// match is the Nix boundary; everything after it belongs to the user-facing
+/// error block, even if a user-authored multi-line `throw` contains later
+/// blank-line paragraphs that start with `error:`.
+///
+/// Returns `("", text)` when no qualifying `error:` line is found, so callers
+/// can fall back to a single-block rendering.
+pub(crate) fn split_trailing_error(text: &str) -> (&str, &str) {
```
sandydoo: Option would be nice here. Would require changes further down.
```suggestion
pub(crate) fn split_trailing_error(text: &str) -> (Option<&str>, &str) {
```

sandydoo (commented):
👍

CHANGELOG.md
domenkozar: Good catch — split out into separate PRs. The `devenv inputs add` subdir fix is now #2883, and the macOS import-path validation fix is #2884. This PR is now just the error-surfacing work for #2820.

devenv/src/main.rs
domenkozar: Added `debug`/`warn` traces for the three failure modes (unreadable cwd, no enclosing `devenv.nix`, and a found-but-failed chdir, which is the one that actually leaves `inputs add` writing to the wrong place). Since this `devenv inputs add` change is unrelated to error surfacing, I moved it to #2883 — the traces live there now.

On the design questions: agreed that resolving the project root once, earlier in the flow, would be cleaner than the two ad-hoc discovery sites, and `devenv allow` probably wants it too (incl. whether we should chdir at all vs. threading the root through). I'd rather not expand the split-out PR's scope, so I captured it as a follow-up note on #2883.

devenv-nix-backend/src/error.rs
domenkozar: Done in d4770c26 — `split_trailing_error` now returns `(Option<&str>, &str)` and `format_eval_error` matches on it directly. A trace that trims down to empty (input opens with `error:`, or only blank lines precede the first `error:`) maps to `None` rather than `Some("")`, via a small shared `split_at` helper.

CHANGELOG.md
domenkozar: I split them into separate PRs

sandydoo:
I've figured out the cause for #2820.

The color of the warning in the logs changed, causing `looks_like_warning` to return false.
We then defaulted to treating the warning line as an error and selected it as the main error, ignoring the real error returned by eval (the `getInput` suggestion).
This particularly affects untrusted users because Nix will reliably print out a bunch of warnings in that cause.

We need to:

- Always print out return eval errors (those returned by calling eval)
- Avoid defaulting to error level if parsing fails

As for this PR, it currently prints the suggestion above the trace, requiring users to scroll up to find it. What we had before was better.

domenkozar:
Addressed in c0174f8e:

- **Always render the eval-returned error**: already the case since the first commit dropped `select_raw_error`/`looks_like_warning`/`peek_pre_repl_errors`, so no captured log line can replace the eval error anymore, and no color matching is left to rot.
- **No defaulting to error level on parse failure**: the only remaining color-based check is `is_nix_error()` in the log bridge, and it fails toward Debug (demoting), not Error, so a format change there can't promote a warning into an error.
- **Ordering reverted**: the error is rendered as-is in Nix's natural order again, trace frames first, the actionable `error: …` paragraph last at the bottom of the output. The headline/`help:` reshaping and `split_trailing_error` are gone. The only text processing left is dedenting the uniform left margin from the C bindings.

The integration tests now assert the suggestion appears below the trace frames and that the stale warning never shadows the real error.

sandydoo:
Rebased, plus some minor cleanup.

PR #2855, opened 2026-05-20
hook: clear stale exit-dir state

sweenu:
Disclaimer: PR made with Claude and reviewed by me

## Problem
Exiting the devenv shell was `cd`ing me into a "random" directory, which was in fact the directory in `.devenv/exit-dir` that hadn't gotten cleanup correctly because my `rm` command in fish is an alias to `echo "Stop using rm, put in the trash"`.

## Summary of changes
- bypass shell-level `rm` overrides when the shell hooks clean up `.devenv/exit-dir`
- clear any leftover `.devenv/exit-dir` before launching a new shell, so a failed cleanup cannot redirect a later shell exit to an old path
- add a bash/fish regression test for stale exit-dir cleanup and align the nushell hook with the same cleanup strategy

## Test plan

- [x] `devenv shell -- cargo fmt --check`
- [x] `devenv shell -- cargo nextest run -p devenv --test hook_outer_shell_survives`


domenkozar:
#2882 for a bit thorough fix

PR #2859, opened 2026-05-20
feat: Add `--include-envrc` flags to `init` command

sestrella:
I recently upgraded to the latest devenv and noticed that `.envrc` is no longer generated by `devenv init`. I also came across the page announcing [native auto-activation without direnv](https://devenv.sh/integrations/direnv/), which includes the note:

> .envrc is not created by devenv init. You need to create it manually.

This PR aims to make `.envrc` generation more convenient by avoiding the need to manually copy and paste the snippet from the direnv documentation. Instead, it adds support for generating `.envrc` as part of `devenv init` when the `--include-envrc` flag is explicitly provided.

**Why not generate `.envrc` by default during `devenv init`?**

I’m not fully aware of the reasons behind removing `.envrc` from the default scaffolding. To stay aligned with the current behavior, this PR does not reintroduce it by default but instead makes it opt-in via a flag.

That said, the flag name (`--include-envrc`) may be somewhat confusing, and I’m open to suggestions for a clearer alternative.

### DIFF
--- a/devenv/src/cli.rs
+++ b/devenv/src/cli.rs
@@ -719,10 +719,10 @@ pub enum Commands {
     #[command(about = "Scaffold devenv.yaml, devenv.nix, and .gitignore.")]
     Init {
         target: Option<PathBuf>,
+        #[arg(long, help = "Skip .envrc file (default).")]
+        skip_envrc: bool,
         #[arg(long, help = "Include .envrc file.")]
-        envrc: bool,
-        #[arg(long, help = "Skip .envrc file.")]
-        no_envrc: bool,
+        no_skip_envrc: bool,
     },
 
     #[command(about = "Generate devenv.yaml and devenv.nix using AI")]
--- a/devenv/src/cli.rs
+++ b/devenv/src/cli.rs
@@ -717,7 +717,13 @@ fn is_flag(arg: &std::ffi::OsString) -> bool {
 #[derive(Subcommand, Clone)]
 pub enum Commands {
     #[command(about = "Scaffold devenv.yaml, devenv.nix, and .gitignore.")]
-    Init { target: Option<PathBuf> },
+    Init {
+        target: Option<PathBuf>,
+        #[arg(long, help = "Include .envrc file.")]
+        envrc: bool,
+        #[arg(long, help = "Skip .envrc file.")]
+        no_envrc: bool,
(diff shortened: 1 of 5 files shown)


devenv/src/cli.rs
```diff
@@ -717,7 +717,19 @@ fn is_flag(arg: &std::ffi::OsString) -> bool {
...
+        #[arg(
+            long,
+            conflicts_with = "no_skip_envrc",
+            help = "Skip .envrc file (default).",
+            env = "DEVENV_SKIP_ENVRC",
+            value_parser = clap::builder::BoolishValueParser::new()
+         )]
+        skip_envrc: bool,
+        #[arg(long, conflicts_with = "skip_envrc", help = "Include .envrc file.")]
+        no_skip_envrc: bool,
```
domenkozar: Shouldn't we also need this flag if it's false by default?

sestrella: @domenkozar done! I replaced `--skip-envrc` and `--no-skip-envrc` with a single `--include-envrc` flag.

joscha:
The documentation says that this works:

SCR-20260609-jyoe

however 2.1.2 as of today does not seem to respect the flag:

SCR-20260609-jvff

and the `tip` to pass it as a value just results a new devenv initialization in a subdirectory named `--include-envrc`. All in all very confusing. Shall we update the documentation until this landed?

domenkozar:
Thanks for flagging this @joscha. Good news: it's not a bug, just a release timing issue.

The `--include-envrc` flag was merged in this PR on 2026-05-21, but v2.1.2 (which you're on) was released on 2026-05-13, so the flag doesn't exist yet in your build. On 2.1.2, `devenv init --include-envrc` hits clap's "unexpected argument" error, and the suggested `-- --include-envrc` makes it a positional `target`, which is why you got a subdirectory named `--include-envrc`.

The flag is on `main` now and will ship in **2.2**.

I've updated the docs to add a "New in version 2.2" note so the page no longer implies it works on current releases, with a pointer to the manual `.envrc` snippet for earlier versions: 82287166.

AmentumStudio:
Glad that https://github.com/cachix/devenv/pull/2644 was eventually implemented, even if by someone else.
Second time's the charm it would seem.

PR #2867, opened 2026-05-22
shell: exit cleanly on terminal hangup instead of orphaning

### LINKED ISSUE #2845: Orphaned .devenv-wrap processes persist after VS Code terminal sessions, consume 100%+ CPU
**Describe the bug**

Stale `.devenv-wrap` processes persist after VS Code terminal sessions end, each consuming 100-150% CPU. They are not cleaned up when the terminal session is interrupted, when VS Code is quit, or after stopping devenv services manually.

On my MacBook Air M5 (macOS, aarch64-darwin), I discovered three such processes running simultaneously after roughly two weeks of normal use:

```
PID    COMMAND      %CPU  TIME
3583   .devenv-wrap 144.2 16:06:49
96920  .devenv-wrap 144.0 31:13:34
51030  .devenv-wrap 125.0 13:57:12
```

Combined, they consumed ~400% CPU continuously, with PID 96920 having accumulated 31 hours of CPU time. This caused severe battery drain — under 2 hours of battery life on a machine that should comfortably get 8+ hours. The processes were only discovered by running `top -o cpu`; they were not obvious in Activity Monitor's standard app view.

A `killall .devenv-wrap` cleared them, and they did not respawn after reboot, which suggests they are orphaned rather than being relaunched by a supervisor.

**To reproduce**

I'm unable to provide a minimal reproduction because I'm not a developer and I'm not sure what specifically triggered the orphaning. I can describe my usage pattern in case it helps narrow it down:

- I use `devenv` inside VS Code's integrated terminal.
- Typically 1-2 VS Code windows open at a time, with one terminal per window.
- When I finish working with a service or process, I stop it with Ctrl+C in the terminal.
- I sometimes keep VS Code open for extended periods (across days) rather than quitting it fully between sessions.
- I do not explicitly run any devenv cleanup command on session end.

Claude's hypothesis is that something about long-lived VS Code sessions, terminal interrupts (Ctrl+C), or both, can leave the `.devenv-wrap` wrapper process orphaned. Over time, these accumulate and consume significant CPU.

My hypothesis is that Claude Code opens the shell when testing and then doesn't shut it down.

I'm happy to provide any additional diagnostic information if you can tell me what to capture next time it happens.

**Version**

devenv 2.1.1+a3ebee0 (aarch64-darwin)

domenkozar:
## Summary

Fixes [#2845](https://github.com/cachix/devenv/issues/2845): `devenv shell` could linger as a background process, often pinned at 100%+ CPU, after its terminal window or tab was closed.

**Root cause:** devenv runs several tokio runtimes (renderer, backend, owner) and `tokio::signal` ties delivery to a single runtime's driver, so SIGHUP/SIGINT/SIGTERM were silently dropped while a shell session held the runtime. The shell was never told to exit, and the inner shell (in its own PTY session) outlived it.

**Fix:**
- Handle SIGINT/SIGTERM/SIGHUP on a dedicated `signal-hook` OS thread — runtime-independent and `sigaction`-based (no signal-mask blocking, so child processes keep normal signal behavior). On a signal it aborts any in-flight Nix eval; if a shell is active it kills the inner shell and exits, otherwise drives graceful shutdown (second signal force-exits).
- A POLLHUP watchdog on stdin detects controlling-terminal hangup directly (belt-and-suspenders — a closed terminal is exactly what a closed VS Code tab does).
- Teardown threads use `_exit` rather than `process::exit`, because running atexit handlers (Boehm GC / Nix) from a non-GC-registered thread can deadlock.
- Hardened the session so a VT-thread error always kills the PTY and notifies the coordinator, instead of orphaning the inner shell and hanging coordinator teardown.
- Added a regression test (`terminate_active_shell_kills_registered_inner_shell`).

## Test plan

Verified on Linux with a PTY harness driving the real event loop — all teardown paths exit in ~1s and kill the inner shell, no orphan/spin/hang:

- [x] Close controlling terminal → exit 129, inner shell gone
- [x] `kill -HUP` → exit 129, inner shell gone
- [x] `kill -TERM` → exit 143, inner shell gone
- [x] Normal `exit` → exit 0 (graceful), inner shell gone
- [x] `devenv shell -- echo` still works (rc 0, no regression)
- [x] `cargo nextest -p devenv-shell` → 131/131 pass; fmt + clippy clean
- [ ] **macOS** smoke test: close a VS Code tab running `devenv shell`, confirm no `.devenv-wrap` lingers
- [ ] Ctrl-C during `devenv up` / `build` / `test` (signal handling now applies to all commands)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Fixed stale task and option results that lingered when `devenv.nix` was edited while a command was already evaluating. The eval cache no longer stores results whose tracked input files were modified mid-evaluation, so the next run no longer needs `.devenv/nix-eval-cache.db*` to be wiped to see the new definitions ([#2745](https://github.com/cachix/devenv/issues/2745)).
 - Fixed long lines in `devenv shell` getting a hard newline inserted at the wrap point when copying to clipboard. The shell now preserves the soft-wrap when flushing wrapped output into the terminal's scrollback, so clipboard copy keeps the original single line ([#2865](https://github.com/cachix/devenv/issues/2865)).
 - Fixed files declared with the `files` option not being regenerated when an auto-loaded (`devenv allow`) shell reloaded after `devenv update`. enterShell tasks (including `devenv:files`) now re-run on hot-reload, matching a fresh shell entry, instead of only updating environment variables ([#2864](https://github.com/cachix/devenv/issues/2864)).
+- Fixed `devenv shell` lingering as a background process, often pinned at 100%+ CPU, after its terminal window or tab was closed. The shell now detects the controlling-terminal hangup and exits, terminating the inner shell with it, instead of orphaning and draining the battery ([#2845](https://github.com/cachix/devenv/issues/2845)).
 
 ### Improvements
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1843,6 +1843,7 @@ dependencies = [
  "serde",
  "serde_json",
  "shell-escape",
+ "signal-hook 0.4.4",
  "similar 3.1.0",
  "sqlx",
  "tempfile",
--- a/devenv-shell/src/lib.rs
+++ b/devenv-shell/src/lib.rs
@@ -36,7 +36,9 @@ pub use status_line::{
 };
 
 // Main session
-pub use session::{SessionConfig, SessionError, SessionIo, ShellSession, TuiHandoff};
+pub use session::{
+    SessionConfig, SessionError, SessionIo, ShellSession, TuiHandoff, terminate_active_shell,
+};
 
 // Re-export for convenience
 pub use portable_pty::{CommandBuilder, PtySize};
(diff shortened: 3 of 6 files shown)


domenkozar:
Verified on real macOS (arm64, macOS 26.3.1, Apple M1) — built the branch with `cargo build` and ran the PTY harness against it:

| teardown | result |
|---|---|
| close controlling terminal (= closed VS Code tab) | devenv exits (code 129), inner shell killed ✅ |
| `kill -HUP` | devenv exits (code 129), inner shell killed ✅ |
| `kill -TERM` | devenv exits (code 143), inner shell killed ✅ |

No orphaned `devenv`/inner shell after any teardown. The crate also compiles cleanly on Darwin (`signal-hook`, `libc::_exit`, the `#[cfg(unix)]` POLLHUP watchdog).

Note: this exercises tearing down a **running** shell (the reported scenario). Closing the terminal *during* the initial build/eval (before the interactive prompt is up) goes through a different path and isn't covered here.

sandydoo:
I'm not sure we need another signal handler function when we already install signals via `tokio-shutdown`.
_And this ^ should be installed early on in main too, not in the backend thread._

Maybe the issue is that the reload shell doesn't query `shutdown`? I don't think we need so much wiring here.

PR #2883, opened 2026-05-29
cli: walk up to project root for `devenv inputs add`

domenkozar:
Split out of #2849 (the actionable-error PR bundled this unrelated project-dir fix).

## Problem

`Inputs::Add` is dispatched before `resolve()` runs, so the parent-directory discovery that `devenv shell` benefits from never ran for it. Running `devenv inputs add` from a subdirectory silently created a stray `devenv.yaml` in the subdir that no other command would read.

## Fix

Apply the same chdir-up-to-`devenv.nix` step in the dispatch arm so the input lands in the enclosing project's config, matching where `devenv shell` would run. The discovery's failure modes (unreadable cwd, no enclosing `devenv.nix`, failed chdir) now emit `debug`/`warn` traces so the best-effort behaviour is observable.

Also compares against the physical path in the `cli-subdir-discovery` test so it passes on macOS, where `TMPDIR` resolves via the `/var -> /private/var` symlink.


domenkozar:
Design follow-up carried over from @sandydoo's review on #2849:

Project-root discovery currently happens at two ad-hoc sites — here in the `inputs add` dispatch arm, and again in `resolve()` for `devenv shell` et al. Worth considering as a follow-up:

- Resolving the project root once, earlier in the flow, and reusing it everywhere.
- `devenv allow` likely wants the same root.
- Whether we should `chdir` at all, or thread the discovered root through instead of mutating the process cwd.

Out of scope for this fix; noting it here so it isn't lost.

PR #2890, opened 2026-06-01
feat(processes): `start = "interactive-shell"` and attachable `devenv up`

### LINKED ISSUE #2863: Add command to start project and go to shell
In every development session, I use `devenv up -d && devenv shell` to run the project and access required tools. When I'm done, I have to exit twice: 1) exit the shell and 2) run `devenv processes down`.

This process could be streamlined into a single command that:

- starts the project
- drops the user into a shell with project tools
- on exit, stops processes and exits the shell

domenkozar:
## Summary

Adds `processes.<name>.start.enable = "interactive-shell"` (and a global `process.start` default): such processes start when you enter an interactive `devenv shell` and stop when you exit, replacing the `devenv up -d && devenv shell` two-step plus the matching `exit` + `devenv processes down`.

`devenv up` now **attaches** to an already-running native manager (from `devenv up -d` or `devenv shell`) over its control socket and starts the up-enabled processes, instead of failing with "Processes already running". Dependency ordering (`after`/`before`) is driven by the daemon's own task scheduler via a new `ApiRequest::Up` forwarded to `Tasks::start_with_deps`, so out-of-subset and already-running dependencies resolve against the live task graph exactly like a cold start.

The interactive shell status line shows a live count of running processes (e.g. `watching 5 files | 3 processes`).

Native process manager only.

Fixes #2863.

## What's here
- `start.enable` accepts `"interactive-shell"`; new `process.start` global default; assertion requiring the native manager.
- `devenv up` attaches via `ApiRequest::Up` → daemon scheduler (`start_with_deps`); honours the positional subset and dependency ordering.
- `--background` daemon mode writes the PID file early so shell entry isn't blocked on process readiness.
- Live process-count status line.

## Fixes folded in (from self-review)
- **Relaunch a self-exited/crashed process** on attaching `devenv up`. Such a process stays `Active` with a terminal phase, which `rearm_waiting` refused to touch, so it was silently never relaunched. `start_with_deps` now normalizes it via `stop_and_keep` first. (+ regression test)
- **Register the `Up` handler before the daemon serves requests.** It was set only after every auto-started process became ready, and the foreground in-process `up` never set it at all, so an attaching `devenv up` during startup (or to a foreground `up`) was rejected with "no process scheduler to handle up". Both paths now go through a shared `run_foreground_with_up` helper.

## Test plan
- `cargo nextest run -p devenv-tasks -p devenv-processes` — 249 passed (incl. new `start_with_deps_relaunches_exited_process`).
- `nix build` — passes.
- Integration tests `tests/process-up-attach`, `tests/process-up-attach-deps`, `tests/process-start-shell` — not yet run locally.

## Known follow-ups (not blocking this draft)
- Non-interactive foreground `devenv up` against a running manager currently streams until interrupted; gate on a TTY so scripts fail fast.
- `devenv up <name>` reports success even if the running daemon's graph lacks `<name>` (return/diff the scheduled set).
- Decide whether an explicit `devenv up <name>` should start a `start.enable = false` process (attach path currently force-starts it; cold start does not).
- A self-exited dependency makes `devenv up <dependent>` cancel the dependent (`Exited@ready` is `NeverSatisfiable`); deliberate semantics decision.


omarcinkonis:
Is per-process starting required or a nice to have? How much can we shrink the PR if we keep the global `process.start` only? Is the extra complexity and maintenance burden really justifiable?

domenkozar:
We know people want to start some processes at shell and potentially others later on.

omarcinkonis:
The way I see it, it is important to address user concerns, but it's also important to keep the project maintainable. Adding 1000 lines of code, scattered over so many files and introducing several new edge cases, all for one QoL feature - that does not seem maintainable.

I'd advise reconsidering the solution. Maybe it's possible to add it on top somehow, rather than modify the core functionality of the native process manager?

omarcinkonis:
I tried implementing per-process starting with minimized changes to the process manager. See the 2nd commit on my PR, I'll also leave a summary there https://github.com/cachix/devenv/pull/2912

Maybe it's worth discussing this on a call? I may join devenv.sh Discord if you see the need.

domenkozar:
I've split the attaching feature to https://github.com/cachix/devenv/pull/2936

PR #2891, opened 2026-06-02
fix(processes): keep file watcher alive for one-shot processes

GregorLohaus:
devenv's process `watch` is driven by the per-process supervisor task, which dropped its FileWatcher the moment a non-restarting process exited. For one-shot commands (for example code generators) the command exits immediately, so the watcher was torn down almost as soon as `devenv up` started and later file edits never triggered a re-run — `watch` silently did nothing for them.

Park the supervisor after a clean exit when watch paths are configured, instead of breaking out of the loop. The `to_wait()` select arm is gated on the Exited phase so an already-finished command doesn't busy-spin. A file change then revives the process via the existing `Event::FileChange => Action::Restart` path (the state machine already supported restarting from a terminal phase; the supervisor just exited before it could). The process is still reported as Exited, so `@completed` / `@started` dependencies remain satisfied. Non-watched one-shots keep terminating exactly as before.

Adds a regression test (test_oneshot_reruns_on_file_change) that fails without the fix: the one-shot never re-runs and the test times out.


GregorLohaus:
🙏🏻

Issue #2440, opened 2026-01-27
Manage gitignore


jashan-lco:
This is partly a question and partly a request. I'm `nix flake template`-ing a few projects that already have `.gitignore` files. It was getting tedious to edit them by hand to add devenv related entries to that file, so I created a task in a common devenv like:

```nix
          tasks = {
            "devenv-k8s:gitignore" = {
              exec = ''
                grep -qF ".devenv" .gitignore || echo ".devenv" >> .gitignore
                grep -qF ".pre-commit-config.yaml" .gitignore || echo ".pre-commit-config.yaml" >> .gitignore
                grep -qF ".direnv" .gitignore || echo ".direnv" >> .gitignore
              '';
              before = [ "devenv:enterShell" "devenv:enterTest" ];
            };
          };
```

It could be cleaned up a bit, but do you see any issues with doing something like this?

And if not, could something like this be provided as a config option by devenv? i.e, to ensure a list of lines exist (or maybe even don't exist?) in `.gitignore`

domenkozar:
Yes, would be great to have a gitignore managed by devenv!

jashan-lco:
I'm guessing this is just going to be another integration module here https://github.com/cachix/devenv/tree/main/src/modules/integrations.

Should I just take a stab at an implementation or do you have any ideas on the design?

I'd just like to keep it simple and add something like this to cover my use case:

```nix
{ pkgs, ... }:

{
    gitignore.enable = true;
    gitignore.ensurePresent = [ ".devenv" ".pre-commit-config.yaml" ".direnv" ];
}
```

PR #2895, opened 2026-06-03
Add support for gitignore generation integration

shikanime:
This implementation has been dogfooded for a long time, and I'm relatively confident that sharing it with the rest of the Devenv community would be useful.

Related: https://github.com/cachix/devenv/issues/2440


domenkozar:
I'm not sure we should support the templates via gitnr url format, because that means that then we're tied to that package for generating the gitignore files.

I'm fine if that's under integrations.gitnr

shikanime:
@domenkozar Good point, for regular `.gitignore` generation we should just use the `files` api.

PR #2896, opened 2026-06-03
Regenerate Cargo.nix to fix nix profile add build failure

GregorLohaus:
## Description

  Installing devenv via `nix profile add git+https://github.com/cachix/devenv#default`
  (or `#devenv`) currently fails to build with:

      error[E0433]: cannot find module or crate `detect_coding_agent` in this scope
          --> src/main.rs:1233:9
      1233 |         detect_coding_agent::is_agent()
           |         ^^^^^^^^^^^^^^^^^^^ use of unresolved module or unlinked crate

  ## Root cause

  Commit ca49f158 ("Use detect-coding-agent crate for AI agent detection") added
  `detect-coding-agent` to `Cargo.toml` and `Cargo.lock`, but the crate2nix-generated
  `Cargo.nix` — which the Nix build actually uses to resolve crate dependencies — was
  not regenerated. As a result the crate was never linked, so the FFI/Nix build of the
  `devenv` binary failed even though `cargo build` succeeds.

  ## Fix

  Regenerated `Cargo.nix` from the current `Cargo.lock`:

      devenv tasks run devenv:crate2nix   # crate2nix generate -h nix/crate-hashes.json

  This adds the missing `detect-coding-agent` dependency to the generated build graph.

  > Note: the diff is large because regenerating the file also picks up unrelated drift
  > that had accumulated since the last regeneration. The functional change is the
  > addition of the `detect-coding-agent` crate.


GregorLohaus:
the removed lines are from the task formatting Cargo.nix differently so most "arrays" are a single line now

domenkozar:
fixed on main

PR #2904, opened 2026-06-04
fix(modules): set module system class

BirdeeHub:
This allows modules to know if they are in a devenv module or one from another class, and allows devenv to give a reasonable warning upon importing modules from home manager or nixos that will not work here.

One consideration to keep in mind is that we may wish to set `_class = "devenv";` in the modules we provide here as well.

This is not necessary, however, as modules with `_class = null;`, which is the default, are always allowed.

What setting _class in the modules we provide would do, would be that it would prevent them from being imported in home manager and nixos at top level, where they would not work.

The reason this PR is being made, is that [nix-wrapper-modules](https://github.com/BirdeeHub/nix-wrapper-modules) allows specification of options for target module systems that you will be installing them in. However, [those install options](https://birdeehub.github.io/nix-wrapper-modules/md/getting-started.html#nixos-home-manager-and-friends) [require `_class`](https://github.com/BirdeeHub/nix-wrapper-modules/blob/6d9506b09f8fbe0429153da1cebef2bd61a98848/lib/core.nix#L177-L371) in order to figure out which one it is in.

Setting class for devenv modules is something I believe to make sense, as devenv provides many modules and all are compatible with one another but not other module systems.

### DIFF
--- a/devenv-nix-backend/bootstrap/bootstrapLib.nix
+++ b/devenv-nix-backend/bootstrap/bootstrapLib.nix
@@ -214,6 +214,7 @@ rec {
 
       # Phase 1: Base evaluation to extract profile definitions
       baseProject = lib.evalModules {
+        class = "devenv";
         specialArgs = inputs // {
           inherit inputs secretspec primops;
         };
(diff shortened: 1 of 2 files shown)


BirdeeHub:
I am not entirely sure what the title of this commit should be so I went with `fix(modules)`?

I see no CONTRIBUTING.md either only a claude md? It says to update the changelog, should I be doing that? Or is that instruction specific to claude?

PR #2911, opened 2026-06-06
Fix nushell activation script

### LINKED ISSUE #2905: Nushell auto-activation hook fails when cache directory does not exist
**Describe the bug**

The [auto-activation setup script](https://devenv.sh/auto-activation/#setup) for nushell fails when the `~/.cache/devenv` directory does not exist. The script needs to be updated to create the directory first.

**To reproduce**

Add the setup script to your nushell config, and ensure that the `~/.cache/devenv` directory does not exist. Nushell then fails to load the config on startup, as the hook file cannot be created and sourced.

**Version**

devenv 2.1.2 (x86_64-linux)

LlamaLamp:
Fix https://github.com/cachix/devenv/issues/2905

The [previous fix](https://github.com/cachix/devenv/pull/2906) did not actually solve the issue - in my testing I must have mistakenly created the script manually at some point, which fixed the issue. The actual root problem is that the hook script must be created before the nushell script is parsed. See the nushell documenation for an explanation: https://www.nushell.sh/book/thinking_in_nu.html#example-dynamically-generating-source

This fix instead creates the devenv hook as an autoload script, which will be run by nushell later in its setup process.

### DIFF
--- a/devenv/hooks/hook.nu
+++ b/devenv/hooks/hook.nu
@@ -1,8 +1,7 @@
 # devenv hook for nushell
 # Usage: Add to your config.nu:
-#   source (devenv hook nu | save --force ~/.cache/devenv/hook.nu; "~/.cache/devenv/hook.nu")
-# Or: devenv hook nu | save --force ~/.cache/devenv/hook.nu
-#     source ~/.cache/devenv/hook.nu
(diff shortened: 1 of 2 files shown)


domenkozar:
Could you rebase?

LlamaLamp:
Apologies, rebased now

PR #2918, opened 2026-06-09
perf(shell): speed up nested full-screen TUIs

domenkozar:
## Summary

`devenv shell` now avoids two major sources of rendering overhead:

- The mediated renderer consumes libghostty dirty-row snapshots and skips rows whose cells have not changed.
- Once an application is already on the alternate screen, its output is forwarded directly to the real terminal instead of reconstructing every frame cell-by-cell. The final status-line row remains reserved and protected.

This substantially reduces work for nested full-screen TUIs such as Neovim and Claude Code, especially on larger terminals, without removing the `devenv shell` status line.

## Root cause

Every PTY batch was previously fed through the VT and then re-read across the visible screen. Even a one-line update therefore scaled with terminal area. Full-screen applications also already emit efficient cursor-addressed updates, so reconstructing their output from VT cells duplicated work and fragmented repaint bursts.

## Passthrough correctness

Direct terminal output is handled by a stateful byte-stream rewriter that:

- clamps cursor rows and scrolling regions to the content area above the status line;
- intercepts terminal-size queries and in-band resize mode where the child must see PTY dimensions;
- buffers split CSI, OSC, DCS, APC, PM, SOS, and UTF-8 sequences so host output cannot be inserted inside them;
- safely cancels oversized terminal strings before unavoidable host output;
- reasserts the protected scroll region after DECSTR/RIS resets;
- tracks origin mode, autowrap, keypad state, synchronized output, and forwarded DEC modes for redraw and exit cleanup;
- defers host overlays while the application owns synchronized-output mode or terminal state makes cursor restoration unsafe.

The branch was rebased onto current `main`, preserving the newer tmux title-filtering path in both single and batched PTY reads.

## Validation

- `devenv shell cargo test --workspace`
- `devenv shell cargo test -p devenv-shell` (187 tests)
- `devenv shell cargo clippy -p devenv-shell --lib -- -D warnings`
- `RUSTDOCFLAGS='-D warnings' devenv shell cargo doc -p devenv-shell --all-features --no-deps`
- `devenv shell cargo fmt --all -- --check`
- `git diff --check`

Regression tests cover dirty-row rendering, full-screen scroll invalidation, split control strings and UTF-8, aborted-string recovery, APC/Kitty graphics boundaries, oversized-string cancellation, DECSTR/RIS state resets, synchronized output, and terminal cleanup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,7 @@
 
 ### Improvements
 
+- Sped up `devenv shell` when running nested full-screen TUIs (e.g. neovim, claude-code), where rendering previously lagged on larger terminals. Each frame now reads and redraws only the rows that actually changed instead of the whole screen, so a single-line update no longer costs a full-screen repaint.
 - `DEVENV_HOME` now overrides where devenv stores all per-user data (GC roots, trust database, cached keys), not just the trust database.
 - `DEVENV_RUNTIME` can now be set to override where a project stores its sockets and other runtime files. To relocate runtime files for all projects at once, prefer `XDG_RUNTIME_DIR`, which keeps each project's directory separate.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,7 +48,7 @@
 
 ### Improvements
 
-- Sped up `devenv shell` when running nested full-screen TUIs (e.g. neovim, claude-code), where rendering previously lagged on larger terminals. Each frame now reads and redraws only the rows that actually changed instead of the whole screen, so a single-line update no longer costs a full-screen repaint.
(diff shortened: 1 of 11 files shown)


domenkozar:
I'm still not happy with this, claude scrollbar is slow

AodhanHayter:
I took this branch for a test drive, it does feel better, but really starts to struggle as soon as I hit the bounds of a pane and some kind of scroll happens. I owe you a video as it's more clear what kind of renders it struggles with watching a session.

It's definitely not to the point that I'd willingly run nvim or any other TUI within the devenv TUI itself, there's a perfectly fine workaround with `devenv --no-tui --no-reload`. Although, I could see this catching new users off guard and them not understanding why their system is feeling sluggish all of the sudden. It took me a bit to narrow down devenv as the culprit myself.

I appreciate you burning some tokens for my sake. I can also live with just using devenv a bit differently than I have in the past.

domenkozar:
I want to fix this before we get 2.2 out.

domenkozar:
Split into #3025 and https://github.com/cachix/devenv/pull/3028

PR #2925, opened 2026-06-10
fix(git-hooks): install hooks with absolute config path

### LINKED ISSUE #2924: Githooks are not properly installed if the devenv root is not the repository root
**Describe the bug**

when the devenv root is not the project root, precommit hook is not properly installed

**To reproduce**

https://github.com/Daholli/devenvbug/tree/main
After entering the shell and trying to run git commit you get this error message 
```
❮ git commit
error: config file not found: `.pre-commit-config.yaml`
- To temporarily silence this, run `PREK_ALLOW_NO_CONFIG=1 git ...`
- To permanently silence this, install hooks with the `--allow-missing-config` flag
- To uninstall hooks, run `prek uninstall`
```

if I run prek install it installs fine, but the denvenv shell does not install this githook properly

**Version**

❮ devenv --version
devenv 2.1.3+8228716 (x86_64-linux)

domenkozar:
Fixes #2924.

## Problem

When the devenv root is a subdirectory of the git repository root, git hooks fail at commit time:

```
error: config file not found: `.pre-commit-config.yaml`
```

The `devenv:git-hooks:install` task ran `prek install -c .pre-commit-config.yaml` from the devenv root, and prek embeds that path verbatim into `.git/hooks/pre-commit`. Git always runs hooks with cwd set to the worktree toplevel, so the relative path resolved against the repo root, where the generated config does not exist.

Manually running `prek install` worked because without `-c` prek falls back to workspace discovery (it embeds `--cd` instead of a config path), which is why the issue only bit the devenv-managed install.

## Fix

Pass an absolute config path (`$DEVENV_ROOT/<configPath>`) to `install` and `run`. This keeps working with a custom `git-hooks.configPath` (which prek's workspace discovery would not find) and with `git-hooks.package` overridden back to Python pre-commit (which has no workspace mode). Since devenv reinstalls hooks on every shell entry, the embedded absolute path cannot go stale in practice.

## Tests

Added `tests/git-hooks-subdirectory-root`: git repo at the top, devenv project in a `java/` subdirectory, enters the shell from the subdir, commits from the repo root, and asserts the hook ran. The failure mode was reproduced standalone with prek 0.3.11 (relative `-c` from a subdir breaks commits from the repo root; absolute `-c` fixes it). The new test passes with the patched modules, and the 5 existing git-hooks tests still pass.


Daholli:
Thanks for the quick fix!

PR #2927, opened 2026-06-10
fix(processes): make the runtime directory independent of TMPDIR

### LINKED ISSUE #2923: A `devenv shell` terminal and Claude Code can't find each other's running processes
Hi! We're trying to setup devenv (really enjoying it 🙏) on our stack. I hit a
confusing case where `devenv up -d` in one terminal and `devenv processes …` in
another end up not seeing each other, leaving orphaned processes and sometimes
two copies of the stack. I traced it to `$TMPDIR` and wanted to ask whether
this is expected — or if there's a recommended way to make the runtime dir
stable.

**Describe the bug**

The process-manager runtime directory (where the daemon socket/pid live) is
derived from `$TMPDIR`: `$DEVENV_RUNTIME = $TMPDIR/devenv-<hash>`. The
`<hash>` is stable (keyed on the project), but the **base is `$TMPDIR`**.
So if `$TMPDIR` differs between invocations, the second `devenv` call computes a
*different* runtime dir, can't find the running daemon
(`No process manager is running`), and the processes started by the first call
are left orphaned. A second `devenv up -d` from the other `$TMPDIR` then starts
a **second, independent** instance.

We hit this with **Claude Code**:

- **Terminal A** — a normal shell with the project's devenv environment active
  (via `devenv shell` / direnv). The dev starts the stack:
  ```console
  $ devenv up -d
  # A's $TMPDIR is the macOS default; daemon lands under /tmp/devenv-<hash>/processes/
  ```
- **Claude Code**, launched *from that same shell* — so it inherits the devenv
  environment, **including `$DEVENV_RUNTIME`**. But Claude runs its shell
  commands with a sandboxed `TMPDIR=/tmp/claude-501`. When it invokes `devenv`:
  ```console
  $ devenv processes list           # from Claude's context (TMPDIR=/tmp/claude-501)
  × No process manager is running. Start processes first with `devenv up -d`
  ```
  It doesn't see A's stack — it recomputes the runtime dir as
  `/tmp/claude-501/devenv-<hash>`. A `devenv up -d` from here starts a **second**
  stack; a `devenv processes down` from here can't stop A's (they orphan), and
  vice-versa. A and B are mutually invisible.

Two things that surprised me:
- **The inherited `$DEVENV_RUNTIME` is ignored.** Even though Claude inherits
  `DEVENV_RUNTIME=<A's value>` in its environment, devenv recomputes the runtime
  dir from `$TMPDIR` rather than honoring the exported variable.
- On macOS it's easy to miss in everyday use: the default `$TMPDIR`
  (`/var/folders/…/T/`) appears to normalize to `/tmp` for the runtime dir, so
  the problem only appears once something sets a custom `$TMPDIR` (an agent like
  Claude Code, some `direnv`/CI setups, etc.).

**To reproduce**

Agent-free repro — a second shell that just exports `TMPDIR=/tmp/claude-501`
behaves exactly like Claude Code's subprocess.

```nix
# devenv.nix
{ ... }: {
  processes.sleeper.exec = "sleep infinity";
}
```

```yaml
# devenv.yaml
inputs:
  nixpkgs:
```

```console
$ devenv allow

# The runtime dir base follows $TMPDIR (note: same hash, only the base differs):
$ TMPDIR=/tmp/dvA devenv info | grep DEVENV_RUNTIME
- DEVENV_RUNTIME: /tmp/dvA/devenv-94396f1
$ TMPDIR=/tmp/dvB devenv info | grep DEVENV_RUNTIME
- DEVENV_RUNTIME: /tmp/dvB/devenv-94396f1

# --- Terminal A (normal shell, default $TMPDIR) ---
$ devenv up -d
# 'sleeper' running; daemon under /tmp/devenv-<hash>/processes/

# --- "Claude" shell (what Claude Code does to its subprocesses) ---
$ export TMPDIR=/tmp/claude-501

# Inherited DEVENV_RUNTIME is ignored — still recomputes from $TMPDIR and misses A:
$ DEVENV_RUNTIME=/tmp/devenv-<hash> devenv processes list
× No process manager is running. Start processes first with `devenv up -d`

$ devenv up -d                     # doesn't see A → starts a SECOND instance
$ pgrep -fl 'sleep infinity'
<two 'sleep infinity' processes>   # A's and B's, independent

# Each side sees only its own; `devenv processes down` only stops its own,
# leaving the other orphaned.
```

**Expected:** the two shells manage the *same* daemon (or there's a documented
way to pin the runtime dir), regardless of `$TMPDIR`.
**Actual:** the daemon is keyed on a `$TMPDIR`-derived path, so a changed
`$TMPDIR` → mutual invisibility, orphans, and double-starts.

**Version**

- `devenv 2.1.3+d5e9138` (aarch64-darwin)
- Determinate `Nix 3.21.0` (`nix 2.34.6`)
- macOS (Darwin 23.3.0, arm64)
- default `native` process manager (also relevant to `process-compose`, whose
  socket defaults to `${config.devenv.runtime}/pc.sock`).

**Questions / notes**

- Is the runtime dir intentionally based on `$TMPDIR`? For multi-instance /
  multi-worktree isolation we'd have expected it keyed on the project path
  (`devenv.state`) only — which the `devenv-<hash>` suffix already is; it's the
  `$TMPDIR` *base* that makes it environment-dependent. Would keying the base on
  `XDG_RUNTIME_DIR` (when set) and otherwise a fixed location be reasonable?
  Could an exported `$DEVENV_RUNTIME` be honored when present?
- Our current workaround, in case it helps others — wrap the `devenv` CLI so
  every invocation runs with a normalized `TMPDIR`, and (for the process-compose
  backend) pin the control socket to a `$TMPDIR`-independent path:
  ```nix
  # devenv/shared.nix
  scripts.devenv.exec = ''
    export _DEVENV_WRAPPED=1
    real="$(type -ap devenv | sed -n '2p')"
    exec env TMPDIR=/tmp "$real" "$@"
  '';

  process.managers.process-compose.unixSocket.path =
    let h = builtins.substring 0 7 (builtins.hashString "sha256" config.devenv.state);
    in "/tmp/devenv-pc-${h}.sock";
  ```
  Both work, but a stable runtime dir out of the box (or a documented knob)
  would be lovely.

Happy to help to patch and/or test, if it is indeed a bug.
Thanks a lot the hard work.

_(PS: The debug session and report have been done with the assistance of Claude Code, feel free to point out any errors or inaccurate things i may have missed)_

domenkozar:
Fixes #2923.

## Problem

`DEVENV_RUNTIME` was derived from `XDG_RUNTIME_DIR || TMPDIR || /tmp`. Although its hash suffix was project-stable, using `$TMPDIR` as the base made the complete path invocation-dependent. Sandboxed tools such as Claude Code, `nix develop`, and some CI environments can change `$TMPDIR`, so two commands for the same project resolved different process-manager sockets and could start duplicate or orphaned process stacks.

Reading the exported `$DEVENV_RUNTIME` back as configuration is not safe either: it leaks through nested shells and makes it ambiguous whether a child invocation should reuse the parent project's or profile's runtime directory. `DEVENV_RUNTIME` therefore remains an output-only variable.

## Changes

- Runtime directories now resolve from `$XDG_RUNTIME_DIR`, falling back to `/tmp`; `$TMPDIR` is never used for runtime rendezvous.
- `$DEVENV_RUNTIME` is ignored as an input. Users who need to relocate runtime files should set `$XDG_RUNTIME_DIR`.
- The Rust resolver and direct-flake Nix default both hash `devenv.dotfile`, so CLI and `nix develop` compute the same path.
- Profiles retain separate runtime directories because each profile has a distinct dotfile path.
- Documentation and the changelog now describe `DEVENV_RUNTIME` as read-only and document the migration behavior.

## User impact

Independent invocations for the same project now find the same process manager even when their `$TMPDIR` values differ. Projects and profiles remain isolated. A daemon still using its previous runtime path may need to be restarted once after upgrading.

## Testing

- `cargo nextest run -p devenv-core` — 204 tests passed.
- `cargo check -p devenv-core -p devenv-processes -p devenv`.
- `devenv-run-tests run tests --only cli-flake-bare`, covering different `$TMPDIR` values, ignored inherited `$DEVENV_RUNTIME`, and an explicit `$XDG_RUNTIME_DIR`.
- Rust formatting, Nix parsing/formatting, shell syntax, and `git diff --check`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@
 - Fixed the TUI crashing with an arithmetic overflow when a download or task reported progress beyond its expected total.
 - Fixed the quit confirmation prompt overflowing past the right edge on terminals narrower than ~82 columns; the prompt now adapts to the available width.
 - Fixed the bottom navigation hint bar overflowing the right edge on standard ~80 column terminals when a process was selected; the hints are now more compact on narrower terminals.
+- Fixed devenv invocations with different `$TMPDIR` values resolving different runtime directories and therefore failing to find each other's process manager. Runtime directories now resolve from `$XDG_RUNTIME_DIR`, then `/tmp`; `$TMPDIR` is never used. A daemon still using its previous runtime path may need to be restarted once ([#2923](https://github.com/cachix/devenv/issues/2923)).
 - Fixed edits to local `path:` inputs (e.g. `url: path:../shared-config`) never taking effect: the evaluation cache served the old configuration until `.devenv` was deleted, and `devenv update` did not help. Files from local path inputs are now read directly from disk and tracked by the cache, so changes apply on the next command and direnv reloads when they change.
 - Fixed the evaluation cache missing a `devenv.local.nix` (or any other tracked file) created within the same second as the previous evaluation, which could happen in scripts that run devenv commands back to back.
 - Fixed editing `enterShell` while a `devenv shell` is running causing a parse error (e.g. `(eval):6: parse error near '\n'`) on reload under zsh and fish. The `enterShell` output was being mixed into the environment data the reload applies; it now goes to the terminal as it does on a fresh shell entry ([#2919](https://github.com/cachix/devenv/issues/2919)).
@@ -57,8 +58,6 @@
 
 - Reduced renderer fragmentation in `devenv shell` by reading PTY output in larger batches, lowering syscall and event-allocation overhead during full-screen repaint bursts.
 - `DEVENV_HOME` now overrides where devenv stores all per-user data (GC roots, trust database, cached keys), not just the trust database.
-- `DEVENV_RUNTIME` can now be set to override where a project stores its sockets and other runtime files. To relocate runtime files for all projects at once, prefer `XDG_RUNTIME_DIR`, which keeps each project's directory separate.
-
 - Non-TUI console output is now buffered and flushed in batches, reducing write overhead during verbose evaluation.
 
 - Traces now identify whether devenv was invoked by the CLI, direnv, or the native shell hook with a `devenv.caller` span attribute. Caller information is passed explicitly by integrations, so nested commands are not mistaken for automatic activation ([#2965](https://github.com/cachix/devenv/issues/2965)).
--- a/devenv-core/src/paths.rs
+++ b/devenv-core/src/paths.rs
@@ -2,6 +2,7 @@
 
 use miette::{Result, miette};
 use sha2::{Digest, Sha256};
+use std::ffi::OsString;
 use std::path::{Path, PathBuf};
 
 #[derive(Debug, Clone)]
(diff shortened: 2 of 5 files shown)


domenkozar:
Closing this as superseded by 76c7f8bc, which implements the focused fix for the reported Claude Code scenario by honoring an inherited `DEVENV_RUNTIME`.

The broader work in this PR—making the default fully independent of `TMPDIR`, migration to legacy daemons, cross-platform UID resolution, and shared-`/tmp` hardening—should be handled as separate changes. Nested-project isolation and direct regression coverage are also being kept focused rather than carrying this full runtime-directory policy forward.

Issue #2923 remains open for the stronger case where independent invocations have different `TMPDIR` values and no inherited runtime state.

domenkozar:
Correction to my previous comment: I reopened this PR because 76c7f8bc only fixes the inherited-`DEVENV_RUNTIME` path. The stronger reproduction in #2923—two invocations with different `TMPDIR` values and no inherited runtime—still resolves different directories on current `main`.

A focused replacement is now implemented locally: both the Rust resolver and the existing Nix default resolve `DEVENV_RUNTIME` from a same-project override, then `XDG_RUNTIME_DIR`, then `/tmp`; `TMPDIR` is never part of the rendezvous path. The locally built CLI reports the same runtime for `TMPDIR=/tmp/dvA` and `TMPDIR=/tmp/claude-501`, and the nested-project guard plus resolver regression tests pass.

Leaving this PR open until that complete replacement is published/landed.

sandydoo (commented):
We should undo the change I pushed that allowed reading `DEVENV_RUNTIME` as an input. Users should stick with changing `XDG_RUNTIME_DIR`. There's no safe way to both read and write that variable given nested shells, so it's the wrong approach.

Removing that constraint should then simplify this PR: remove the absolute/relative rules, the `DEVENV_DOTFILE` checks, etc.

AFAIU this PR also changes how profiles share runtime dirs?

Issue #2123, opened 2025-08-23
[feature request] add "copy mode" option to files


landure:
Please add a "copy mode" option to files, with these values:

- symbolic-link : create a symbolic link to the file in the derivation.
- create: create the file or directory by copying the contents of the derivations, and running `chmod --recursive 'u+r'` on the copied contents.
- replace: delete the existing path, and rerun create.
- overwrite / amend: for a directory, create only the missing files.

The aim is to allow using files to initialize configuration files with templates that the user can edit to fit his/her project.

landure:
a example use case is the creation of "recipes".
See the PHP recipes in [Biapy's devenv-recipes @ GitHub](https://github.com/biapy/devenv-recipes).

Allsimon:
Another use case is the $KUBECONFIG file that many tools try to overwrite when changing namespace/cluster/etc

PR #2932, opened 2026-06-13
fix: generate a devcontainer file instead of a symlink

### LINKED ISSUE #2700: `devcontainer.enable` creates symlink to /nix/store which breaks github codespaces
**Describe the bug**
Since commit 6ad6b871 `devcontainer.enable` creates `.devcontainer.json` as a symlink to `/nix/store` instead of a regular file:
```
lrwxrwxrwx 1 hilorioze users 61 Apr  5 14:57 .devcontainer.json -> /nix/store/z25m7zzxg1qlzh7ncsmvjsjrj7hd6z49-devcontainer.json
```

This breaks github codespaces because the symlink target only exists on the local machine. When codespaces clones the repo, the symlink is broken.

**To reproduce**
1. Clone the gist files from https://gist.github.com/hilorioze/e1901a14c06b3316bfa238dd9e516899 into a directory
2. Run `devenv shell`
3. Check `.devcontainer.json` - it's a symlink to `/nix/store` instead of a regular file

**Version**
devenv 2.0.6 (x86_64-linux)

First-Non-Interesting-Username:
The `.devcontainer.json` generated by the devcontainer module was previously tied to the local setup of whoever created it, making it non-functional in GitHub Codespaces and any environment not running devenv natively.

This change ensures the generated config is portable and works correctly across all developer setups, regardless of whether devenv is installed locally.

This change also changes the path of generated devcontainer config to `.devcontainer/devcontainer.json` which helps to reduce clutter in the project root of users and is generally considered to be a better option. 

fixes #2700 
partially reverts https://github.com/cachix/devenv/commit/6ad6b871fd258a1a160e263e206354ca4ae99ce6


Allsimon:
this is a good use case for https://github.com/cachix/devenv/issues/2123

First-Non-Interesting-Username:
How would you propose that to be done? I see no good options, that wouldn't require significant overhaul of files module.

First-Non-Interesting-Username:
I will look into that

src/modules/integrations/devcontainer.nix
```diff
@@ -53,6 +55,9 @@ in
   };
 
   config = lib.mkIf config.devcontainer.enable {
-    files.".devcontainer.json".json = cfg.settings;
+    enterShell = ''
+      mkdir -p ${config.env.DEVENV_ROOT}/.devcontainer
+      cat ${file} > ${config.env.DEVENV_ROOT}/.devcontainer/devcontainer.json
+    '';
```
Toxaris, line 71: Beyond copying the file instead of symlinking it, I believe this changes the behavior in the following ways:

1. `devenv tasks run devenv:files` does not re-generate the file (old behavior: `devenv:files` task handles files).
2. `devcontainer.enable = false` does not clean up the file (old behavior: clean up when `devenv:files` task runs next).
3. Existing files are silently overwritten (old behavior: print "conflicting file" but don't overwrite it).

Regarding change 1, I guess instead of adding to `enterShell` we could add a task that runs after `devenv:files`?

Regarding changes 2 and 3, I don't know how to recognize whether the file is managed by devenv or not 🤔

I am a most concerned by change 3, because it has the potential of destroying hand-crafted configuration, which `devenv` is usually avoiding.

Toxaris (commented):
Some thoughts.

src/modules/integrations/devcontainer.nix
First-Non-Interesting-Username: 1. It overwrites it, which is functionally the same from the user perspective. I implemented the devcontainer generation as a separate task
2 & 3 You won't like my solution, but I haven't thought of a better one and it works. 

I wouldn't be concerned about "destroying" anything, since any developer that uses devenv should also be using git, but fine, you're the main guy here.

Toxaris: Just to clarify, I am not a main guy at all, just got one tiny documentation PR merged yesterday. Sorry if my comment came across as overly definitive. I really just wanted to share some thoughts.

domenkozar:
> The .devcontainer.json generated by the devcontainer module was previously tied to the local setup of whoever created it, making it non-functional in GitHub Codespaces and any environment not running devenv natively.

Can you say more about this? What's the error you get?

This change feels like a lot of code and makes it difficult to maintain, so I'd like to understand if we can take an approach with less complexity.

First-Non-Interesting-Username:
Running `devenv shell` with `devcontainer.enable` set to true, creates a `.devocontainer.json` file which is a symlink to the nix store. Tools like git doesn't copy the file symlink is leading to, but symlink itself. This way, the file on github and other platforms isn't the devcontainer configuration, but a path to nix store file: https://github.com/First-Non-Interesting-Username/test/blob/main/.devcontainer.json

This file doesn't exist on systems that didn't run `devenv shell` which might be problematic for people who don't want to use devenv on host, which coincidentally are the target audience of devcontainers.

sandydoo:
> this is a good use case for #2123

This is the correct approach.

domenkozar:
https://github.com/cachix/devenv/pull/2948

First-Non-Interesting-Username:
@domenkozar should I close this PR, now when your solution is merged? (Let's be real, you're the reviewer, I see no world where you choose my solution over yours).

domenkozar:
@First-Non-Interesting-Username I've merged support for copying the path, you can now redo this PR to be way simpler.

First-Non-Interesting-Username:
Okay, should I still include that test?

domenkozar:
Just needs a comment in the fix to the issue why we need it

PR #2936, opened 2026-06-15
feat(processes): attach `devenv up` to a running process manager

### LINKED ISSUE #971: Add `devenv processes tui` to connect to detached processes

### LINKED ISSUE #2930: Cannot start single process with `devenv processes start <name>`
**Describe the bug**

`devenv processes start -h` claims that I can do `devenv processes start <name>` to start a process but it actually doesn't work.

**To reproduce**

In an empty directory:

```bash
# set up minimal devenv project with a process
echo '{ processes.some_name.exec = "echo hello"; }' > devenv.nix

# observe help
devenv processes start --help | grep Usage

# try to start process by name
devenv processes start some_name
```

Actual behavior:

```
Usage: devenv processes start [OPTIONS] [NAME]
✓ Validating lock                                                                                                                                                                          0ms
  × No process manager is running. Start processes first with `devenv processes start` or `devenv up -d`
```

Not sure what I expect, some ideas:

- Start named process in the foreground (like with `devenv processes start`)?
- Start process manager but with only the named process?
- Don't support the `[NAME]` argument at all?

**Version**

devenv 2.1.2 (x86_64-linux)

My devenv is managed by home-manager on WSL2.

domenkozar:
## Summary

Lets a second `devenv up`/`devenv processes start` **attach** to an already-running native process manager instead of failing with "Processes already running", and adds a dedicated `devenv processes attach` subcommand for a read-only live view. Along the way the process manager becomes the single owner of process phase, the daemon streams attach sessions instead of polling, and the launch path is unified behind one truthful protocol.

Closes #971, closes #2930.

## What's new

- **`devenv up` attach** — a second `up`/`up -d` against a live manager attaches over the control socket and starts the requested up-enabled processes, with `after`/`before` ordering resolved against the daemon's own task scheduler. Foreground attach streams status, ports, and logs until Ctrl-C; detached/non-interactive callers still fail fast.
- **`devenv processes attach`** — attaches to a running manager and streams a live view without starting anything, failing fast when no manager is running.
- **Cold-start from a named start** — `devenv processes start <name>` with no manager running now cold-starts one in the background launching just that process (and its dependency closure), like `devenv up -d <name>`.
- **Attach UX** — attaching announces itself in the TUI, and Ctrl-C on an attached session prompts detach (leave processes running) vs stop the whole manager.

## Internals

- **Manager owns process phase** — removed the task graph's cached phase copy and the leaking `process_status_watcher` synchronizers; dependency evaluation reads the manager's phase directly, and process-map transitions are now total. Fixes double-launch, mid-relaunch shutdown, processes vanishing from `list`, stuck-masked relaunches, watcher leaks, and two RwLock self-deadlocks.
- **Streamed attach** — replaced the 500ms List+Logs poll/diff loop with a server-pushed snapshot/status/log event stream over a new `Attach` request.
- **One launch verb** — collapsed the dep-blind `Start` and scheduler-driven `Up` into a single truthful `Start { names }` that routes every not-running→running transition through the task graph and returns per-name outcomes (scheduled/skipped/unknown/failed).
- **Review fixes** — restart ordering, transitive oneshot `wait`, listen-only readiness, terminal phase preservation in `stop_and_keep`, bounded attach channel, stale log cleanup, and name dedup.

## Testing

- Replaces `processes-detach-twice` with `process-up-attach` and `process-up-attach-deps` integration tests.
- Test helpers are notification-driven (no sleep polling).


euphemism:
Didn't get a response in the Discord discussion around this, and bringing it back up to make sure it's not a blocking concern:

How does this address the fact that the current generation of the devenv-defined environment might differ from the configuration that was initially responsible for booting up the detached processes? As in, how can `devenv` and `nix` make any guarantees about the state of the world at this point? Examples being: environment variables have changed, new ports have been defined for services, a different binary has been mandated for use in a given process, etc.

domenkozar:
> Didn't get a response in the Discord discussion around this, and bringing it back up to make sure it's not a blocking concern:
> 
> How does this address the fact that the current generation of the devenv-defined environment might differ from the configuration that was initially responsible for booting up the detached processes? As in, how can `devenv` and `nix` make any guarantees about the state of the world at this point? Examples being: environment variables have changed, new ports have been defined for services, a different binary has been mandated for use in a given process, etc.

We need to integrate reload into native process manager, that's a missing feature. I've opened https://github.com/cachix/devenv/issues/2942

PR #2938, opened 2026-06-16
Docs for containers

hugosenari:
- Add information about deps
- Make copyToRoot a suggestion when not needed
- Improve module isRunning desc and example


hugosenari:
to be honest I don't know how to regenerate docs for modules

PR #2940, opened 2026-06-16
Out-of-tree devenvs: persist --from via `devenv allow`

domenkozar:
Hey @evantravers — this came out of a conversation about making `--from` stick so you can use a devenv in a directory that has no local `devenv.nix`.

## The idea

Today `devenv --from <source>` is per-invocation: you have to repeat the flag on every command. This binds the source to the directory once, via the existing trust mechanism, so it's remembered:

```console
$ cd ~/some/empty/dir
$ devenv --from github:cachix/devenv allow
devenv: allowed /home/me/some/empty/dir from github:cachix/devenv

$ devenv shell      # no --from needed — loads from the bound source
$ cd ~/some/empty/dir   # shell hook auto-activates, just like a local project
```

## How it works (high level)

**1. `allow` remembers the source.** The trust database (`~/.local/share/devenv/allowed`) moves from plain path-per-line to **JSONL**, with an optional `from` per entry:

```jsonl
{"path":"/home/me/project"}
{"path":"/home/me/out-of-tree","from":"github:cachix/devenv"}
```

`devenv --from X allow` stores the source; the `devenv.nix`-must-exist check is relaxed when a source is given. Legacy plain and `<hash>:<path>` lines are read transparently and migrated on the next write.

**2. Every command consults the binding.** In `resolve()`, an effective source is computed before project discovery:

```
--from on CLI        →  highest priority
-O module overrides  →  next
local devenv.nix     →  next (find_project_root walks up)
allow --from binding →  fallback, only in an otherwise project-less dir
```

When the binding wins, it feeds the *same* input-injection path `--from` already used, so behavior is identical to typing the flag. Normal projects pay zero cost — `find_project_root` short-circuits before the trust DB is ever read.

**3. The shell hook auto-activates bound dirs.** `check_activation` falls back to the trust DB when there's no local `devenv.nix`. A bound directory (and its subdirectories) is trusted by the binding itself, so it activates directly on `cd`. The hook scripts are unchanged — they act on whatever directory `hook-should-activate` prints.

## Commits

- Persist `--from` (JSONL trust DB + `resolve()` consumption)
- Auto-activate bound directories on `cd`

Unit tests cover JSONL round-trip, legacy-format fallback, `from` persistence, binding resolution across subdirectories, and the in-tree-vs-binding distinction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@
 - Added a `--include-envrc` flag to `devenv init` (also settable via `DEVENV_INCLUDE_ENVRC`) to scaffold a direnv `.envrc` file. By default `devenv init` no longer creates an `.envrc` ([#2859](https://github.com/cachix/devenv/pull/2859)).
 - `devenv --from <source> allow` now binds a directory to an out-of-tree source, so you can use a devenv without a local `devenv.nix`. Every subsequent `devenv` command in that directory then loads its configuration from `<source>` without repeating `--from`, and the shell hook auto-activates the environment on `cd` just as it does for a local project.
 - `--from path:<dir>` sources now load their full configuration: the source's `devenv.yaml` (inputs and imports, including sibling imports within its git repository) is merged, and its modules are imported from the live directory so edits apply immediately without re-fetching.
+- `devenv --from <source> --profile <name> allow` also persists the profiles, so every subsequent command in the bound directory activates them automatically; explicit `--profile` flags still take priority.
 
 ### Breaking Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@
 - `devenv hook fish`/`devenv hook nu` are now loaded automatically by fish and nushell, via `share/fish/vendor_conf.d/devenv.fish` and `share/nushell/vendor/autoload/devenv.nu` respectively. Bash and zsh have no equivalent mechanism, so they still need manual configuration.
 - Bumped secretspec to 0.12, which adds a `[providers]` alias map in `secretspec.toml`, a key prefix for the AWS Secrets Manager provider, audit logging and access reasons for secret reads, and support for custom Bitwarden instances.
 - Added a `--include-envrc` flag to `devenv init` (also settable via `DEVENV_INCLUDE_ENVRC`) to scaffold a direnv `.envrc` file. By default `devenv init` no longer creates an `.envrc` ([#2859](https://github.com/cachix/devenv/pull/2859)).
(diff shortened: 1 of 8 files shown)


evantravers:
Lemme give it a swing here in a bit, this is exciting @domenkozar!

evantravers:
I like the UI. I'm running into some kinda thingy…

```
  × Failed to get drvPath from shell derivation: warning: ignoring the client-specified setting 'system', because it is a restricted setting and you are not a trusted user
```

trying to work through on my end, I'm going to try a very basic devenv just to see.

evantravers:
A very bare devenv works great. It's either because I'm using unfree, or profiles. I'll keep testing. I think it's because my devenv is very bare unless you are using one of the profiles, like I showed you today.

evantravers:
Will the trust database persist the profile, or how do you think that should work?

evantravers:
(I've been running this version of devenv for the past couple of weeks… just the way I've been running it before. I'm going to take some time today to try the new feature)

evantravers:
Tried persistent --from on a profile that imports a shared sibling common/ module — the shared config never loads. --from seems to read only the source's devenv.nix and ignore its devenv.yaml, so the inputs/imports there never apply (devenv info shows only from + nixpkgs). Moving the import to the Nix level doesn't help since the source evals in a restricted mode — ../common, absolute paths, symlinks, and git+file?dir= all get dropped; only a subdir inside the profile loads. Is ignoring the source's devenv.yaml intended?  (Written by Claude Code)

domenkozar:
> Tried persistent --from on a profile that imports a shared sibling common/ module — the shared config never loads. --from seems to read only the source's devenv.nix and ignore its devenv.yaml, so the inputs/imports there never apply (devenv info shows only from + nixpkgs). Moving the import to the Nix level doesn't help since the source evals in a restricted mode — ../common, absolute paths, symlinks, and git+file?dir= all get dropped; only a subdir inside the profile loads. Is ignoring the source's devenv.yaml intended? (Written by Claude Code)

I've pushed a fix that also remembers the profile, thanks for catching that!

evantravers:
I'm about to be traveling, but I'll give it a whirl soon.

evantravers:
I did test this a little bit. It does persist the profile in the file, and attempts to load it. However, I'm running into some kind of permissions issue where a devenv file that works with my current import based solution fails to build when I use `--from`. I'll try to poke at it some more and post a trace.

evantravers:
So playing with this today… I think I'm running into unfree packages, because the flag for that is in the remote devenv.yaml, and I don't know if the --from is reading that.

Adding a local (and gitignored) devenv.yaml with the allowed package as the error message states does seem to work, making this the first version of --from that I can kinda run. 👏

PR #2951, opened 2026-06-23
fix: create devcontainer file instead of symlinking

### LINKED ISSUE #2700: `devcontainer.enable` creates symlink to /nix/store which breaks github codespaces

First-Non-Interesting-Username:
The .devcontainer.json generated by the devcontainer module was previously tied to the local setup of whoever created it, making it non-functional in GitHub Codespaces and any environment not running devenv natively.

This change ensures the generated config is portable and works correctly across all developer setups, regardless of whether devenv is installed locally.

This change also changes the path of generated devcontainer config to .devcontainer/devcontainer.json which helps to reduce clutter in the project root of users and is generally considered to be a better, more standard option.

fixes #2700


domenkozar (changes requested):
Let's just always copy

First-Non-Interesting-Username:
Thanks for a quick review!

PR #2956, opened 2026-06-25
docs(git-hooks): information about the git-hooks input in the git-hooks page

Alexis-Fouquet:
Without hooks, we may remove (or never add) the git-hooks input. This adds a sentence about it to prevent any build failure.
I had this issue when adding hooks into my project, so I suggest a (very) small modification to avoid anyone to have the same issue. The error message was not clear about the source of the build failure.


sandydoo:
devenv normally tells users when this input is needed. The broken error message is a bug in the current release.

Happy to add this to the docs, but it should say that 1. this integration _requires_ adding git-hooks.nix as an input, and 2. provide a devenv.yaml or `devenv inputs add` command to add the input.

Alexis-Fouquet:
I have edited my modifications according to your suggestions. 

For the _requires_ I put _may_ in the first version as I supposed there is other ways to add this input. It has been replaced.

For the error message, I started using devenv recently. I saw that there is an issue opened.

Issue #2567, opened 2026-03-06
ports.*.allocate does not detect ports bound to 0.0.0.0 or [::]


bdj:
## Summary

`ports.*.allocate` only checks if `127.0.0.1:<port>` is in use when determining port availability. It does not detect ports bound to `0.0.0.0` (all IPv4 interfaces) or `[::]` (all IPv6 interfaces), causing multiple devenv instances to allocate the same port.

## Reproduction

1. Create a devenv.nix with a process that binds to `0.0.0.0`:

```nix
processes.server = {
  ports.http.allocate = 8000;
  exec = "python -m http.server ${toString config.processes.server.ports.http.value} --bind 0.0.0.0";
};
```

2. Run `devenv up -d` — it starts on port 8000
3. In a separate directory with the same devenv.nix, run `devenv up -d`
4. The second instance also gets port 8000, causing a bind conflict

## Expected behavior

Port allocation should detect that port 8000 is already in use (bound to `0.0.0.0`) and allocate the next available port (8001, etc.).

## Observed behavior

Only ports bound to `127.0.0.1` are detected. Processes bound to `0.0.0.0:8000` or `[::]:6006` are invisible to the allocator.

Evidence from `lsof` — ports devenv correctly detects as taken all bind to `localhost`:
```
mailpit     TCP localhost:1025 (LISTEN)    ← detected, worktree gets 1026
meilisear   TCP localhost:7700 (LISTEN)    ← detected, worktree gets 7701
redis-ser   TCP localhost:6379 (LISTEN)    ← detected, worktree gets 6380
```

Ports devenv fails to detect — all bind to `*` (0.0.0.0/[::]):
```
python3     TCP *:8000 (LISTEN)            ← NOT detected, worktree also gets 8000
node        TCP *:4000 (LISTEN)            ← NOT detected, worktree also gets 4000
node        TCP *:6006 (LISTEN)            ← NOT detected, worktree also gets 6006
```

## Environment

- devenv 2.0.2+6ec181f (aarch64-darwin)
- macOS Darwin 25.3.0

## Workaround

Bind all services to `127.0.0.1` instead of `0.0.0.0`. This is not ideal as it prevents access from other devices on the local network (e.g., mobile testing).

PR #2957, opened 2026-06-26
feat(processes): make port auto-allocation listen-address-aware

alechartung:
## Problem

`services.postgres.listen_addresses` lets each instance bind a different
loopback address, so several instances should be able to share `port = 5432`.
But port auto-allocation reserves each port across **all** interfaces (binding
`0.0.0.0` without `SO_REUSEADDR`) and isn't listen-address-aware, so the same
port on a *different* address is treated as taken and silently incremented to
the next free one. `strict_ports` doesn't help — it only turns the silent bump
into a hard error, since the reservation still binds the wildcard.

## Change

Thread an optional listen address through `allocatePort` →
`PortAllocator::allocate_addr` → `reserve_port`:

- A concrete IP reserves the port on **that address only**.
- Wildcards (`*`, `0.0.0.0`, `::`, empty) and non-IP hosts (e.g. `localhost`)
  keep the all-interfaces default.
- The `postgres` service forwards `services.postgres.listen_addresses`, and a new
  `processes.<name>.ports.<name>.listenAddress` option exposes this for any service.

Result: `listen_addresses = "127.0.0.2"` with `port = 5432` now binds
`127.0.0.2:5432` instead of bumping to `5433` when `5432` is held on another
address.

## Tests

Unit tests in `devenv-core` (`cargo test -p devenv-core`):
`allocate_addr_is_listen_address_aware` (same port reused on a distinct address;
the all-interfaces path still bumps) and
`is_wildcard_addr_classifies_special_listen_addresses`.

## Notes

- Backwards compatible: no `listenAddress`, or a wildcard, keeps the previous
  all-interfaces behaviour. A changelog entry is included.
- The eval-cache replay path still re-acquires across all interfaces (falling
  back to re-evaluation if the wildcard bind conflicts). Happy to scope replay
  to the listen address as a follow-up if you'd prefer.

(written by claude)

### DIFF
--- a/devenv-core/src/ports.rs
+++ b/devenv-core/src/ports.rs
@@ -206,6 +206,27 @@ impl PortAllocator {
     ///
     /// In strict mode, only tries the base port and fails with process info if unavailable.
     pub fn allocate(&self, process_name: &str, port_name: &str, base: u16) -> Result<u16, String> {
+        self.allocate_addr(process_name, port_name, base, None)
+    }
+
+    /// Like [`Self::allocate`], but reserves the port on a specific listen
+    /// address when `listen_addr` names a concrete IP.
+    ///
+    /// This makes allocation listen-address-aware: the same base port can be
+    /// used by several instances as long as each binds a different address
+    /// (e.g. Postgres on `127.0.0.2:5432` and `127.0.0.3:5432`), instead of the
+    /// allocator reserving the port across all interfaces and treating it as
+    /// globally taken — which would silently increment to the next port.
+    ///
+    /// A `None`, wildcard (`""`, `"*"`, `"0.0.0.0"`, `"::"`), or non-IP address
+    /// reserves across all interfaces — the conservative default.
+    pub fn allocate_addr(
+        &self,
+        process_name: &str,
+        port_name: &str,
+        base: u16,
+        listen_addr: Option<&str>,
+    ) -> Result<u16, String> {
         if !self.enabled.load(Ordering::SeqCst) {
             return Ok(base);
         }
@@ -235,7 +256,7 @@ impl PortAllocator {
                 ));
             }
 
-            match reserve_exact_port(base) {
+            match reserve_exact_port(listen_addr, base) {
                 Ok(listeners) => {
                     ports.insert(
                         key,
@@ -270,7 +291,7 @@ impl PortAllocator {
                 continue;
             }
 
-            let Ok(listeners) = reserve_port(port) else {
+            let Ok(listeners) = reserve_port(listen_addr, port) else {
                 continue;
             };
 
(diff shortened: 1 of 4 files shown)


alechartung:
Feel free to close if this isn't a good fit. I put this together for a proof of concept, but I'm starting to think just running the environments in different port ranges on localhost is cleaner anyways.

domenkozar:
Thanks for this, Alec, and sorry for the slow review.

Some important history first: the behaviour this PR changes is intentional. The allocator binds `0.0.0.0` without `SO_REUSEADDR` on purpose, as the fix for #2567, where checking only a specific address let two devenv instances allocate the same port and then collide at bind time. Binding the wildcard without `SO_REUSEADDR` is precisely what lets the allocator detect a port held on *any* address. The commit message here frames that behaviour as the bug to remove, so I think the change predates awareness of #2567.

The consequence is that scoping the reservation to a listen address brings back the class of miss #2567 was about. A bind probe only detects conflicts on the addresses it actually binds, so it is reliable only when the probe covers everything the process binds. For a single concrete `listen_addresses` IP that postgres binds exactly, that holds (I checked: it still detects same address and wildcard holders). But as soon as the process binds wider than the scoped address (a multi address `listen_addresses`, or the generic `ports.<name>.listenAddress` against a process that binds `0.0.0.0`), a real conflict goes undetected and the port is handed out while already in use. That is #2567 again.

There is also a concrete regression. `reserve_specific_port` only falls back to all interfaces on `InvalidInput`, but a valid address that cannot be bound right now returns `AddrNotAvailable` and propagates as a hard failure ("Could not find available port", or "already in use" under strict ports). This fires on the headline example itself: `127.0.0.2` is not bindable on stock macOS without `sudo ifconfig lo0 alias 127.0.0.2`, and the new unit test is gated to Linux (where `127.0.0.2` does bind), so CI stays green while `devenv up` would fail on macOS. More broadly, extra loopback addresses are bindable by default only on Linux, so the multi loopback sharing is effectively Linux only.

On the underlying goal: running several postgres instances at once already works without sharing a port. Auto-allocation gives each instance a distinct port (5432, 5433, and so on), and you connect via the allocated `config.processes.postgres.ports.main.value` / `PGPORT`, or over the per project unix socket that `PGHOST` points at. If you specifically need identical ports distinguished only by address, that is really an explicit "pin this exact port and address, do not probe" request, which is a different thing from the conflict-detecting auto-allocator, and I would rather design that separately if there is demand for it.

Given all that, I do not think we should take this as written, so I am going to close it. I really appreciate the effort and the clear write-up, and if you have a use case the distinct-port or unix-socket path does not cover, please do open an issue describing it so we can scope the right feature.

PR #2959, opened 2026-06-26
feat: add services.surrealdb

Felipe-9:
Adds the `services.surrealdb` module that runs the server under devenv.

### DIFF
--- a/src/modules/services/surrealdb.nix
+++ b/src/modules/services/surrealdb.nix
@@ -1,0 +1,103 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.services.surrealdb;
+
+  basePort = cfg.port;
+  allocatedPort = config.processes.surrealdb.ports.main.value;
+
+  stateDir = "${config.env.DEVENV_STATE}/surrealdb";
+in
+{
+  options.services.surrealdb = {
+    enable = lib.mkEnableOption "Surrealdb process";
+
+    package = lib.mkPackageOption pkgs "surrealdb" { };
+
+    port = lib.mkOption {
+      type = lib.types.port;
+      default = 8080;
+      description = "The TCP port to accept connection.";
+    };
+
+    username = lib.mkOption {
+      type = with lib.types; nullOr str;
+      default = null;
+      description = "Sets master username for the database";
+    };
+
+    password = lib.mkOption {
+      type = with lib.types; nullOr str;
+      default = null;
+      description = "Sets master password for the database";
+    };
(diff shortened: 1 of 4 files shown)


src/modules/services/surrealdb.nix
```diff
@@ -0,0 +1,103 @@
...
+              surreal = "${cfg.package}/bin/surreal";
+            in
+            pkgs.writeShellScriptBin "start-surreal" ''
+              set -euo pipefail
+
+              if [[ ! -d "$SURREAL_PATH" ]]; then
+                mkdir -p "$SURREAL_PATH"
+              fi
+
+              ${surreal} start ${args} memory
```
domenkozar, line 97: The server starts with the in-memory backend (`memory` positional arg), so the database is volatile and lost on every restart. Yet the module defines `SURREAL_PATH = stateDir`, exports it, and runs `mkdir -p "$SURREAL_PATH"` at startup — so that state dir and the mkdir are dead code, and the persistence the options imply never happens. Did you mean `rocksdb:$SURREAL_PATH` (or `file:$SURREAL_PATH`)?

src/modules/services/surrealdb.nix
```diff
@@ -0,0 +1,103 @@
...
+      SURREAL_ENDPOINT = "ws://localhost:${toString allocatedPort}";
+    };
+
+    processes.surrealdb = {
+      ports.main.allocate = basePort;
+      exec =
+        let
+          args =
+            let
+              mk = key: value: lib.optionalString (value != null) "--${key} ${value}";
```
domenkozar, line 76: `username`/`password`/`schema` are interpolated unquoted into the exec command. A value containing a space or shell metacharacter (e.g. `password = "p @ss"`) produces a malformed `surreal start --password p @ss …` and breaks startup. Quote the value, e.g. `lib.optionalString (value != null) "--${key} ${lib.escapeShellArg value}"`.

src/modules/services/surrealdb.nix
```diff
@@ -0,0 +1,103 @@
...
+    schema = lib.mkOption {
+      type = with lib.types; nullOr (either path str);
+      default = null;
+      description = "Path to a SurrealQL (.surql) file that will be imported when starting the server";
+    };
+
+    namespace = lib.mkOption {
+      type = with lib.types; nullOr str;
+      default = "main";
+      description = "Sets the defauilt namespace";
```
domenkozar, line 48: Typo: "defauilt" → "default". This surfaces in the auto-generated `docs/reference/options.md`.

PR #2960, opened 2026-06-26
feat: add services.spacetimedb

Felipe-9:
Add `services.spacetimedb` module that runs the server/database under devenv.

### DIFF
--- a/src/modules/services/spacetimedb.nix
+++ b/src/modules/services/spacetimedb.nix
@@ -1,0 +1,40 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  cfg = config.services.spacetimedb;
+
+  basePort = cfg.port;
+  allocatedPort = config.processes.spacetimedb.ports.main.value;
+
+  stateDir = cfg.env.DEVENV_STATE + "/spacetimedb";
+in
+{
+  options.services.spacetimedb = {
+    enable = lib.mkEnableOption "SpacetimeDB Process";
+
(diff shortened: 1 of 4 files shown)


src/modules/services/spacetimedb.nix
```diff
@@ -0,0 +1,40 @@
...
+  pkgs,
+  ...
+}:
+let
+  cfg = config.services.spacetimedb;
+
+  basePort = cfg.port;
+  allocatedPort = config.processes.spacetimedb.ports.main.value;
+
+  stateDir = cfg.env.DEVENV_STATE + "/spacetimedb";
```
domenkozar, line 13: `cfg.env.DEVENV_STATE` is undefined and makes enabling the service a hard eval error. `cfg = config.services.spacetimedb`, which only defines `enable`/`package`/`port` — there is no `env` attribute on it. Every other service module uses `config.env.DEVENV_STATE`. As written, `services.spacetimedb.enable = true` forces `stateDir` → `cfg.env` → `error: attribute 'env' missing`, so the service can never start.

```suggestion
  stateDir = config.env.DEVENV_STATE + "/spacetimedb";
```

tests/surrealdb/.test.sh
```diff
@@ -0,0 +1,7 @@
+set -e
```
domenkozar, line 1: This spacetimedb test is placed under `tests/surrealdb/`. The content runs `spacetime …` commands but the directory name says `surrealdb`, which both mislabels the test and directly collides with PR #2959 (which creates the same two paths). Please move to `tests/spacetimedb/`.

PR #2972, opened 2026-07-02
Fix POSIX hook sibling activation after cd-out

### LINKED ISSUE #2944: Auto-activation hook doesn't re-activate when cd-ing directly between two sibling devenv projects
**Describe the bug**

With the shell auto-activation hook (`eval "$(devenv hook zsh)"`), `cd`-ing *directly* from one trusted devenv
 project into a sibling trusted devenv project exits the first project's shell but never activates the second.
 You land in the destination directory with no devenv environment active (plain prompt, not `(devenv)`).

It only happens on a direct project→sibling-project hop. Leaving to a non-devenv directory works as documented,
 and entering the second project from a neutral directory works fine — so the destination is correctly trusted and detectable; it just gets skipped on the handoff.

Root cause (from the generated `hook-zsh.sh` / `posix.sh`): when you leave the project, the in-shell branch writes `$PWD` to `$DEVENV_ROOT/.devenv/exit-dir` and exits the subshell. The parent shell then reads that file and does:

```sh
cd "$target_dir"
_DEVENV_HOOK_PWD="$PWD"   # caches the destination
```

On the very next `precmd`, the hook's first guard short-circuits:

```sh
[[ "$_DEVENV_HOOK_PWD" == "$PWD" ]] && return $previous_exit_status
```

Because `_DEVENV_HOOK_PWD` was just set to the destination, `devenv hook-should-activate` is never called for
it, so a trusted devenv project at the destination is never entered. The suppression clears only after a subsequent `cd` to a different directory.

**To reproduce**

Two sibling directories, each a trusted devenv project with a `devenv.yaml` and `devenv allow` run. Minimal `devenv.nix` for each:

```nix
{ pkgs, ... }: {
  enterShell = ''echo "in $(basename $DEVENV_ROOT)"'';
}
```

```sh
# ~/.zshrc:  eval "$(devenv hook zsh)"
mkdir -p ~/projA ~/projB
# add devenv.yaml + devenv.nix to each, then:
(cd ~/projA && devenv allow)
(cd ~/projB && devenv allow)

cd ~/projA      # => (devenv) $   "in projA"   ✅
cd ~/projB      # projA's shell exits, lands in ~/projB
```

Expected: projA's shell exits and projB's shell activates — `(devenv) $ "in projB"`.

Actual: projA's shell exits and you land in `~/projB` with a plain (non-devenv) prompt. Re-activation only happens if you then `cd` elsewhere and back into `~/projB`.

**Suggested fix**

In the parent-shell exit-dir handoff, don't cache the destination as `_DEVENV_HOOK_PWD` after `cd "$target_dir"` (leave it empty), so the next `precmd` re-evaluates the destination and can activate it if it's a trusted project.

**Version**

```
devenv 2.1.2 (aarch64-darwin)
```
Auto-activation via `eval "$(devenv hook zsh)"` (zsh). Not using flakes.

Arshgill01:
Closes #2944

## Summary

- Stop the POSIX hook from marking the `.devenv/exit-dir` target as already processed after the parent shell follows a hook-spawned cd-out.
- Add a bash/zsh regression test that reproduces the sibling-project activation skip.
- Add an unreleased changelog entry for the user-visible shell hook fix.

## Root cause

When a hook-spawned shell cd'd out of project A into sibling project B, the parent POSIX hook followed `.devenv/exit-dir` into B and then set `_DEVENV_HOOK_PWD="$PWD"`. On the next prompt, the hook treated project B as already processed and skipped `devenv shell`, so B never activated.

## Validation

- [x] `cargo test -p devenv --test hook_outer_shell_survives posix_activates_sibling_after_cd_out -- --nocapture`
- [x] `cargo test -p devenv --test hook_outer_shell_survives -- --nocapture`
- [x] `cargo check -p devenv`
- [x] `bash -n devenv/hooks/hook.posix.sh`
- [x] `cargo fmt --check --package devenv`
- [x] `git diff --check`

### DIFF
--- a/devenv/tests/hook_outer_shell_survives.rs
+++ b/devenv/tests/hook_outer_shell_survives.rs
@@ -197,7 +197,8 @@ fn posix_activates_sibling_after_cd_out() {
         fs::create_dir_all(project_b.join(".devenv")).unwrap();
         let (_bin_dir, calls) = sibling_activation_shim(&project_a, &project_b);
         let script = format!(
-            "{src}\n\
+            "unset DEVENV_ROOT _DEVENV_HOOK_DIR\n\
+             {src}\n\
              {po}\n\
              cd {project_a:?}\n\
              _devenv_hook\n\
--- a/devenv/tests/hook_outer_shell_survives.rs
+++ b/devenv/tests/hook_outer_shell_survives.rs
@@ -4,6 +4,7 @@
 //! - `outer_shell_survives_cd_out` — #2805
 //! - `inner_shell_exits_on_cd_out` — hook-spawned shell must `exit` + write exit-dir
 //! - `no_respawn_inside_devenv_shell` — follow-up to #2815
+//! - `posix_activates_sibling_after_cd_out` — #2944
 
 use std::fs;
 use std::os::unix::fs::PermissionsExt;
@@ -63,6 +64,17 @@ fn shells() -> Vec<(&'static str, String, fn(&Path) -> String)> {
     .collect()
 }
 
+fn posix_shells() -> Vec<(&'static str, String)> {
+    let bin = devenv_bin();
+    [
+        ("bash", format!(r#"eval "$({bin} hook bash)""#)),
+        ("zsh", format!(r#"eval "$({bin} hook zsh)""#)),
+    ]
+    .into_iter()
+    .filter(|(s, _)| have(s))
+    .collect()
+}
+
 fn posix_path_override(dir: &Path) -> String {
     format!(r#"export PATH="{}:$PATH""#, dir.display())
 }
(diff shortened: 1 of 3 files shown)


devenv/tests/hook_outer_shell_survives.rs
```diff
@@ -139,6 +187,45 @@ fn no_respawn_inside_devenv_shell() {
...
+fn posix_activates_sibling_after_cd_out() {
+    for (shell, src) in posix_shells() {
+        let parent = tempfile::tempdir().unwrap();
+        let project_a = parent.path().join("project-a");
+        let project_b = parent.path().join("project-b");
+        fs::create_dir_all(project_a.join(".devenv")).unwrap();
+        fs::create_dir_all(project_b.join(".devenv")).unwrap();
+        let (_bin_dir, calls) = sibling_activation_shim(&project_a, &project_b);
+        let script = format!(
+            "{src}\n\
```
sandydoo: The test fails when run inside `devenv shell -- cargo test`. Need to clear the ambient vars first:

```suggestion
            "unset DEVENV_ROOT _DEVENV_HOOK_DIR\n\
            "{src}\n\
```

sandydoo (changes requested):
The test fails when ran inside `devenv shell -- cargo test ...`.

Arshgill01:
@sandydoo Thanks, fixed. The test now clears `DEVENV_ROOT` and `_DEVENV_HOOK_DIR` before sourcing the hook, so it also passes when run from inside `devenv shell -- cargo test ...`.

Verified locally:
- cargo test -p devenv --test hook_outer_shell_survives posix_activates_sibling_after_cd_out -- --nocapture
- cargo test -p devenv --test hook_outer_shell_survives -- --nocapture

PR #2976, opened 2026-07-03
Classify auto-activation traces

### LINKED ISSUE #2965: Classify auto-activation traces distinctly
## Problem

When tracing is enabled, shell auto-activation / `direnv-export` is operationally useful but semantically different from an explicit user command such as `devenv tasks run ...` or `devenv up`.

If auto-activation traces look the same as explicit command traces, service views and trace searches become harder to interpret.

## Desired capability

Keep auto-activation traces under the same `service.name=devenv`, but classify them distinctly with span names and attributes.

For example:

- `span.name=devenv.auto-activation` or equivalent
- `devenv.execution.kind=auto_activation`
- `devenv.trigger=direnv|hook`
- `devenv.command=direnv-export` or equivalent

The exact names are only suggestions; the important part is that auto-activation is clearly distinguishable from explicit user-initiated commands while remaining part of the `devenv` service.

## Why same service

Auto-activation is still `devenv` behavior, so keeping `service.name=devenv` preserves unified service views. Distinct span naming/attributes provide filtering and volume analysis without fragmenting the service identity.

## Acceptance sketch

- Auto-activation traces are identifiable in trace search and service views.
- Explicit user commands and automatic activation can be filtered separately.
- The classification works regardless of whether tracing was enabled explicitly or by a default policy.

<details>

| field | value |
| --- | --- |
| `agent_name` | 🧊 co3-fjord |
| `agent_session_id` | 5170dc85-4f34-4b97-929c-fa4929791e8d |
| `agent_tool` | Codex CLI |
| `agent_tool_version` | 0.141.0 |
| `agent_runtime` | Codex CLI 0.141.0 |
| `agent_model` | unknown |
| `runtime_profile` | /nix/store/gc2lri90hj19xwvpzk58j3hfvwpckc9r-coding-agent-runtime-profile/share/coding-agents/profile.json |
| `skills_manifest` | /nix/store/bnvsypa1klqvyk8zy6kzfm5y9jvwvhz3-agent-skills-corpus/share/agent-skills/manifest.json |
| `worktree` | dotfiles/schickling/2026-06-22-voice-inbox |
| `machine` | dev3 |
| `tooling_profile` | dotfiles@unknown-dirty |
</details>

Arshgill01:
Fixes #2965

## Summary

- Classify `direnv-export` trace roots as `devenv.auto-activation` with `devenv.execution.kind=auto_activation`, `devenv.trigger=direnv`, and `devenv.command=direnv-export`.
- Classify hook-spawned `devenv shell` activation as auto-activation with `devenv.trigger=hook` while keeping explicit CLI commands under the normal `devenv` span.
- Add an integration test that checks JSON trace output for direnv activation, native hook activation, and an explicit command.

## Root cause

`run_backend` always used the same root span name for backend command execution. That kept auto-activation under `service.name=devenv`, but did not provide a direct span name or attribute split between background activation paths and user-initiated commands.

## Validation

- [x] `/nix/var/nix/profiles/default/bin/nix build --accept-flake-config .#devenv`
- [x] `PATH="$PWD/result/bin:$PATH" devenv-run-tests run tests --only tracing-auto-activation`
- [x] `cargo fmt --check --package devenv`
- [x] `git diff --check HEAD~1..HEAD`

Attempted separately:

- `cargo check -p devenv --no-default-features` is still blocked in a plain shell by missing `nix-flake-c.pc` / `PKG_CONFIG_PATH`; the Nix-backed build above covers the full crate with the repository-provided dependencies.


domenkozar (approved):
I'll let @sandydoo take a look at this, to me it looks good.

Arshgill01:
Any update on this @sandydoo ?

sandydoo (changes requested):
Thanks for tackling this!

Lets focus on telling devenv _who_ the caller is instead of trying to statically classify commands.

For example, we could set/pass an env var in the shell hook scripts to tell devenv that its being called from a hook. Read it, unset it, and now you've got caller data without having to manually classify commands or risk misclassifying them (`_DEVENV_HOOK_DIR` and nested shells).

A few more issues:

- The span name should stay static. Use attributes instead.
- There's no need to call an event inside the span with the same data. Set it on the span.
- If the issue is that the JSON formatter isn't printing out the attributes you need, play with the format config for `json_subscriber`.
- If the span name is static, you can use the `instrument` macro to accomplish the above.
- The command -> str conversion should live in the CLI module.
- Needs a CHANGELOG.

Arshgill01:
Hey @sandydoo, thanks for the detailed review — the distinction between caller data and command classification makes sense.

I reworked this around a one-shot `_DEVENV_CALLER` marker that is set only on the exact invocation made by direnv or the native hooks. devenv reads and unsets it immediately, so `_DEVENV_HOOK_DIR` can keep doing its job for the lifetime of the shell without causing nested or explicit commands to be classified as hook calls.

I also went through the tracing points: the span is statically named `devenv` and created with `#[instrument]`, caller and command are span attributes, the duplicate event is gone, and the JSON layer now includes the current span. The command string conversion lives in the CLI module now, and I added the CHANGELOG entry.

There are regression checks for CLI, direnv, hook activation, the nested-shell case, and making sure the caller marker is not inherited by child commands. The full `devenv` test suite and the Nix build are passing.

Pushed in [`c410b7c5`](https://github.com/Arshgill01/devenv/commit/c410b7c54a8e5b9aeb69d798fc53c0ee4a49ce90). When you get a chance, could you take another look?

PR #2985, opened 2026-07-06
fix: fall back to SQLite DELETE journal mode if shmmap fails on WAL

### LINKED ISSUE #2947: Can't use devenv in directories without shared memory map support: (code: 5386) disk I/O error
**Describe the bug**

When running devenv in a directory that is mounted on a filesystem that doesn't support shared memory maps (I believe), such as a virtiofs guest mount inside a VM, it's not possible to evaluate the shell because the SQLite database at `.devenv/nix-eval-cache.db` can't be used in WAL mode [unless exclusive memory maps are used, apparently, though I haven't verified for myself].

```
$ devenv shell which cargo
  × Failed to initialize eval cache database: Database error: while executing migrations: error returned from database: (code: 5386) disk I/O error
```

**To reproduce**

Example: with incus. (Needs incus to have been initialised and your user to be part of the unprivileged incus group)

On NixOS:
```nix
  virtualisation.incus = {
    enable = true;

    socketActivation = true;
  };

  environment.systemPackages = [
    # Incus should use this for directory mounts when available
    pkgs.virtiofsd
  ];
```
(& add your user to `incus`. Run `incus admin init` as root after installation.)

```shell
#!/bin/sh
set -eu

VMNAME="demo"
PROJECT="user-$(id -u)"

# Where to mount the user's home subdir on your host system.
#
# If in an exotic location, may need
# `incus project set "$PROJECT" restricted.devices.disk.paths /path/to/location`
# as admin to allow you to use that weird location.
# Seems your own homedir doesn't need this.
MOUNTDIR="$HOME/vm/$VMNAME"

incus init images:alpine/3.23 "$VMNAME" \
  --vm \
  --project "$PROJECT" \
  --device root,size=100GiB \
  --config limits.cpu=2 \
  --config limits.memory=4GiB \
  --config security.secureboot=false

mkdir -p "$MOUNTDIR"
  
incus config device add "$VMNAME" home disk \
  "source=$MOUNTDIR" \
  path=/home/user/dir \
  type=disk \
  io.bus=virtiofs \
  --project "$PROJECT"

incus start --project "$PROJECT" "$VMNAME"

sleep 20

sh "$(dirname "$0")/setup_userspace.sh" "$VMNAME"

echo "Setting up $VMNAME"

TARGET_UID="$(id -u)"
TARGET_GID="$(id -g)"

incus exec "$VMNAME" --project "$PROJECT" --cwd /home/user -- ash -c "
  apk add e2fsprogs-extra cloud-utils-growpart &&
  growpart /dev/sda 2 &&
  resize2fs /dev/sda2 ;
  apk add --no-cache nano bash nix nix-openrc &&
  addgroup -g $TARGET_GID user ;
  adduser -D -u $TARGET_UID -G user -h /home/user -s /bin/bash user ;
  dd if=/dev/random bs=32 count=1 | passwd user --stdin ;
  passwd user --unlock ;
  rc-update add nix-daemon &&
  rc-service nix-daemon start &&
  nix-env --install --attr devenv -f https://github.com/NixOS/nixpkgs/tarball/nixos-26.05
"

```

Then:

```
$ incus exec demo -- ash -lc "cd /home/user/dir && mkdir -p A && cd A && devenv init && devenv shell echo Hi"
• Creating devenv.nix
• Creating devenv.yaml
• Creating .gitignore
  × Failed to initialize eval cache database: Database error: while executing migrations: error returned from database: (code: 5386) disk I/O error
```

**Version**

```
devenv 2.1.2 (x86_64-linux)
```

Overall, not sure if 'fixing' this is of interest to the project or not, but thought it deserved a place on the issue tracker for other searchers at the very least

alurm:
devenv shell hard-crashed with a SQLite disk I/O error on filesystems that can't provide WAL's required mmap(MAP_SHARED) for the -shm file (seen on virtiofs/9p VM mounts and similar). Detect SQLITE_IOERR_SHMMAP at both the initial connect and the migration step, and retry once in DELETE (rollback-journal) mode, which needs no shared memory at all.

Fixes https://github.com/cachix/devenv/issues/2947.

Assisted-by: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@
 - Fixed `devenv inputs add` from a subdirectory writing to a stray `devenv.yaml` in the subdir instead of the enclosing project. It now walks up to find `devenv.nix` the same way `devenv shell` does, so the input is added where the rest of devenv reads it.
 - Fixed `devenv gc` failing with "File devenv.nix does not exist" when run outside of a project. Garbage collection operates on the global devenv store and no longer requires a `devenv.nix` ([#2928](https://github.com/cachix/devenv/issues/2928)).
 - Fixed unfree package errors suggesting only generic Nix/NixOS configuration. They now point devenv users to `allow_unfree: true` or `nixpkgs.permitted_unfree_packages` in `devenv.yaml` ([#2850](https://github.com/cachix/devenv/issues/2850)).
+- Fixed `devenv shell` failing outright with a SQLite `disk I/O error` on filesystems that don't support shared-memory mmap (seen on virtiofs/9p VM mounts and similar). The eval cache and task cache databases now fall back to a plain rollback-journal mode on these filesystems instead of hard-crashing ([#2947](https://github.com/cachix/devenv/issues/2947)).
 
 ### Improvements
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1959,6 +1959,7 @@ name = "devenv-cache-core"
 version = "2.1.3"
 dependencies = [
  "blake3",
+ "libsqlite3-sys",
  "miette",
  "serde_json",
  "sqlx",
--- a/Cargo.nix
+++ b/Cargo.nix
@@ -6674,6 +6674,11 @@ rec {
             packageId = "blake3";
           }
           {
+            name = "libsqlite3-sys";
+            packageId = "libsqlite3-sys";
+            usesDefaultFeatures = false;
+          }
+          {
             name = "miette";
             packageId = "miette";
             features = [ "fancy" ];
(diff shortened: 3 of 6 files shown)


alurm:
Needs an explicit dep on libsqlite3-sys to detect SQLITE_IOERR_SHMMAP.

Reproduced and tested this fix manually (`devenv init && devenv shell echo hi`) via a VM with a folder shared via host via 9p (causes the same problem as in the issue), since this is a bit difficult to test automatically.

Migrator should trigger some SQLite statement execution which will cause fallback to be applied in case the underlying FS doesn't support shared mmap.

PR #2988, opened 2026-07-07
fix(reload): surface file watcher failures in the TUI instead of hanging

### LINKED ISSUE #2987: Devenv shell appears to be hanging when fs watches are exhausted
**Describe the bug**

`devenv shell` appears to be stuck, after debugging it I found that the issue was caused by an IntelliJ plugin bug causing it to exhaust watches (fs.inotify.max_user_watches)

This bug was already reported in #2892 but closed by the reporter, I think it would be nice to at least print a warning in the TUI if we hit this case

**To reproduce**

Setup:
```shell
# set max_user_watches to a low value in order to leak it more easily
sudo sysctl -w fs.inotify.max_user_watches=128
# leak the watches
mkdir -p /tmp/leak && for i in $(seq 1 200); do mkdir -p /tmp/leak/d$i; done && inotifywait -r -m /tmp/leak >/dev/null &
```

Run devenv shell
```shell
devenv shell
```
It should be stuck looking like this:

```shell
❯ devenv shell
✓ Validating lock 1 files                                                                                                                                27ms
✓ Configuring shell                                                                                                                                      22ms
  └ ✓ Evaluating shell cached                                                                                                                             9ms
✓ Loading tasks                                                                                                                                           3ms
  └ ✓ Evaluating devenv.config.task.config cached                                                                                                         0ms
✓ Running tasks                                                                                                                                         180ms
  └ ✓ devenv:enterShell                                                                                                                                  15ms
    └ ✓ devenv:python:virtualenv 2 lines → ~/dev/devenv                                                                                                  42ms
      └ ✓ devenv:compile-requirements skipped                                                                                                             0ms
    └ ✓ devenv:git-hooks:install 1 lines → prek installed at `.git/hooks/pre-commit` with specified config `/home/p/dev/devenv/.pre-commit-config.yaml`  28ms
      └ ✓ devenv:files 2 lines → Creating .devcontainer/devcontainer.json                                                                                63ms
        └ ✓ devenv:files:cleanup                                                                                                                         71ms
  └ ✓ devenv:enterTest skipped                                                                                                                            1ms

 7 of 7 tasks                                                                                                                                     ↑↓ navigate
```

The error message is shown via:
```shell
❯ devenv shell --verbose --no-tui
...
notify watch() error
fs worker for devenv-reload stopped: cannot send internal runtime error: channel closed
```

**Version**
`devenv 2.1.2 (x86_64-linux)`

Allsimon:
When the fs worker failed to register a watch (commonly after reaching the inotify watch limit, fs.inotify.max_user_watches), watchexec reported it as a non-fatal runtime error. But the error receiver was dropped right after FileWatcher::new() returned, so the send failed with "channel closed", which watchexec escalates to a fatal error. The worker exited, its failure was only logged via tracing (visible with --verbose --no-tui), and pending watch() calls waited on fs_ready forever — so devenv shell appeared to hang under the TUI.

Keep the error receiver alive with a drain task so a failed watch stays non-fatal and the worker keeps running, and surface the failure to the user via the activity system so it shows in the TUI. The first error is reported once (to avoid flooding when many watches fail at once) with a hint about the inotify limit; the rest go to tracing.

fixes #2987

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug Fixes
 
+- Fixed `devenv shell` appearing to hang under the TUI when the file watcher failed to register a watch — most commonly after reaching the inotify watch limit (`fs.inotify.max_user_watches`). The watcher no longer dies on the first failed watch, and now surfaces a warning in the TUI instead of only logging it under `--verbose --no-tui`.
 - Fixed a task whose `status` command exits nonzero being rendered as a failed `check status` activity, making successful runs look like they contained a failure. A nonzero status is a normal cache miss that runs the task's command; only a failure to spawn the status command itself is now reported as a failure ([#2984](https://github.com/cachix/devenv/issues/2984)).
 - Fixed `devenv processes restart` leaving a restarted process reported as "exited" by `devenv processes list`/`status` even though it was running. When the process had previously exited (restart policy said stop), the restart spawned a fresh job and supervisor but never published a new status, and processes without readiness probes produce no further status events until they exit again. Restarting such a process now also replaces a supervisor parked after exit (kept alive only for file watching), which previously stopped monitoring the restarted job ([#2982](https://github.com/cachix/devenv/issues/2982)).
 - Fixed devenv overriding user-configured experimental features in `nix.conf` ([#2931](https://github.com/cachix/devenv/issues/2931)).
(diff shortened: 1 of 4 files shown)


Allsimon:
It shows this warning:
image

Arshgill01:
@domenkozar I opened a small follow-up PR for this merge: https://github.com/cachix/devenv/pull/2990

The merge commit's main-branch `Release` and `Generate docs/examples` workflows are failing because `devenv-event-sources` now uses `devenv_activity`, but the generated `Cargo.nix` entry was not updated to pass `--extern devenv_activity`.

I reproduced it from a clean `origin/main` worktree with `nix build .#devenv --print-out-paths`; adding the missing generated dependency lets that build pass.

PR #2991, opened 2026-07-08
Add default trace destination

### LINKED ISSUE #2963: Add default trace destination with explicit precedence semantics
## Problem

`devenv` supports native tracing via `--trace-to` and `DEVENV_TRACE_TO`, but there is no separate default trace destination that applies only when the user has not explicitly configured tracing for a command.

For managed environments and CI, this means callers either need to remember `--trace-to` on every invocation or wrap the `devenv` CLI externally.

## Desired capability

Add a first-class default trace destination mechanism, for example:

- `DEVENV_TRACE_DEFAULT_TO=<trace-to>`; or
- an equivalent config-level setting in `devenv.yaml` / `devenv.nix`.

The default should apply only when no explicit tracing config is present.

## Precedence

Suggested precedence:

1. CLI `--trace-to ...` wins.
2. `DEVENV_TRACE_TO=...` wins.
3. Explicit disable, e.g. `DEVENV_TRACE_DISABLE=1`, disables default tracing.
4. Default destination applies only after the above are absent.

## Why native support

`devenv` already owns trace initialization and child context propagation. A native default avoids shell wrappers, PATH-order assumptions, and duplicated parsing around `--trace-to`.

## Acceptance sketch

- A user/admin can configure a default trace destination once.
- Explicit `--trace-to` and `DEVENV_TRACE_TO` are never overridden.
- There is a documented way to disable default tracing for a command/session.
- Existing opt-in tracing behavior remains compatible.

<details>

| field | value |
| --- | --- |
| `agent_name` | 🧊 co3-fjord |
| `agent_session_id` | 5170dc85-4f34-4b97-929c-fa4929791e8d |
| `agent_tool` | Codex CLI |
| `agent_tool_version` | 0.141.0 |
| `agent_runtime` | Codex CLI 0.141.0 |
| `agent_model` | unknown |
| `runtime_profile` | /nix/store/gc2lri90hj19xwvpzk58j3hfvwpckc9r-coding-agent-runtime-profile/share/coding-agents/profile.json |
| `skills_manifest` | /nix/store/bnvsypa1klqvyk8zy6kzfm5y9jvwvhz3-agent-skills-corpus/share/agent-skills/manifest.json |
| `worktree` | dotfiles/schickling/2026-06-22-voice-inbox |
| `machine` | dev3 |
| `tooling_profile` | dotfiles@unknown-dirty |
</details>

Arshgill01:
## Summary

- Add `DEVENV_TRACE_DEFAULT_TO` for default trace destinations when no explicit tracing is configured.
- Keep `DEVENV_TRACE_TO`, `--trace-to`, and legacy `--trace-output` precedence ahead of the default.
- Add `DEVENV_TRACE_DISABLE=1` to suppress the default for a command or session.

## Root cause

Tracing could only be enabled explicitly through CLI flags or `DEVENV_TRACE_TO`, which made wrapper-managed defaults override user intent unless every invocation reimplemented precedence. The trace resolver already centralizes explicit sources, so it now applies an environment default only after those sources are absent.

## Validation

- `cargo fmt --check --package devenv`
- `git diff --check`
- `nix flake check --accept-flake-config`
- `nix build .#devenv --print-out-paths --accept-flake-config`

Fixes #2963

### DIFF
--- a/devenv/src/cli.rs
+++ b/devenv/src/cli.rs
@@ -139,6 +139,8 @@ pub enum ParseTraceOutputError {
 pub enum TracingArgsError {
     #[error("DEVENV_TRACE_TO: {0}")]
     EnvParse(#[source] ParseTraceOutputError),
+    #[error("DEVENV_TRACE_DEFAULT_TO: {0}")]
+    DefaultEnvParse(#[source] ParseTraceOutputError),
     #[error("duplicate trace destination '{spec}' (would interleave output)")]
     DuplicateDestination { spec: TraceOutputSpec },
 }
@@ -477,7 +479,9 @@ pub struct TracingCliArgs {
             Formats: json (default), pretty, full, otlp-grpc, otlp-http-protobuf, otlp-http-json\n\n\
             When format is omitted, defaults to json.\n\
             Tracing is disabled by default.\n\n\
-            [env: DEVENV_TRACE_TO=] (comma-separated)"
+            [env: DEVENV_TRACE_TO=] (comma-separated)\n\
+            [env: DEVENV_TRACE_DEFAULT_TO=] (comma-separated, used only when no explicit tracing is configured)\n\
+            [env: DEVENV_TRACE_DISABLE=1] (disables DEVENV_TRACE_DEFAULT_TO)"
     )]
     pub trace_to: Vec<TraceOutputSpec>,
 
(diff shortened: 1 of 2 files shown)


devenv/src/cli.rs
```diff
@@ -570,6 +588,10 @@ impl TracingCliArgs {
             specs.push(TraceOutputSpec::Render(self.trace_format, sink.clone()));
         }
 
+        if specs.is_empty() && !Self::default_tracing_disabled() {
+            specs = Self::default_specs_from_env()?;
+        }
```
sandydoo, line 587: `DEVENV_TRACING_DISABLE` sounds to me like it disables tracing completely, but instead appears to skip setting `DEVENV_TRACE_DEFAULT_TO`.

Besides the name being misleading, I'd remove `DEVENV_TRACING_DISABLE` entirely. The default env var can either be unset or set to an empty string as needed.

devenv/src/cli.rs
```diff
@@ -1095,6 +1117,49 @@ pub enum InputsCommand {
...
+            let saved = vars.iter().map(|&var| (var, env::var(var).ok())).collect();
+            for var in vars {
+                unsafe { env::remove_var(var) };
+            }
+            Self { _lock: lock, saved }
+        }
+
+        fn set(&self, name: &str, value: &str) {
+            unsafe { env::set_var(name, value) };
+        }
```
sandydoo: Any vars set here don't get restored on drop.

sandydoo (changes requested):

sandydoo (approved):

PR #2992, opened 2026-07-08
Use login shell fallback for shell selection

### LINKED ISSUE #2880: zsh integration not getting user shell on VSCodium
**Describe the bug**
Even though my projects have no reference to `bash` and my nix user has `zsh` as shell defined for terminals the shell auto-activated has `bash` instead of `zsh`.

**To reproduce**

When auto activation is configured with `eval "$(devenv hook zsh)"` on `programs.zsh.initContent`.

A devenv.nix where this occur (but also happening in all the projects I configured:

```nix

{
  devcontainer.enable = true;

  # https://devenv.sh/basics/
  env.GREET = "devenv";

  # https://devenv.sh/packages/
  packages = [
    pkgs.mdbook
    pkgs.mdbook-linkcheck2
    pkgs.mdbook-mermaid
  ];

  languages = {
    rust.enable = true;
  };

  enterTest = ''
    echo "Running tests"
    prek run --all-files
    mdbook test
  '';

  git-hooks.hooks = {
    check-added-large-files.enable = false;
    nixpkgs-fmt = {
      enable = true;
      args = [ "--check" ];
    };
    check-yaml.enable = true;
    end-of-file-fixer.enable = true;
    shellcheck.enable = true;
    trim-trailing-whitespace.enable = true;
    yamllint.enable = true;
  };
}
```

**Version**
devenv 2.1.2 (x86_64-linux)

Arshgill01:
## Summary

- Resolve the interactive shell from the user's login shell when `SHELL` is missing or unsupported.
- Keep explicit `--shell`, `DEVENV_SHELL_TYPE`, and config shell values ahead of ambient shell detection.
- Add shell-selection tests for env precedence, login-shell fallback, and unsupported shell handling.

## Root cause

Editor-launched terminals can strip or replace `SHELL`, which made auto-activation fall back to Bash even for users whose login shell is zsh. Shell resolution now treats `SHELL` as an ambient hint and falls back to the login shell before using the default.

## Validation

- `cargo test -p devenv-core -- --nocapture`
- `cargo fmt --check --package devenv-core`
- `git diff --check`
- `nix build .#devenv --print-out-paths --accept-flake-config`
- `nix flake check --accept-flake-config`

Fixes #2880


devenv-core/src/settings.rs
```diff
@@ -375,6 +429,44 @@ pub struct InputOverrides {
 mod tests {
     use super::*;
     use crate::config::Config;
+    use std::sync::{Mutex, MutexGuard};
+
+    static ENV_LOCK: Mutex<()> = Mutex::new(());
+
+    struct EnvVarGuard {
```
domenkozar: I'm not sure this is a good idea, mostly because these tests don't run in sandbox so it's not guaranteed the value of $SHELL

devenv-core/src/settings.rs
```diff
@@ -320,6 +321,59 @@ impl ShellSettings {
...
+    login_shell_path_from_passwd(&username, &passwd)
+}
+
+#[cfg(not(unix))]
+fn login_shell_path() -> Option<String> {
+    None
+}
+
+#[cfg(unix)]
+fn login_shell_path_from_passwd(username: &str, passwd: &str) -> Option<String> {
```
domenkozar: I'm sure there's a crate for passwd parsing

Arshgill01:
Hey @domenkozar, thanks for the review — both points were spot on.

**On the tests:** You're right. The `EnvVarGuard` approach was fighting the wrong problem. The injectable fn pattern was already half there with `resolve_with_login_shell`, so I extended it to cover the `$SHELL` env source too — renamed to `resolve_with_env_and_login_shell` with a second `env_shell` fn param. The tests are now fully pure: no `unsafe`, no mutex, no live env mutation at all.

**On the passwd parsing:** Agreed. I've replaced the hand-rolled `/etc/passwd` reader with `nix::unistd::User::from_name()`, which uses `getpwnam_r` and properly goes through NSS — so it works with LDAP, systemd-homed, etc. The `nix` crate was already a workspace dependency, just needed the `user` feature added. The custom `login_shell_path_from_passwd` function and its test are now gone.

Pushed in [`de7893c`](https://github.com/Arshgill01/devenv/commit/de7893c3). Let me know if anything else looks off.

devenv-core/src/settings.rs
```diff
@@ -320,6 +321,46 @@ impl ShellSettings {
...
+
+#[cfg(unix)]
+fn login_shell_path() -> Option<String> {
+    let username = std::env::var("USER")
+        .ok()
+        .or_else(|| std::env::var("LOGNAME").ok())?;
+    nix::unistd::User::from_name(&username)
+        .ok()
+        .flatten()
+        .and_then(|u| u.shell.to_str().map(String::from))
```
domenkozar: The motivation for this PR is that editor-launched terminals strip environment variables, but this lookup depends on `USER`/`LOGNAME` being present and correct. A terminal that strips `SHELL` can just as easily strip these, and they are also stale under `su`/`sudo` without a login shell. Resolving by uid needs no env at all:

```suggestion
    nix::unistd::User::from_uid(nix::unistd::getuid())
        .ok()
        .flatten()
        .and_then(|u| u.shell.to_str().map(String::from))
```

devenv-core/src/settings.rs
```diff
@@ -320,6 +321,46 @@ impl ShellSettings {
...
+}
+
+fn shell_name_from_path(shell: &str) -> Option<String> {
+    std::path::Path::new(shell)
+        .file_name()?
+        .to_str()
+        .map(String::from)
+}
+
+fn login_shell_name() -> Option<String> {
```
domenkozar: This reduces the passwd shell (e.g. `/run/current-system/sw/bin/zsh` on NixOS) to a basename, and `resolve_shell_path` in `devenv/src/devenv/mod.rs` later re-resolves it via `$SHELL` (absent in this scenario by definition) or `which` on `PATH`. In exactly the stripped-environment case this PR targets, `PATH` may not contain the login shell's directory, so devenv would exec a bare `zsh` and fail where it previously fell back to bash and worked. Plumbing the absolute path through `ShellSettings` would fix it properly; if that is too invasive here, a comment or follow-up issue would be good.

devenv-core/src/settings.rs
```diff
@@ -279,36 +288,28 @@ impl ShellSettings {
...
-                        .map(String::from)
-                })
-            })
-            .unwrap_or_else(|| "bash".to_string());
-
-        // Only bash, zsh, fish, and nu are supported; fall back to bash for anything else
-        let shell = match shell.as_str() {
-            "bash" | "zsh" | "fish" | "nu" => shell,
-            _ => "bash".to_string(),
+        let (shell, source) = if let Some(shell) = options.shell.to_owned() {
```
domenkozar: When `config.shell` (or `options.shell` from a non-CLI caller) holds an unsupported value, the debug log reports `shell=bash (source: devenv.yaml)`, i.e. the fallback value with the explicit source label. This quirk predates the PR, but since the branch structure is being rewritten anyway it is cheap to fix, and a `warn!` when an explicitly requested shell gets silently downgraded to bash would be even better.

devenv-core/src/settings.rs
```diff
@@ -653,6 +694,54 @@ mod tests {
...
+
+        assert_eq!(settings.shell, "bash");
+    }
+
+    #[test]
+    fn shell_settings_unsupported_env_shell_uses_login_shell() {
+        let settings = ShellSettings::resolve_with_env_and_login_shell(
+            ShellOptions::default(),
+            &Config::default(),
+            || Some("/bin/sh".to_string()),
```
domenkozar: The real `shell_name_from_env` always returns a basename (the comment in the test above says so), so this injected value breaks the closure's contract. The test only passes because a full path is also unsupported. Using the basename exercises the actual scenario, an unsupported shell name:

```suggestion
            || Some("sh".to_string()),
```

devenv-core/Cargo.toml
```diff
@@ -10,6 +10,7 @@ snix = []
 [dependencies]
 devenv-activity.workspace = true
 nix-conf-parser.workspace = true
+nix.workspace = true
```
domenkozar: Two things here:

1. The `#[cfg(not(unix))]` stub in settings.rs suggests non-unix builds should work, but this unconditional dependency would not build there. The workspace convention already handles this: `devenv-tasks` and `devenv-shell` declare `nix` under `[target.'cfg(unix)'.dependencies]`.
2. Rather than adding `user` to the workspace-level features in the root `Cargo.toml` (which switches it on for every consumer: tokio-shutdown, devenv-processes, devenv), request it locally like devenv-tasks does.

```toml
[target.'cfg(unix)'.dependencies]
nix = { workspace = true, features = ["user"] }
```

domenkozar (commented):
Reviewed. The approach is sound: explicit sources keep priority, and the `FnOnce` closures make the passwd lookup lazy, so the common path pays nothing. Nice testability refactor and the CHANGELOG entry follows house style.

Inline comments below. The main ask is resolving the login shell by uid instead of `USER`/`LOGNAME` (the stripped-env scenario this PR fixes can strip those too). The Cargo.toml scoping and the test input are quick fixes; the rest is optional polish.

One small test gap, no inline anchor for it: nothing pins that an explicit `options.shell`/`config.shell` suppresses the login shell lookup entirely. A test passing a panicking closure as `login_shell` would lock that in cheaply.

Arshgill01:
Hey @domenkozar, good catches all around. Went through all five and took the open-ended one a bit further too.

**`from_uid(getuid())`** — yeah that's a much better call, hadn't thought about `USER`/`LOGNAME` having the same stripping problem. Switched over, no more env dependency at all for the lookup.

**nix dep scoping** — moved it to `[target.'cfg(unix)'.dependencies]` in `devenv-core/Cargo.toml` with the `user` feature requested locally (same pattern as `devenv-tasks`), and reverted `user` from the workspace root.

**The path stripping issue** — decided to fix this properly rather than just leave a note. Added `shell_path: Option<PathBuf>` to `ShellSettings` so the absolute path from `getpwuid` gets carried through. `mod.rs` now uses it directly when it's set, skipping `resolve_shell_path()` altogether — which was the broken part (`$SHELL` gone, `which` fails on a stripped PATH, falls back to a bare name, exec fails). Config and env sources still go through `resolve_shell_path` as before.

**warn on unsupported shell** — added a `warn!` when CLI `--shell` or `devenv.yaml` has an unsupported value and we silently fall back to bash. Also fixed the source label in the debug line so it actually reflects what happened.

**`/bin/sh` → `sh`** — fixed. Also threw in assertions on `shell_path` in the login-shell tests while I was at it.

Pushed in [`b5755e34`](https://github.com/Arshgill01/devenv/commit/b5755e34).

devenv/src/devenv/mod.rs
```diff
@@ -786,7 +786,16 @@ impl Devenv {
...
-            Some(resolve_shell_path(dialect.name()))
+            // Prefer the absolute path from the login-shell database (populated when the shell
+            // was resolved via getpwuid). In stripped environments $SHELL and PATH may not
+            // contain the shell's Nix store path, so resolve_shell_path would fail.
+            let path = self
+                .shell_settings
+                .shell_path
+                .as_ref()
+                .map(|p| p.to_string_lossy().to_string())
+                .unwrap_or_else(|| resolve_shell_path(dialect.name()));
```
domenkozar, line 798: This covers `prepare_shell`, but the primary interactive flow does not go through here: for `use_pty` + `Commands::Shell`, `run_backend` in `devenv/src/main.rs` hands only `shell_settings.shell.clone()` (the name) to `DevenvShellBuilder`, and `build_interactive` in `devenv/src/reload/mod.rs` still calls `resolve_shell_path(dialect.name())` unconditionally. So in the exact scenario this PR fixes (stripped env, interactive `devenv shell`), the login shell's absolute path is dropped on the main path.

Suggested fix: add `shell_path: Option<PathBuf>` to `DevenvShellBuilder`, pass `devenv.shell_settings.shell_path.clone()` through `ReloadShellArgs` in `main.rs`, and apply the same prefer-`shell_path` logic in `build_interactive`.

Optional robustness tweak for both call sites: `.filter(|p| p.exists())` before trusting the passwd path, so a stale passwd entry still falls back to `resolve_shell_path`.

domenkozar (commented):
Thanks, verified the follow-up commits locally: `cargo check -p devenv` is clean and all 18 `shell_settings` tests pass. The uid-based lookup, the Cargo.toml scoping under `[target.'cfg(unix)'.dependencies]`, the basename test input, and the warn-on-downgrade with accurate source labels all look good.

One gap remains in the `shell_path` plumbing, inline below.

domenkozar:
Went through b5755e34 and verified it locally: `cargo check -p devenv` is clean and all 18 `shell_settings` tests pass. The uid-based lookup, the dep scoping, the warn on fallback with accurate source labels, and the test fixes all look good. Taking the path issue on properly with `shell_path` instead of a note was the right call.

One thing left before merge: the `shell_path` plumbing only reaches `prepare_shell`. The primary interactive flow (`use_pty` + `Commands::Shell` in `main.rs` goes through `run_reload_shell` into `DevenvShellBuilder::build_interactive` in `devenv/src/reload/mod.rs`) still receives only the shell name and calls `resolve_shell_path(dialect.name())` unconditionally. So in the exact scenario this PR fixes, the login shell's absolute path is dropped on the main path. Details in the inline comment on my last review: carry `shell_path: Option<PathBuf>` through `ReloadShellArgs` and `DevenvShellBuilder`, and apply the same prefer `shell_path` logic in `build_interactive`. A `.filter(|p| p.exists())` guard at both call sites would also protect against stale passwd entries.

Arshgill01:
done, also fixed the PTY path (`build_interactive` had the same issue) and updated the CHANGELOG — [`ec4b9c2c`](https://github.com/Arshgill01/devenv/commit/ec4b9c2c)

domenkozar:
Verified ec4b9c2c locally: `cargo fmt --check` is clean, all 18 `shell_settings` tests pass, and `cargo check -p devenv` succeeds. The plumbing through `ReloadShellArgs` and `DevenvShellBuilder` covers the PTY path now, the `.exists()` guard is applied at both call sites, and the Cargo.lock update from the previous commit is included. This looks complete to me.

domenkozar:
> done, also fixed the PTY path (`build_interactive` had the same issue) and updated the CHANGELOG — [`ec4b9c2c`](https://github.com/Arshgill01/devenv/commit/ec4b9c2c)

build seems to fail

Arshgill01:
> > done, also fixed the PTY path (`build_interactive` had the same issue) and updated the CHANGELOG — [`ec4b9c2c`](https://github.com/Arshgill01/devenv/commit/ec4b9c2c)
> 
> build seems to fail

on it.

Arshgill01:
@domenkozar yeah, missed regenerating `Cargo.nix` after adding the target-scoped `nix` dependency. Cargo was fine locally, but the crate2nix build never passed `nix` to `devenv-core`, which is why all three platform builds and docs failed. Regenerated it and verified the exact x86_64 CI build plus fmt and all 18 shell settings tests — pushed in [`e7a523be`](https://github.com/Arshgill01/devenv/commit/e7a523be). The new runs just need fork workflow approval.

Issue #2801, opened 2026-05-08
zoxide: infinite loop detected


pbek:
**Describe the bug**

The command `cd` using [Zoxide](https://github.com/ajeetdsouza/zoxide) with `cd` integration in a devenv shell activated by `devenv allow` is failing with:

```
zoxide: infinite loop detected
Avoid aliasing `cd` to `z` directly, use `zoxide init --cmd=cd fish` instead
```

You cannot use `cd` anymore in that project.
Note: I'm using fish shell with the devenv fish hook.
With direnv (`direnv allow`) I don't have this issue, just with devenv (`devenv allow`).

home-manager config:

```nix
    programs = {
      fish = {
        enable = true;
        shellInit = ''
          devenv hook fish | source
        '';
      };

      # A smarter cd command
      # https://github.com/ajeetdsouza/zoxide
      zoxide = {
        enable = true;
        enableFishIntegration = true;
        options = [ "--cmd cd" ];
      };
    };
```

**To reproduce**

- Enable zoxide in home-manager with `cd` integration
- Run `devenv allow` in a directory with devenv config (devenv v2.1 needed)
- Try using `cd`
- See error `zoxide: infinite loop detected` instead of changing directory

**Version**

devenv 2.1.0 (x86_64-linux)

pbek:
With an overlay of the current main branch of devenv `devenv allow` didn't load the devenv env for me anymore. I can't access the packages of the env, I cannot see the "watching" on the bottom. 🤔

See, https://github.com/cachix/devenv/issues/2803#issuecomment-4444490320

pbek:
Sorry, "I was holding it wrong". I can confirm that as of commit https://github.com/cachix/devenv/commit/c733274dc2900f4bf8b3de279de8c5577930d982 `devenv allow` is working and has no zoxide infinite loop! Thank you very much!

pbek:
@sandydoo, I was able to reproduce the zoxide infinite loop again with devenv from 2 days ago from my overlay.

- `cd` into project directory with devenv config, but where there is no `devenv allow` yet
- `direnv allow` config is built and "devenv watching" is active
- `cd ..` still works
- `cd` back to project, see that "devenv watching" is active
- `cd ..` get `zoxide: infinite loop detected`

adiSuper94:
I am on devenv 2.1.2 (x86_64-linux) + fish + zoxide.
I have ` devenv hook fish | source` in in my fish config.

Steps to repro
[...]

PR #3000, opened 2026-07-12
hook: fix activation/deactivation edge cases across bash, zsh, fish, nushell

### LINKED ISSUE #2853: `cd -` doesn't get you back to the project directory
**Describe the bug**

When you leave a project directory where you have `devenv allow` enabled and try to get back to it with `cd -` you will get to another previous directory.

I'm not sure if this is a problem in combination with fish and zoxide, but I paste my config.

home-manager config:

```nix
    programs = {
      fish = {
        enable = true;
        shellInit = ''
          devenv hook fish | source
        '';
      };

      # A smarter cd command
      # https://github.com/ajeetdsouza/zoxide
      zoxide = {
        enable = true;
        enableFishIntegration = true;
        options = [ "--cmd cd" ];
      };
    };
```

**To reproduce**

- `cd your-project`
- `devenv allow` (see that devenv env is built)
- `cd ..` get out of project directory
- `cd -` expect to get back into project directory
- See that you are in a different directory or the directory didn't change at all

**Version**

devenv 2.1.3 (x86_64-linux) from main branch yesterday (8eff3cd84a4c2a86a02fe706582ae348650e3e76)

### LINKED ISSUE #2861: Changing directory in zellij/tmux in project dir with `devenv allow` enabled closes session in terminal multiplexer
**Describe the bug**

Changing directory in terminal multiplexer like zellij or tmux in a project dir with `devenv allow` enabled closes the session (tab or whole session) in terminal multiplexer.

This issue is a bit like https://github.com/cachix/devenv/issues/2803, but in a terminal multiplexer.

home-manager config:

```nix
    programs = {
      fish = {
        enable = true;
        shellInit = ''
          devenv hook fish | source
        '';
      };
    };
```

**To reproduce**

- `mkdir devenv-test && cd devenv-test`
- `devenv init`
- `devenv allow` see that env is built (currently not working for me, because of #2842) and devenv is "watching"
- `zellij` run zellij or tmux
- `cd ..` try to get out of directory in zellij session
- See that zellij session closes

**Version**

devenv 2.1.3 (x86_64-linux)

alurm:
Four related fixes found while testing the shell hook across all
four dialects:

- `_DEVENV_HOOK_DIR` leaking into descendant shells (new tmux/zellij panes, manually started nested shells) and wrongly exiting them on cd-out. Fixes #2861.
- Cd-out detection depended on the spawned shell's own rc file re-sourcing the hook script a second time, which some configs skip (e.g. fish config gated behind `status is-login`). Detection now lives directly in each dialect's own generated init file.
- Fish spawning a redundant nested devenv shell when something else (a manually entered `devenv shell`, direnv's `use devenv`, etc.) activated the same directory during the one-prompt delay between fish deciding to activate and actually spawning.
- `cd -` didn't return to the project directory once you left it. When you `cd` out, the outer shell that takes back over needs to catch up to wherever you actually ended up — it does this with `builtin cd`, not `cd`, a workaround from an earlier zoxide fix (#2801) that avoids retriggering a user-overridden `cd` (e.g. zoxide's wrapper). But `builtin cd` skips fish's own directory-history bookkeeping (`$dirprev`), which plain `cd` updates. That catch-up `cd` now replicates the bookkeeping directly. Fixes #2853.

Assisted by Claude Sonnet 5.


devenv-shell/src/dialect/bash.rs
```diff
@@ -48,6 +48,27 @@ __devenv_compute_diff "$_devenv_before_file"
...
+    __devenv_exit_on_cd_out() {{
+        case "$PWD" in
+            "$DEVENV_ROOT"|"$DEVENV_ROOT"/*) ;;
+            *)
+                printf '%s' "$PWD" > "$DEVENV_ROOT/.devenv/exit-dir"
+                exit
+                ;;
+        esac
+    }}
+    PROMPT_COMMAND="__devenv_exit_on_cd_out${{PROMPT_COMMAND:+;$PROMPT_COMMAND}}"
```
sandydoo: This should be set _after_ sourcing bashrc to be on the safe side

sandydoo (changes requested):
Otherwise, LGTM! Tests pass.

devenv-shell/src/dialect/bash.rs
alurm: Done, now it's set after sourcing bashrc.

alurm:
Done. `_DEVENV_HOOK_DIR` is still consumed immediately so it cannot leak to descendants, but the cd-out callback is now registered after sourcing user configuration. I applied this ordering change to zsh and nu as well (fish was already correct). I also strengthened regression tests to verify that both (1) devenv’s hook remains registered if the user replaces the hook configuration and (2) a user's dummy callback is preserved by us. Rebased the PR onto current main.

PR #3008, opened 2026-07-14
Documentation improvement attempt. Change nav bar organization, new page about pinning and add examples from github. 

chrissound:
I've previously tried to look into devenv but was a bit intimidated or confused by the existing docs. This is my attempt at maybe improving it. I've also added an additional pinning page - as this is one of the most useful features I looked for initially and didn't easily find.

Full disclosure I did create the pinning document with AI/LLM, but I have read through it line by line a couple of times and it makes sense to be (maybe the wording is a bit clunky though). 

Mostly it is just re-arranging things under the nav bar: 

Minor hiccup -  the tutorial link has been moved to the top, which links to https://github.com/cachix/nixcon-2024-workshop which maybe is outdated now?


docs/src/pinning.md
```diff
@@ -0,0 +1,78 @@
...
+`devenv info` also shows locked inputs.
+
+## Pinning to a specific revision
+
+Use `?rev=` in the input URL for an exact commit:
+
+```yaml title="devenv.yaml"
+inputs:
+  nixpkgs-stable:
+    url: github:NixOS/nixpkgs?rev=ac62194c3917d5f474c1a844b6fd6da2db95077d
```
domenkozar:
```suggestion
    url: github:NixOS/nixpkgs/ac62194c3917d5f474c1a844b6fd6da2db95077d
```

domenkozar:
Looks good to me! cc @sandydoo 

We'll want to also update `docs/_redirects` for old links to work.

docs/src/pinning.md
chrissound: There is also a similar reference here if it also has to change? https://github.com/cachix/devenv/blob/bd1c175d8aff2cd47e1999f6be7b7d79a4253d93/docs/src/inputs.md?plain=1#L82

domenkozar: Yes, that should change too.

chrissound:
Will fix those references when I'm back on my pc.  About the redirects though is that for me? I don't think any links change, I didn't modify the filenames?

sandydoo:
I pushed the two redirects to main, both minor.

chrissound:
Just curious if those links with the `#id` are valid? Previously I tried right clicking copy on them in the nav and didn't spot them anywhere? 

https://devenv.sh/creating-files/ is also doing an infinite redirect it seems right now?

sandydoo:
Yeah, and cloudflare also doesn't support fragments in the source of the redirects. I've cleaned up the broken redirects we had. Thanks!

Issue #2915, opened 2026-06-07
High CPU usage when idle in `devenv up` foreground mode


polaris64:
## Summary

`devenv up` consumes significant CPU while idle, even when all managed processes are idle and producing no output.

The issue disappears when running in detached mode (`devenv up -d`), which suggests the CPU usage originates from the foreground TUI rather than the managed processes themselves.

## Environment

* Devenv version: 2.1.2
* OS: Arch Linux (using Nix as a package manager)
* Terminal: reproduced in multiple terminals (including Ghostty and Alacritty)

## Expected behaviour

When all managed processes are idle, `devenv up` should consume minimal CPU.

## Actual behaviour

With idle managed processes, `devenv up` continuously consumes approximately 10-15% CPU per managed process.

The managed processes themselves (`python`, `watchexec`, etc.) remain idle.

Running the same environment with:

```bash
devenv up -d
```

reduces CPU usage to near zero.

### Process behaviour

The CPU is consumed by the foreground `devenv up` process itself, not by child processes.

### strace output

Attaching `strace` to the busy process shows a repeating pattern:

```text
epoll_wait(3, [], 1024, 0) = 0
openat(AT_FDCWD, "/dev/tty", O_RDONLY|O_CLOEXEC) = 40
ioctl(40, TIOCGWINSZ, {ws_row=57, ws_col=105, ...}) = 0
close(40) = 0
write(2, "\33[?2026h", 8) = 8
write(2, "\33[?2026l", 8) = 8
sched_yield() = 0
sched_yield() = 0
...
epoll_wait(3, [], 1024, 10) = 0
...
```

This pattern repeats continuously while the system is otherwise idle.

The process appears to be:

1. Querying terminal size.
2. Emitting terminal control sequences.
3. Yielding repeatedly.
4. Recomputing/redrawing without any apparent external events.

### perf output

Sampling the process with:

```bash
perf top -p <pid>
```

shows a large proportion of CPU time spent in:

```text
taffy::compute::flexbox::compute_flexbox_layout
```

For example:

```text
23.27%  .devenv-wrapped  [.] taffy::compute::flexbox::compute_flexbox_layout
```

This suggests the foreground UI is continuously recomputing layout despite no visible changes occurring.

polaris64:
I forgot to mention, running `devenv up --no-tui` reduces CPU usage significantly, so I'm pretty sure this is a TUI layout computation issue

domenkozar:
Could you try #2916

[...]

PR #3013, opened 2026-07-16
tui: stop spinning when the terminal disconnects

alurm:
## Summary

Stop the interactive TUI and backend when their terminal disconnects.

Without this, closing some terminal hosts can leave `devenv up` running and consuming 100%+ CPU indefinitely.

## Reproduction

Reproduced with a Zed terminal on macOS:

1. Exit the auto-activated `devenv shell`, if one is active.
2. Run `devenv up`.
3. Close the terminal pane.

Before this change, `devenv up` remained alive and consumed 100%+ CPU. After this change, it exits.

I couldn't reproduce this in Terminal.app, over a Zed Linux remote connection, or when `devenv up` ran inside `devenv shell`. In the last case `ShellSession` runs the nested command behind an inner PTY and #2871 handles teardown of that PTY on shutdown. A direct `devenv up` bypasses `ShellSession` and attaches its TUI directly to the host terminal.

## Root cause

Closing a PTY causes terminal reads to return EOF or an error such as `EIO`. Ordinary programs treat either result as terminal and exit.

The affected crossterm Unix event source instead continues its inner read loop when the read returns zero bytes or a non-retryable error. Because the hung-up descriptor remains readable, it immediately reads again and busy-loops. Iocraft does not turn this condition into a TUI exit.

Devenv handles lifecycle signals cooperatively rather than relying on their default process-terminating behavior. The TUI also intentionally remains active during ordinary shutdown so it can display backend cleanup. It therefore needs an explicit terminal-disconnection path that stops both the renderer and backend.

## Fix

While the interactive TUI is active, poll stdin for:

- `POLLHUP` when the PTY peer closes;
- `POLLERR` when the descriptor reports an error;
- `POLLNVAL` when the descriptor is no longer valid.

Polling checks descriptor state without reading from stdin.

On disconnect, the watcher sets the TUI's exit flag, notifies its redraw loop, and requests backend shutdown. Exiting the renderer stops the failed terminal-input loop, while backend shutdown cleans up managed processes.

The watcher requests `POLLIN` because Darwin only reported the reproduced PTY exit as `POLLIN | POLLHUP` when read readiness was requested. It still treats only `POLLHUP`, `POLLERR`, or `POLLNVAL` as disconnection, and does not consume keyboard input.

The watcher checks every 100 milliseconds. Normal TUI teardown stops, unparks, and joins it immediately.

## Testing

The manual reproduction above verifies the complete behavior before and after the change.

Also added a PTY-level regression test that:

1. writes input to the PTY;
2. confirms polling reports input readiness without consuming it;
3. confirms the input remains readable;
4. closes the PTY master;
5. confirms the slave reports terminal hangup.


domenkozar (commented):
Deep review of this PR. The approach is sound and the concurrency details check out: the park/unpark/join lifecycle in `TerminalDisconnectWatcher::drop` is race-free (an unpark before `park_timeout` returns immediately, so no missed wakeup and no blocking join), the set-flag-then-notify ordering is right, and the test's `contains`/`intersects` assertions hold on multi-bit revents. The findings below are ranked by severity. 13 are confirmed by code tracing or empirical tests, 2 are plausible.

## Correctness

1. **Transient `poll()` errors tear down healthy sessions** (`devenv-tui/src/app.rs:98`, confirmed). `Err(_) => PollFlags::POLLERR` maps every non-EINTR failure to a disconnect. A single transient `ENOMEM` (Linux) or `EAGAIN` (macOS, documented as retryable) kills the TUI and all managed processes with the terminal fully attached. A genuinely revoked fd is always reported via revents (`POLLHUP`/`POLLNVAL`), never via `Err`, so treating errors like EINTR (ignore, re-poll next tick) loses nothing.

2. **`eprintln!` panic in the target scenario** (`devenv-tui/src/app.rs:456` and below, confirmed). The disconnect path reliably drives `run()` into the final error-render block. Its bare `eprintln!` calls panic on write failure ("failed printing to stderr") once stderr is a revoked PTY, while every surrounding write correctly uses `let _ = ...`. Disconnect during a failing build (`has_errors` includes incomplete builds) exits 101, and on the PTY-shell path the panic can preempt backend teardown.

3. **Disconnect during a long Nix eval still lingers** (`devenv-tui/src/app.rs:391`, confirmed). `shutdown.shutdown()` only cancels the token; nothing calls `trigger_interrupt()` on that path (only reload/post-session paths do), and `run()` blocks on `event_processor_handle.await` until `backend_done`. So a disconnect mid-evaluation leaves the process alive until the eval completes, with the crossterm reader spinning the whole time (see next point).

4. **The 100% CPU spin itself is not stopped, only outlived** (confirmed). crossterm's reader thread busy-loops in `try_read`: `Ok(0)` falls through the inner loop without breaking, and non-WouldBlock errors like EIO do too (verified in the vendored `cachix/crossterm` `0.29-cachix` source, `src/event/source/unix/mio.rs`). The watchdog cannot reach that thread; the core stays pinned for the entire graceful teardown and indefinitely if teardown stalls. Since the repo already patches crossterm via `[patch.crates-io]` (root `Cargo.toml:209`), breaking the loop on `read() == 0` / hard errno in the fork kills the spin at the source and would let this watchdog only handle the cooperative exit.

5. **Exit code is 0 instead of 129** (`devenv-tui/src/app.rs:134`, confirmed). The watchdog never records a signal, but `devenv/src/main.rs:746-752` recovers the exit code from `Shutdown::last_signal()`, so a hangup-aborted session reports success to supervisors. It also delays the second-signal force-exit escalation by one signal (`last_signal != 0` check). `tokio_shutdown` already exposes `set_last_signal(Signal::SIGHUP)` for exactly this.

6. **macOS: `poll()` on `/dev/tty` reports POLLNVAL on a healthy terminal** (`devenv-tui/src/app.rs:123`, plausible). A long-standing Darwin limitation of the `/dev/tty` alias device. `devenv up < /dev/tty` (a common force-interactivity idiom) passes `is_terminal()`, then the first poll returns POLLNVAL and the watchdog kills the session within ~100ms of startup. Worth an empirical check on a Mac before merge, and it argues for requiring the condition to persist across a couple of polls before acting.

7. **Explicit `--tui` with non-tty stdin gets no protection** (`devenv-tui/src/app.rs:111`, confirmed). `--tui` bypasses the tty gate in `devenv/src/main.rs:329`, the watchdog silently returns `None`, iocraft supplies no input stream, and on PTY revocation without SIGHUP the process lingers forever heartbeat-rendering to dead stderr. A stdin-based watchdog cannot cover this case at all (a pipe never reports POLLHUP on terminal death); it would need a different fd, e.g. stderr.

8. **Thread-spawn failure is silently swallowed** (`devenv-tui/src/app.rs:141`). `.ok()?` discards the error and the caller never inspects the `Option`, so the watchdog can be absent with nothing in traces explaining why the old spin behavior returned. A `warn!` here would match the repo's tracing guidance for degraded paths.

9. **The one-shot `notify_waiters()` can be lost** (`devenv-tui/src/app.rs:133`, confirmed, bounded). `Notify::notify_waiters` stores no permit, and the watchdog fires exactly once from a foreign thread. If it lands while the render loop is between selects or during a view switch, exit falls back to the ~1s idle heartbeat. Never a hang (a view is always mounted and re-checks `exit_flag`), but the promptness silently depends on the heartbeat existing.

10. **Test may flake on old kernels** (`devenv-tui/src/app.rs:1020`, plausible). The zero-timeout POLLIN poll immediately after the master write races the ldisc flip-buffer push on kernels older than ~5.12, where pty input goes through a workqueue. Empirically clean on a modern kernel (0 misses in 400k iterations including 6-way concurrent load), so this only matters if CI ever runs on old kernels; a bounded retry loop would make it robust everywhere.

11. **`tui-replay --hold` ignores the shutdown token** (`devenv-tui/src/bin/tui-replay.rs:262`, low severity, predates this PR). Hold mode keeps `backend_done_tx` alive and its select has no shutdown branch, so after a disconnect the watchdog fires but the process parks forever. Adding a `wait_for_shutdown()` branch to the hold select fixes it.

## Cleanup

12. **Third hand-rolled copy of the exit sequence** (`devenv-tui/src/app.rs:132`). The `exit_flag.set()` + wake pair now exists at three sites (132, 288-294, 340-345), two of which carry a "kept in lockstep" comment. Extracting one `request_exit` helper (or moving the `Notify` inside `ExitFlag` so `set()` does both) removes the drift risk.

13. **A 10Hz wakeup thread for the whole TUI lifetime** (`devenv-tui/src/app.rs:68`). This sits next to the #2915 fix that deliberately reduced idle wakeups to a 1s heartbeat. The zero-timeout `poll_terminal()` check could piggyback on that existing heartbeat tick (1s detection latency is fine for a permanent condition), or use tokio `AsyncFd` for a fully event-driven version.

14. **`AtomicBool` + park/unpark vs the repo's stop-channel idiom** (`devenv-tui/src/app.rs:75`). `devenv-shell/src/session.rs:1020-1050` already uses `mpsc` `recv_timeout` for the same interval-thread shape; holding a `Sender` in the watcher and looping on `recv_timeout(interval)` is behaviorally identical with one mechanism instead of three.

15. **Inlined tty check** (`devenv-tui/src/app.rs:111`). `devenv_shell::terminal::is_tty()` is exactly this check and devenv-tui already depends on devenv-shell.

Refuted during verification, for the record: the `revents()` unknown-bits concern (both Linux and Darwin mask revents to requested-plus-ERR/HUP/NVAL, all modeled by nix 0.31), a watched-fd vs event-fd mismatch (iocraft never creates an event stream when stdin is not a tty), and "non-TUI modes are equally affected" (non-TUI devenv never reads the terminal, so there is no busy-loop counterpart).

alurm:
Thanks for the feedback. After looking into this further, I think the watcher may be a downstream workaround for bugs in crossterm and iocraft. I'm investigating fixing the error propagation upstream instead. If that proves viable, I'll update this PR to a smaller devenv change without the watcher; otherwise I'll return to hardening the current approach.

domenkozar:
Let's fix it upstream then :pray:

PR #3014, opened 2026-07-16
languages/rust: put cargo subcommands ahead of host rustup proxies

truongsinh:
cargo resolves external subcommands (`cargo clippy`, `cargo fmt`, `cargo miri`, ...) from `$CARGO_HOME/bin` before `$PATH`. On a host with a rustup install, `~/.cargo/bin`'s rustup proxies shadow a devenv-provided Rust toolchain, so `cargo clippy` and friends silently run the host toolchain instead of this environment's, with missing targets and mismatched compiler versions as typical symptoms.

Per the [cargo docs](https://doc.rust-lang.org/cargo/reference/external-tools.html#custom-subcommands), appending `$CARGO_HOME/bin` to `$PATH` overrides that precedence so devenv's toolchain wins. This keeps `~/.cargo` shared (crate cache, credentials, global config all intact), unlike isolating `CARGO_HOME`. It is unconditional and side-effect-free, so there is no module option.

Thanks @sandydoo for the PATH-append pointer.

### DIFF
--- a/src/modules/languages/rust.nix
+++ b/src/modules/languages/rust.nix
@@ -374,6 +374,15 @@ in
             ]
           })
           export PATH="$PATH:$CARGO_INSTALL_ROOT/bin"
+        ''
+        # cargo resolves external subcommands (cargo clippy, cargo fmt, ...)
+        # from $CARGO_HOME/bin *before* $PATH by default, so ~/.cargo/bin's
+        # rustup proxies would shadow this environment's toolchain. Append
+        # $CARGO_HOME/bin to $PATH to flip that precedence (per the cargo
+        # docs) so devenv's toolchain wins, while keeping the shared cache,
+        # credentials and global config in ~/.cargo intact.
+        + ''
+          export PATH="$PATH:''${CARGO_HOME:-$HOME/.cargo}/bin"
         '';
 
         packages =


domenkozar:
I'm not sure about this, it shouldn't be the default at least. `rustup` and such usually want their own way of doing things, might not be easy to please.

truongsinh:
Makes sense — flipped it to opt-in (`default = false`) in the latest push. The isolation only matters for the specific case of a devenv-provided Rust toolchain colliding with a host rustup install (cargo resolves external subcommands from `~/.cargo/bin` before `$PATH`), so defaulting off and letting people opt in when they hit it is the safer call. Kept the option and the docs explaining the failure mode so it stays discoverable.

sandydoo:
I found this curious tidbit in the [cargo docs](https://doc.rust-lang.org/cargo/reference/external-tools.html#custom-subcommands):

> Cargo defaults to prioritizing external tools in $CARGO_HOME/bin over $PATH. Users can override this precedence by adding $CARGO_HOME/bin to $PATH.

In theory, if we append the `CARGO_HOME` path to `PATH`, that should flip the priority:

```
export PATH="$PATH:${CARGO_HOME:-$HOME/.cargo}/bin"
```

Otherwise, the cost of this is just too high: duplicate crate caches (scales with number of projects), loss of credentials and global config. Though, perhaps it's possible to symlink those.

truongsinh:
Good catch, that's much cheaper than isolating CARGO_HOME. Appending `$CARGO_HOME/bin` to `$PATH` flips the precedence while keeping the shared cache, credentials and config intact. I'll swap to that. Cheap enough to be the default now, or keep it behind a flag?

sandydoo:
> Cheap enough to be the default now, or keep it behind a flag?

If this solves the precedence issue, you can set it as the default: no need for a module toggle at all.

domenkozar:
I'm curious why do we need this fix? Is that a bug in cargo or devenv?

sandydoo:
> Is that a bug in cargo or devenv?

It looks like a design mistake on cargo's end, but the [official workaround](https://doc.rust-lang.org/cargo/reference/external-tools.html#custom-subcommands) suggests to me that they were worried about breaking existing code if they fixed the default. So they added an escape hatch.

On our end, we obviously thought that PATH would take precedence over CARGO_HOME. The current changes look correct to me.

PR #3025, opened 2026-07-22
shell: use ghostty's change tracking to optimize renders

sandydoo:


devenv-shell/src/session.rs
```diff
@@ -419,15 +827,71 @@ impl Renderer {
...
+        self.normalize_fallback_boundary(vt_scrollback);
+        let boundary = self
+            .flush_boundary
+            .as_ref()
+            .and_then(|b| b.point(PointSpace::Screen).ok().flatten());
+        let start = boundary
+            .map(|p| p.y as usize)
+            .or(self.flush_boundary_rows)
+            .unwrap_or(0);
+        let resize_end = self
```
domenkozar: [P1] Prefer the tracked resize endpoint over the numeric fallback. `resize_flush_end_rows` is populated for every primary-screen shrink, so this `or_else` never consults `resize_flush_end`. When columns change in the same resize, the old numeric row is stale after Ghostty reflow. In a 10x5 to 4x4 reproduction, the numeric endpoint remained 3 while the tracked endpoint moved to 7; the renderer flushed only rows 0..3, re-anchored to the active viewport, cleared the resize marker, and permanently dropped the remaining pending rows. Please resolve the tracked point first and use the numeric value only when the pin is absent or invalid. The same precedence is needed in `unflushed()`.

domenkozar (changes requested):
Requesting changes because resize reflow can drop pending scrollback; details inline.

PR #3062, opened 2026-08-03
docs: added example for python with CUDA support

fcoelhomrc:
Added dedicated example in the documentation for using Python with CUDA support (common use case, eg. deep learning w/ PyTorch). 

Requires manually injecting the host CUDA driver path in `LD_LIBRARY_PATH` via `libraries` option. 

Maybe we should add a dedicated flag in Python configuration options? 

Screenshot from 2026-08-03 20-29-41

### DIFF
--- a/docs/src/languages/python.md
+++ b/docs/src/languages/python.md
@@ -255,6 +255,33 @@ devenv adds these libraries to `LD_LIBRARY_PATH` (Linux) or `DYLD_LIBRARY_PATH` 
 
 For more control over library search paths, use the `libraries` option.
 
+### CUDA support
+
+Some Python packages require CUDA support. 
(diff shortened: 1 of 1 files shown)


sandydoo:
Does this also need [`cuda_support`](https://devenv.sh/reference/yaml-options/#nixpkgscuda_support) enabled in devenv.yaml? `config_unfree`?

fcoelhomrc:
> Does this also need [`cuda_support`](https://devenv.sh/reference/yaml-options/#nixpkgscuda_support) enabled in devenv.yaml? `config_unfree`?

Needs `config_unfree` but `cuda_support` can be skipped.

I tested different combinations and this ended up being the minimum set of options that made PyTorch happy about my gpu

PR #3071, opened 2026-08-06
rust: preserve LLD with clang linker driver

domenkozar:
## Summary

- keep Clang as the default Rust linker driver on Linux, avoiding GCC's `collect2` hop
- preserve Rust's LLD backend on `x86_64-linux` when no explicit `mold`, `lld`, or `wild` linker is selected
- verify the actual linker backend while retaining project-level Cargo rustflags

## Root cause

#2981 configured Clang through Cargo's per-target linker setting to avoid GCC `collect2` hitting `ARG_MAX`. Setting an explicit linker driver, however, makes rustc stop adding the target's default self-contained LLD flags. Clang then falls back to GNU BFD `ld`.

A small Clang driver wrapper now adds `-fuse-ld=lld` on `x86_64-linux`. This retains the original `collect2` fix without setting `RUSTFLAGS`, so project `.cargo/config.toml` flags continue to work. Explicit devenv linker selections remain authoritative.

## User impact

Rust 1.90+ builds in default Linux Rust shells use LLD as intended instead of silently falling back to GNU BFD `ld`, avoiding the slower and substantially more memory-intensive linker behavior seen in parallel Cargo builds.

## Validation

- `devenv shell -- devenv-run-tests run tests --only rust`
- `devenv shell -- devenv-run-tests run tests --only rust-clang-linker-disabled`
- `devenv shell -- nixpkgs-fmt --check src/modules/languages/rust.nix tests/rust/devenv.nix`
- commit hooks (`nixpkgs-fmt`)

### DIFF
--- a/docs/src/languages/rust.md
+++ b/docs/src/languages/rust.md
@@ -206,6 +206,8 @@ This avoids GCC’s ` collect2 ` wrapper, which can hit ` Argument list too long
 in large Nix development environments before the final linker receives
 Rust’s response file\.
 
+On x86_64 Linux (glibc), Clang uses LLD unless another linker is enabled\.
+
 
 
 *Type:*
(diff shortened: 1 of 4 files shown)


domenkozar:
cc @Arshgill01

domenkozar:
`ld` is eating all my memory so I need this one merged so that `lld` is used.

PR #3077, opened 2026-08-09
feat: Add VHDL language support

kleindotc:
This PR adds support for VHDL language in devenv project:

features:
1.  Choose from 4 compilers: `ghdl-gcc`, `ghdl-llvm`, `ghdl-mcode`, or `nvc` (Default compiler `ghdl-gcc`).
2. `vhdl-ls` LSP support for VHDL.

### DIFF
--- a/src/modules/languages/vhdl.nix
+++ b/src/modules/languages/vhdl.nix
@@ -1,0 +1,51 @@
+{
+  pkgs,
+  config,
+  lib,
+  ...
+}: let
+  cfg = config.languages.vhdl;
+in {
+  options.languages.vhdl = {
+    enable = lib.mkEnableOption "tools for VHDL Development";
+    backend = lib.mkOption {
+      type = lib.types.enum ["ghdl-gcc" "ghdl-llvm" "ghdl-mcode" "nvc"];
+      default = "ghdl-gcc";
+      description = ''
+        The VHDL compiler backend to use.
+        - ghdl-gcc: GHDL with GCC backend
+        - ghdl-llvm: GHDL with LLVM backend
+        - ghdl-mcode: GHDL with built-in mcode backend
+        - nvc: VHDL compiler and simulator
+      '';
+    };
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = pkgs.ghdl-gcc;
+      description = "The VHDL package to use (automatically set based on backend).";
+      apply = backend: let
+        packages = {
+          "ghdl-gcc" = pkgs.ghdl-gcc;
+          "ghdl-llvm" = pkgs.ghdl-llvm;
+          "ghdl-mcode" = pkgs.ghdl-mcode;
+          "nvc" = pkgs.nvc;
+        };
+      in
+        packages.${backend} or (throw "Unknown VHDL backend: ${backend}");
+    };
+
+    lsp = {
+      enable = lib.mkEnableOption "VHDL Language Server" // {default = true;};
+      package = lib.mkOption {
+        type = lib.types.package;
+        default = pkgs.vhdl-ls;
+        defaultText = lib.literalExpression "pkgs.vhdl_ls";
+        description = "The VHDL language server package to use.";
+      };
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    packages = with pkgs; [cfg.package] ++ lib.optional cfg.lsp.enable cfg.lsp.package;
+  };
+}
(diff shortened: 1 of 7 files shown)


src/modules/languages/vhdl.nix
```diff
@@ -0,0 +1,50 @@
...
+        packages.${cfg.backend} or pkgs.ghdl-gcc;
+      description = "The VHDL package to use (automatically set based on backend).";
+    };
+
+    lsp = {
+      enable = lib.mkEnableOption "VHDL Language Server" // {default = true;};
+      package = lib.mkOption {
+        type = lib.types.package;
+        default = pkgs.vhdl-ls;
+        defaultText = lib.literalExpression "pkgs.vhdl_ls";
```
sandydoo:
```suggestion
        defaultText = lib.literalExpression "pkgs.vhdl-ls";
```

src/modules/languages/vhdl.nix
```diff
@@ -0,0 +1,50 @@
...
+        description = "The VHDL language server package to use.";
+      };
+    };
+  };
+
+  config = lib.mkIf cfg.enable {
+    packages = with pkgs; [cfg.package] ++ lib.optional cfg.lsp.enable cfg.lsp.package;
```
sandydoo:
```suggestion
    packages = [cfg.package] ++ lib.optional cfg.lsp.enable cfg.lsp.package;
```

src/modules/languages/vhdl.nix
```diff
@@ -0,0 +1,50 @@
...
+      type = lib.types.package;
+      default = let
+        packages = {
+          "ghdl-gcc" = pkgs.ghdl-gcc;
+          "ghdl-llvm" = pkgs.ghdl-llvm;
+          "ghdl-mcode" = pkgs.ghdl-mcode;
+          "nvc" = pkgs.nvc;
+        };
+      in
+        packages.${cfg.backend} or pkgs.ghdl-gcc;
```
sandydoo: 1. packages maps the backend type one-to-one. Use `pkgs` directly.
2. The `or` would otherwise hide errors in the mapping.

```suggestion
        pkgs.${cfg.backend};
```

src/modules/languages/vhdl.nix
```diff
@@ -0,0 +1,50 @@
...
+        The VHDL compiler backend to use.
+        - ghdl-gcc: GHDL with GCC backend
+        - ghdl-llvm: GHDL with LLVM backend
+        - ghdl-mcode: GHDL with built-in mcode backend
+        - nvc: VHDL compiler and simulator
+      '';
+    };
+    package = lib.mkOption {
+      type = lib.types.package;
+      default = let
```
sandydoo: Needs a `defaultText` of some kind to avoid leaking system/revision-specific store paths into the docs.

src/modules/languages/vhdl.nix
```diff
@@ -0,0 +1,50 @@
...
+  pkgs,
+  config,
+  lib,
+  ...
+}: let
+  cfg = config.languages.vhdl;
+in {
+  options.languages.vhdl = {
+    enable = lib.mkEnableOption "tools for VHDL Development";
+    backend = lib.mkOption {
```
sandydoo: I'm not too familiar with this field, but is `backend` the most accurate name for this? Perhaps `compiler`?

ghdl-gcc isn't supported on macOS, so another default should be set and documented:

```
default = if pkgs.stdenv.isDarwin then "ghdl-llvm" else "ghdl-gcc";
```
or `nvc` 🤷 

Something like the following could be a bit nicer, though again the mapping is one-to-one, so the only reason to keep it would be for readability and to share the map between the two options.

```nix
let
    compilerPackages = {
      ghdl-gcc = pkgs.ghdl-gcc;
      ghdl-llvm = pkgs.ghdl-llvm;
      ghdl-mcode = pkgs.ghdl-mcode;
      nvc = pkgs.nvc;
    };
  in {
    compiler = lib.mkOption {
      type = lib.types.enum (builtins.attrNames compilerPackages);
      default = if pkgs.stdenv.isDarwin then "ghdl-llvm" else "ghdl-gcc";
    };

    package = lib.mkOption {
      type = lib.types.package;
      default = compilerPackages.${cfg.compiler};
    };
  }
```

sandydoo (changes requested):
Overall, this needs a basic test in `/tests` so that we can catch regressions. Same goes for #3078 and #3091.

Run `devenv shell -- prek run -a` to format the code.

kleindotc:
Changes applied, thank you!

I will apply the changes to the other PRs when I have time. .w.

kleindotc:
Latest modifications for `languages.vhdl`, with this, it would be good to start developing.

PR #3080, opened 2026-08-10
Respect an explicit --system override in container_build on macOS

### LINKED ISSUE #3072: `devenv container build --system x86_64-linux` ignores target system on Apple Silicon
**Describe the bug**

On an Apple Silicon Mac, `devenv container build` ignores an explicit `--system x86_64-linux` argument and always selects the `aarch64-linux` container output.

This prevents building an AMD64 image through a configured `x86_64-linux` remote builder.

Container support on macOS was introduced by:

- #430
- #2085

PR #2085 appears to derive the container target directly from the architecture of the devenv binary:

```rust
let target_system = if host_os == "macos" {
    match host_arch {
        "aarch64" => "aarch64-linux",
        "x86_64" => "x86_64-linux",
        _ => bail!("Unsupported container architecture for macOS: {host_arch}"),
    }
} else {
    &self.global_options.system
};
```

Consequently, `self.global_options.system` is not considered on macOS.

General cross-compilation support is tracked in #145. This particular case should only require evaluating the existing `x86_64-linux` output and dispatching its derivations to a compatible builder.

**To reproduce**

Create a devenv project with a minimal container.

`devenv.yaml`:

```yaml
inputs:
  nixpkgs:
  nix2container:
    url: github:nlewo/nix2container
    inputs:
      nixpkgs:
        follows: nixpkgs
  mk-shell-bin:
    url: github:rrbutani/nix-mk-shell-bin
```

`devenv.nix`:

```nix
{ pkgs, ... }:

{
  containers.runtime = {
    name = "runtime";
    copyToRoot = [ pkgs.hello ];
    startupCommand = "${pkgs.hello}/bin/hello";
  };
}
```

On an `aarch64-darwin` host with a working `x86_64-linux` builder, run:

```sh
devenv container build --system x86_64-linux runtime
```

The command evaluates the ARM Linux output:

```text
Evaluating devenv.perSystem.aarch64-linux.containerBuilds.runtime.derivation
```

The equivalent Nix option does not change the selected output either:

```sh
devenv container build \
  --nix-option system x86_64-linux \
  runtime
```

It also evaluates:

```text
devenv.perSystem.aarch64-linux.containerBuilds.runtime.derivation
```

Expected behavior:

```text
Evaluating devenv.perSystem.x86_64-linux.containerBuilds.runtime.derivation
```

The resulting derivations should be dispatched to the configured `x86_64-linux` builder.

If selecting a different architecture is intentionally unsupported, devenv should reject the conflicting `--system` argument instead of silently ignoring it.

A possible solution would be to use an explicitly provided `--system` value and retain the current host-to-Linux mapping only as the default.

**Version**

```text
devenv 2.2.1 (aarch64-darwin)
```

yottanami:
Closes #3072

## What

`container_build` unconditionally derived `target_system` from the host CPU
architecture whenever `host_os == "macos"`, completely ignoring
`self.options.nix_settings.system`:

```rust
let target_system = if host_os == "macos" {
    match host_arch {
        "aarch64" => "aarch64-linux",
        "x86_64" => "x86_64-linux",
        _ => bail!("Unsupported container architecture for macOS: {host_arch}"),
    }
} else {
    &self.options.nix_settings.system
};
```

So `devenv container build --system x86_64-linux runtime` on Apple Silicon
always evaluated `devenv.perSystem.aarch64-linux...` instead of the
requested `x86_64-linux`, exactly as reported.

`devenv_core::settings::default_system()` never produces a `-linux` system
on macOS — it's always `<arch>-darwin`. So `nix_settings.system` ending in
`-linux` while running on a macOS host can only happen when the user
explicitly overrode it (via `--system` or `--nix-option system`, both of
which the issue confirms are silently ignored today). This checks for that
override first and only falls back to the host-arch → Linux mapping when
there isn't one:

```rust
let target_system = if self.options.nix_settings.system.ends_with("-linux") {
    self.options.nix_settings.system.as_str()
} else if host_os == "macos" {
    match host_arch {
        "aarch64" => "aarch64-linux",
        "x86_64" => "x86_64-linux",
        _ => bail!("Unsupported container architecture for macOS: {host_arch}"),
    }
} else {
    &self.options.nix_settings.system
};
```

Non-macOS behavior is unchanged (it already just used `nix_settings.system`
directly).

## Testing

I don't have a Rust toolchain in the environment I worked in, so I couldn't
run `cargo build`/`cargo test` or the project's own Nix-based build/test
workflow. I read `NixSettings::resolve` and `default_system()` in
`devenv-core/src/settings.rs` carefully to confirm the "`-linux` on macOS
implies an explicit override" assumption actually holds (it does — the
macOS default branch of `default_system()` only ever returns
`aarch64-darwin`/`x86_64-darwin`), and reviewed the type flow through the
`if`/`else if`/`else` by hand (all three branches coerce to `&str`, matching
the existing coercion the original code already relied on for its two
branches).

I'd appreciate a maintainer or CI running the actual build/test suite,
since I can't confirm that locally.

Marking this AI-assisted: this PR was authored by an AI coding agent
(Claude).

### DIFF
--- a/devenv-core/src/settings.rs
+++ b/devenv-core/src/settings.rs
@@ -176,9 +176,19 @@ impl NixSettings {
     /// falling back to `config.impure` when `None`.
     pub fn resolve(options: NixOptions, config: &Config) -> Self {
         let defaults = NixBuildDefaults::defaults();
+        // Nix options are applied after the dedicated settings, so the last
+        // `--nix-option system` value is the system Nix will actually use.
+        let system = options
+            .nix_options
+            .chunks_exact(2)
+            .rev()
+            .find(|pair| pair[0] == "system")
+            .map(|pair| pair[1].clone())
+            .or(options.system)
+            .unwrap_or_else(default_system);
         Self {
             impure: options.impure.unwrap_or(config.impure),
-            system: options.system.unwrap_or_else(default_system),
+            system,
             max_jobs: options.max_jobs.unwrap_or(defaults.max_jobs),
             cores: options.cores.unwrap_or(defaults.cores),
             offline: options.offline.unwrap_or(false),
@@ -575,6 +585,36 @@ mod tests {
     fn nix_settings_system_from_options() {
         let options = NixOptions {
             system: Some("x86_64-linux".into()),
+            ..Default::default()
+        };
+        let config = Config::default();
+        let settings = NixSettings::resolve(options, &config);
+        assert_eq!(settings.system, "x86_64-linux");
+    }
+
+    #[test]
+    fn nix_settings_system_from_nix_options() {
+        let options = NixOptions {
+            nix_options: vec!["system".into(), "x86_64-linux".into()],
+            ..Default::default()
+        };
+        let config = Config::default();
+        let settings = NixSettings::resolve(options, &config);
+        assert_eq!(settings.system, "x86_64-linux");
+    }
+
+    #[test]
+    fn nix_settings_last_system_nix_option_wins() {
+        let options = NixOptions {
+            system: Some("aarch64-linux".into()),
+            nix_options: vec![
+                "system".into(),
+                "i686-linux".into(),
+                "sandbox".into(),
+                "false".into(),
+                "system".into(),
+                "x86_64-linux".into(),
+            ],
             ..Default::default()
         };
         let config = Config::default();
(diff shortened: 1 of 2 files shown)


domenkozar:
Pushed a commit to address `--nix-option`.

PR #3089, opened 2026-08-13
fix Zig version input source

### LINKED ISSUE #3088: Cannot activate environment with Zig support
**Describe the bug**
I'm having trouble getting a devenv shell setup with Zig support. The primary error I'm seeing now is,

```
error: path '/nix/store/nzi23xj0cxs4mghp7rf0md87mfaflicf-source/pkgs/development/compilers/zig/passthru.nix' does not exist
```

**To reproduce**

devenv.nix
```nix

{
  languages.zig = {
    enable = true;
    version = "0.16.0";
  };
}
```

devenv.yaml
```yaml
# yaml-language-server: $schema=https://devenv.sh/devenv.schema.json
inputs:
  nixpkgs:
  zig-overlay:
    url: github:mitchellh/zig-overlay
    inputs:
      nixpkgs:
        follows: nixpkgs
```

Full `devenv.lock` and log output is here: https://gist.github.com/nickmonad/258325869ef4fbe21849448a13b865b1

**Version**
```
devenv 2.2.1 (x86_64-linux)
```

I don't believe I'm using flakes in this context. At least, I haven't activated it in my primary NixOS configuration as far as I can tell. (Sorry, I'm still quite new to nix/nixos in general!)

Appreciate any help or guidance!

domenkozar:
## Summary

- import `zig-overlay` with the active package set and its real Nixpkgs source path
- preserve support for `zig-overlay.inputs.nixpkgs.follows = "nixpkgs"`
- add an end-to-end regression test for Zig 0.16.0

## Root cause

`zig-overlay` imports `pkgs/development/compilers/zig/passthru.nix` directly from its `nixpkgs` input. When that input follows `cachix/devenv-nixpkgs`, the input source is the wrapper repository rather than the Nixpkgs source used to construct `pkgs`, so the imported path does not exist.

Importing `zig-overlay/default.nix` with `nixpkgs = pkgs.path` keeps the package set and its source aligned.

## Impact

Projects using `languages.zig.version` can keep the recommended Nixpkgs follow when their root input is `cachix/devenv-nixpkgs`.

Fixes #3088.

## Validation

- `devenv-run-tests run --only zig-version tests`
- `nixpkgs-fmt --check src/modules/languages/zig.nix tests/zig-version/devenv.nix`
- parsed both changed Nix files with `nix-instantiate --parse`
- parsed the regression fixture with `yaml2json`
- `git diff --check`

### DIFF
--- a/src/modules/languages/zig.nix
+++ b/src/modules/languages/zig.nix
@@ -25,6 +25,15 @@ let
     follows = [ "nixpkgs" ];
   };
 
+  # zig-overlay imports files directly from its nixpkgs input. devenv-nixpkgs
+  # is a wrapper flake, so its source root is not the package set's source.
+  # Import the overlay with the source that produced pkgs instead.
+  zigPackages = import "${zig-overlay.outPath}/default.nix" {
+    inherit pkgs;
+    nixpkgs = pkgs.path;
+    system = pkgs.stdenv.hostPlatform.system;
+  };
+
   zls = config.lib.getInput {
     name = "zls";
     url = "github:zigtools/zls/${zlsVersion}";
@@ -70,7 +79,7 @@ in
 
   config = lib.mkIf cfg.enable {
     languages.zig.package = lib.mkIf (cfg.version != null) (
-      zig-overlay.packages.${pkgs.stdenv.system}.${cfg.version}
+      zigPackages.${cfg.version}
     );
 
     languages.zig.lsp.package = lib.mkIf (cfg.version != null) (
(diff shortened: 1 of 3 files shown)


tonyd33:
@domenkozar `devenv-run-tests run --only zig-version tests` worked for me but it fails with `lsp.enable = true;`

```
...
  │ error: To use 'languages.zig.version', run the following command:
  │
  │   $ devenv inputs add zls github:zigtools/zls/0.16.0 --follows nixpkgs --follows zig-overlay
```

PR #3093, opened 2026-08-14
feat(tasks): improve tasks list output

b4nst:
Small proposal to enhance `devenv tasks list` output.

I feel like the description was lacking from this output, sometimes making it hard to navigate. I took the opportunity to also change the way to display tasks attributes (watch, status, etc...) however this is quite opinionated, as any UI changes.

We move all our moon tasks towards devenv native tasks, and felt we lost some visibility in the process. Currently we're relying on a custom wrapper using the `--json` arg, so this is not blocking. However we'd prefer not having wrappers to maintain.

Thanks!


domenkozar:
Would be great to see before/after

b4nst:
Here you go (note that I do have a really minimal theme so colour are not really representative):

image

_edit here's a more traditional / bare terminal theme_

image

domenkozar:
Cool :heart_eyes: rebase on main and let's merge

b4nst:
done (sorry for the delay, wrapped my day before seeing your message)

PR #3096, opened 2026-08-16
fix(cli): merge devenv.yaml of a --from flake reference

Allsimon:
`--from path:...` merged the source's devenv.yaml, but a flake reference only registered the source as an input named `from` and imported its devenv.nix. The source's inputs and imports were dropped, so a project using an integration such as treefmt or git-hooks failed to evaluate.

Flake references are now fetched into the store before the configuration is read, then merged like a `path:` source, so both forms share one code path. The reference itself is no longer recorded in devenv.lock; pin it with an explicit revision instead.

### DIFF
--- a/devenv/src/devenv/mod.rs
+++ b/devenv/src/devenv/mod.rs
@@ -125,6 +125,9 @@ pub struct DevenvOptions {
     pub secret_settings: SecretSettings,
     pub input_overrides: InputOverrides,
     pub from_external: bool,
+    /// `devenv.lock` of a `--from` source, used as the starting point for a
+    /// project that has none of its own. See [`Devenv::new`].
+    pub from_lock: Option<PathBuf>,
     pub require_version_match: bool,
     pub devenv_root: Option<PathBuf>,
     pub devenv_dotfile: Option<PathBuf>,
@@ -189,6 +192,7 @@ impl Default for DevenvOptions {
             secret_settings: SecretSettings::default(),
             input_overrides: InputOverrides::default(),
             from_external: false,
+            from_lock: None,
             require_version_match: false,
             devenv_root: None,
             devenv_dotfile: None,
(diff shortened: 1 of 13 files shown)


tests/cli/.test.sh
```diff
@@ -55,6 +55,32 @@ for path in "path:$from_test_dir" "path:./from-test"; do
...
+echo "$out" | grep -q "from-flake-import" || fail "--from flake ref dropped the source's imports"
+! echo "$out" | grep -q "python3" || fail "--from flake ref leaked local python3"
+
+step "--from seeds devenv.lock from the source"
+# from-flake pins flake-compat to an old revision in its lock while leaving it
+# unpinned in devenv.yaml. A project with no lock of its own starts from the
+# source's, so this must keep that revision instead of resolving the current
+# head of flake-compat.
+from_flake_dir="$(cd from-flake && pwd)"
+pinned_rev=$(sed -n 's/.*"rev": "\([0-9a-f]*\)".*/\1/p' from-flake/devenv.lock)
```
domenkozar, line 76: This test fails before it exercises lock seeding because `from-flake/devenv.lock` is neither included in this PR nor generated before this command. The earlier `devenv` invocation writes the project-root lock, not one inside `from-flake`. Please add or generate the source lock before initializing and committing the nested repository.

devenv/src/main.rs
```diff
@@ -432,7 +432,23 @@ fn prepare_command(mut cli: Cli, shell_hint: Option<&str>) -> Result<PreparedCom
             fs::canonicalize(&full_path).unwrap_or(full_path)
         });
 
-    let mut config = Config::load_with_source(from_path.as_deref())?;
+    // Any other source (a flake ref, via --from or a persisted binding) is
+    // fetched into the store first, then merged from there like a `path:`
+    // source: its devenv.yaml only reaches the config loader if the source is
+    // already on disk when the config is read.
+    let from_dir = match (&from_source, &from_path) {
+        (Some(from), None) => Some(devenv_nix_backend::source::fetch(from)?),
```
domenkozar, line 440: This fetch runs before `NixSettings::resolve` and `apply_nix_settings`, while `source::fetch` applies only the experimental features. As a result, CLI settings needed for private sources (`access-tokens`, `netrc-file`) and fetch controls such as offline/refresh are ignored. Please pass the relevant CLI settings into this prefetch path and apply them before opening the store.

devenv/src/devenv/mod.rs
```diff
@@ -533,6 +537,25 @@ impl Devenv {
...
+        // that has no lock of its own starts from those instead of resolving
+        // every input afresh; the seeded lock then belongs to this project, so
+        // `devenv update` moves it independently of the source. Locking runs
+        // next and reconciles it with the merged input set.
+        if let Some(source_lock) = &options.from_lock {
+            let lock = devenv_root.join("devenv.lock");
+            if !lock.exists()
+                && let Err(e) = seed_lock_from_source(source_lock, &lock).await
+            {
+                warn!(
```
domenkozar, line 551: If the source lock exists but cannot be read, this falls back to resolving fresh revisions, but `tracing::warn!` is hidden in a normal invocation. Because this changes the environment away from the source's tested revisions, please emit the fallback through the activity system at warning level so ordinary users see it.

domenkozar (commented):
Inline review findings from the current head.

domenkozar:
@Allsimon ping :)

Allsimon:
Sorry about the noise, I initially thought that `devenv build --from github:cachix/devenv` was just the analog of `nix build github:cachix/devenv`

But it seems to be way more complicated than that:
- the source's devenv.yaml is merged at the lowest precedence under the local config
- it try to merge/create devenv.lock (while ignoring the source devenv.lock)
- yet the local devenv.nix is ignored (?)

I'm not completely sure I understand the use-cases of `devenv build --from ...`  yet so I drafted this PR until I get a clear picture

PR #3100, opened 2026-08-18
fix(tasks): complete dependencies and propagate process failures

waneon:
## Summary

- complete the prerequisite closure for dependent tasks selected by `RunMode::All`
- propagate native process-task failures from `devenv up` and `devenv test`
- stop already-started sibling processes when startup tasks fail
- add unit and integration regression coverage

## Background

#2348 split `RunMode::All` into independent incoming and outgoing traversals to avoid direction bouncing through shared prerequisites. That correctly excludes unrelated dependents, but a dependent reached from a process root can have another prerequisite that is never selected.

Separately, the native startup path ignored the task runner's completion status, so a failed process-dependent task could still let `devenv test` exit successfully.

This was observed in https://github.com/waneon/neoseq/actions/runs/32113974737: the browser task started without its web build prerequisite, failed because `apps/client/dist` was absent, and the workflow was reported as successful.

The added closure pass follows only incoming edges from the already-selected set, preserving #2348's no-direction-bouncing behavior.

## Tests

- `cargo test -p devenv-tasks` (142 passed)
- `devenv-run-tests run tests --only process-task-failure`
- `nix build .#devenv`
- `cargo fmt --all -- --check`
- `nixpkgs-fmt --check tests/process-task-failure/devenv.nix`
- `shellcheck tests/process-task-failure/.test.sh`

`cargo clippy -p devenv-tasks --all-targets -- -D warnings` currently stops on the pre-existing `clippy::for-kv-map` warning in `devenv-core/src/cachix.rs:621`.

Developed with Codex.

### DIFF
--- a/devenv-tasks/src/tasks.rs
+++ b/devenv-tasks/src/tasks.rs
@@ -586,6 +586,23 @@ impl Tasks {
                         }
                     }
                 }
+
+                // A selected dependent may have prerequisites other than the root that
+                // brought it into the schedule. Include those prerequisites without
+                // traversing back out from them, which preserves the no-direction-
+                // bouncing behavior above while ensuring every scheduled task has its
+                // complete dependency closure.
+                to_visit.extend(visited.iter().copied());
+                while let Some(node) = to_visit.pop() {
+                    for neighbor in self
+                        .graph
+                        .neighbors_directed(node, petgraph::Direction::Incoming)
+                    {
+                        if visited.insert(neighbor) {
+                            to_visit.push(neighbor);
+                        }
+                    }
+                }
             }
         }
 
--- a/devenv-tasks/src/tests/mod.rs
+++ b/devenv-tasks/src/tests/mod.rs
@@ -3469,6 +3469,64 @@ async fn test_run_mode_all_excludes_unrelated_entry_points() -> Result<(), Error
     Ok(())
 }
 
+/// RunMode::All must include every prerequisite of a dependent selected from
+/// the root's downstream graph, without pulling in unrelated dependents of a
+/// shared prerequisite.
+#[tokio::test]
+async fn test_run_mode_all_includes_selected_dependents_prerequisites() -> Result<(), Error> {
+    let temp_dir = TempDir::new().unwrap();
+    let db_path = temp_dir.path().join("tasks.db");
+
+    let tasks = Tasks::builder(
+        Config::try_from(json!({
+            "roots": ["devenv:processes:server"],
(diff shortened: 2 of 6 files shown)


domenkozar:
Thanks for this, and for the detailed background linking it to #2348. I pulled the branch and verified both fixes against current `main` (it test-merges with zero conflicts, and 142 `devenv-tasks` unit tests pass).

**The closure fix is well covered.** I rebuilt with only the `devenv-tasks/src/tasks.rs` hunk reverted to check the integration test actually guards it:

```
both fixes        → exit 0
no closure fix    → exit 1, PROCESS_TASK_PREREQUISITE_MISSING
```

Confirms `test:prerequisite` is reachable only through the new third pass: `devenv test` schedules with the process tasks as roots in `RunMode::All`, so pass 1 finds nothing upstream of `server`, pass 2 picks up `test:failure` going outward, and the prerequisite sits on an incoming edge of a node only discovered in pass 2. The `unrelated:test` assertion in the unit test is a nice touch too, since it pins the #2348 property the extra pass could have undone.

**One assertion does not bite.** The leftover process check:

```bash
if test -f "$pid_file" && kill -0 "$(cat "$pid_file")" 2>/dev/null; then
  echo "FAIL: process sibling was left running after task failure"
```

I removed only the `stop_all()` line, keeping the status check and `bail!`, and the test still passed:

```
stop_all removed  → exit 0, server.pid alive=no
```

The cause is `exec sleep 300`. That makes the payload the manager's direct child, and tokio's kill on drop reaps it while devenv unwinds out of `bail!`, whether or not `stop_all()` ran. So the one line the assertion exists to protect is currently untested.

Backgrounding the payload makes it a grandchild, the way a service that forks is, and only signalling the session reaches it:

```nix
  processes.server = {
    exec = ''
      mkdir -p ${markerDir}
      sleep 300 &
      echo "$!" > ${markerDir}/server.pid
      touch ${markerDir}/server.ready
      wait
    '';
```

With that change:

```
stop_all removed + backgrounded  → exit 1, FAIL: process sibling was left running
both fixes       + backgrounded  → exit 0
```

Worth knowing generally: any devenv test asserting a process was cleaned up needs to background its payload rather than `exec` it. We hit the same trap in #3103. Everything else here looks good to me.

One note on CI: this last ran against base `c18de525`, which predates #3103 and #3104. A rebase would be worth it before merge.

*Posted by Claude Code on behalf of @domenkozar. Context: https://github.com/cli/cli/issues/13904*

PR #3104, opened 2026-08-19
Refine explanation of pin resolution process

chrissound:


chrissound:
Following on from https://github.com/cachix/devenv/pull/3099#event-29658049542

Issue #3040, opened 2026-07-28
devenv shell --no-tui shows TUI


supermarin:
**Describe the bug**

`devenv shell --no-tui` still shows the status bar on the bottom.
Not sure if this is by design, but at most I'd expect the prompt change, not another tmux-like shell.

<img width="1018" height="425" alt="Image" src="https://github.com/user-attachments/assets/f4e0dbcc-8781-41ec-94a7-8156d20cad53" />

**To reproduce**

`devenv shell --no-tui` 

**Version**

devenv 2.1.2 (x86_64-linux)

sandydoo:
Ah, that _is_ quite confusing. `--no-tui` disables the interactive interface for evals, builds, processes, etc.
The status bar can is linked to the `reload` feature and can be disable with `--no-reload`.

sandydoo:
Related to https://github.com/cachix/devenv/issues/2793

Issue #2995, opened 2026-07-11
Enable Auto-activation without using the TUI


rohrsben:
As in title! I can't find a way to disable the TUI (and reloads/file watching) when the devenv environment was entered using auto-activation. I have no idea if this is relevant, but personally I only care about the auto-_entering_ the shell; I don't mind exiting the shell on my own.

yurrriq:
I find the TUI extremely distracting, and can't find a way to actually disable it either. I tried `--no-tui` and `DEVENV_TUI=false`. I'm on v2.1.2. This was my first time trying the direnv replacement, but I guess I'll go back to that for now.

cwegener:
Same here. I would very much appreciate an option to disable the TUI when running `devenv shell`.
I'm happy with the subshell itself, but the 'sub-terminal' can be really awkward to use (especially when it comes to scrollback)

I'm not sure what the better way is, disabling the TUI, or collecting feedback about individual users' gripes with the TUI and address them one-by-one.

domenkozar:
So you'd want just the statusline not to be shown?

`statusline = "off"`
`statusline = "top"`

rohrsben:
Personally, I would want auto-activation to be the same as manually running `devenv shell --no-tui --no-reload`. I have not yet had the chance to try out/need `devenv up`, but perhaps the most general solution would be like a "devenv auto-activation command" setting in `devenv.yaml`?

cwegener:
> So you'd want just the statusline not to be shown?
> 
> `statusline = "off"` `statusline = "top"`

In my case the issue is more about the wonky terminal scrollback when inside of the embedded terminal and not actually about the statusline.

Maybe my issue is rather different from what @rohrsben is describing.

The scrollback issue is not a super high priority issue for me. If it starts annoying me more, I'll just create a new GH issue for it.

domenkozar:
See https://github.com/cachix/devenv/pull/2918

Urie96:
I used a temporary workaround by modifying the code generated by `devenv hook`. In config.fish, I changed `devenv hook fish | source` to:

`devenv hook fish | sed 's/and devenv shell/and devenv shell --no-tui --no-reload/g' | source`

PR #3117, opened 2026-08-24
tui: add configurable viewport and user settings

### LINKED ISSUE #2662: TUI: Ability to remap keybinds
## Problem

As of now, what keys are needed to be pressed to perform an action in TUI is hardcoded:
1. https://github.com/cachix/devenv/blob/v2.0.6/devenv-tui/src/app.rs#L417-L435
1. https://github.com/cachix/devenv/blob/v2.0.6/devenv-tui/src/app.rs#L587-L646

This will inevitably lead (ackthually, already does: https://github.com/cachix/devenv/issues/2659) to keybind conflicts with other software (such as terminal, desktop environment, compositor, etc.).

## Proposed solution

As such in the long term, these keybinds ought to be configurable by user (without recomiplation).
I propose it to be done from within `devenv.nix`:
```nix
{...}: {
    devenv.keybinds = {
        reload = [
            "CTRL"
            "META"
            "R"
        ];
        stopWatching = [
            "CTRL"
            "ALT"
            "SHIFT"
            "D"
        ];
    };
}
```

Then it can be decided per-project, should this reside directly in `devenv.nix` (for homogenous and all-batteries included setups), or as a local override in `.devenv.local.nix`.

Third option would be to put this in a "global" `.devenv.local.nix`. This might be preferable for many people, as for someone working on 3 monorepos each with 10-20 `devenv.nix`, maintaining 30-60 `.devenv.local.nix` is unfathomable.

## TL;DR

- [ ] Change `devenv-tui` source code to no longer hardcode `KeyCode` values, but take them from configuration
- [ ] [Optional] Use existing hardcoded values as defaults for said configuration
- [ ] Introduce field `.devenv.keybinds` or equivalent to `devenv.nix`, which would be used to override keybinds
- [ ] [Optional] Think of a way to do this override "globally" (i.e. for all `devenv.nix` some user would use)

### LINKED ISSUE #2793: Hide status/bottom bar in the TUI shelll
Hi,

Is it possible to hide the bottom/status bar in the shell?
I find it quite annoying that the bottom bar is always visible when using nvim for example.

Image

### LINKED ISSUE #3140: Allow remapping or disabling `devenv shell` status line keybindings
### Summary

`devenv shell` claims four fixed key chords: `Ctrl-Alt-D`, `Ctrl-Alt-W`, `Ctrl-Alt-E`, and `Ctrl-Alt-R`. Only one of them can be changed, and only on zsh. There is currently no way to relocate the other three, and no way to release them.

This causes a collision with my existing setup, where I am using `Ctrl-Alt` as my `kitty_mod` in Kitty, and most `Ctrl-Alt-*` chords are bound in Kitty.

### The problem

```
keypress -> terminal emulator -> multiplexer -> devenv statusline -> shell
```

These chords live in a namespace devenv shares with terminal emulators, multiplexers, and shells. A collision is therefore not unusual, and it resolves badly in both directions.

If the terminal or multiplexer binds the chord first, it never reaches devenv. The devenv feature is then silently unreachable, with no alternate binding and no indication that anything is missing.

If devenv consumes the chord, it never reaches the shell. A fish or zsh user's own binding is then silently dead inside `devenv shell`, and cannot be recovered by editing their shell config.

Either way the user loses a binding they cannot restore. Both cases are solved by the same two changes: make the bindings configurable, and provide an opt-out.

### There is no opt-out today

`classify_stdin` matches these byte sequences unconditionally ([session.rs#L1563](https://github.com/cachix/devenv/blob/97135e8/devenv-shell/src/session.rs#L1563)). It is not gated on `show_status_line`, and `--no-tui` / `DEVENV_TUI=false` affect only the `devenv up` TUI.

The practical result is that disabling the status line entirely does not release the three chords bound to it.

### The precedent already exists

`DEVENV_RELOAD_KEYBIND` already makes the reload binding configurable ([zsh.rs#L127](https://github.com/cachix/devenv/blob/97135e8/devenv-shell/src/dialect/zsh.rs#L127)). It applies to zsh only, and it does not appear in the documentation.

The equivalent bindings in fish ([fish.rs#L135](https://github.com/cachix/devenv/blob/97135e8/devenv-shell/src/dialect/fish.rs#L135)) and nushell ([nushell.rs#L253](https://github.com/cachix/devenv/blob/97135e8/devenv-shell/src/dialect/nushell.rs#L253)) are hardcoded. The other three chords are hardcoded everywhere ([session.rs#L49-L51](https://github.com/cachix/devenv/blob/97135e8/devenv-shell/src/session.rs#L49-L51)).

So this is a request for consistency rather than a new capability.

It is also not the first collision of this kind. The `2.1` release notes describe `2.0`'s `Ctrl+Alt+R` clashing with reverse search on macOS, which #2595 resolved by making reload automatic. That fix addressed one chord on one platform, and the general case remains.

## Proposal

In rough order of preference:

1. A user-level config section for all four bindings, honored across every shell dialect.
2. Failing that, extend the `DEVENV_RELOAD_KEYBIND` pattern to the remaining three chords, and support it on every dialect rather than zsh alone.
3. At minimum, a documented way to disable status line key interception entirely, so that users with conflicts can opt out.

Accepting an empty value as "unbound" would cover the escape hatch in all three designs.

One design note: this should be user-level configuration rather than a `devenv.nix` option. `devenv.nix` is committed and shared across a team, whereas keybinding conflicts are specific to an individual's terminal and shell.

### Smaller related point

Since `2.1` applies environment changes automatically at the next prompt, `Ctrl-Alt-R` is largely redundant. It may be worth leaving it unbound by default and letting users opt in.

### Environment

- **devenv version**: devenv `2.2.2 (x86_64-linux)`
- **Shell**: fish `4.6.0`
- **Terminal emulator**: [kitty](https://sw.kovidgoyal.net/kitty/) `0.47.1`
- **Multiplexer**: [herdr](https://github.com/herdrdev/herdr) `0.8.2`

Source references are pinned to `97135e8`, so line numbers may differ from the released version.

vimjoyer:
Fixes #2662
Fixes #3140
Fixes #2793

Related to #3040 and #2995.


devenv-tui/src/expanded_view.rs
```diff
@@ -865,7 +984,14 @@ fn handle_log_search_key(
                 }
                 log_search.set(Some(search));
             }
-            KeyCode::Char(character) if !control && !alt => {
+            None if !key_consumed
+                && matches!(key_event.code, KeyCode::Char(_))
+                && !control
+                && !alt =>
```
domenkozar, line 991: With the default keymap, `n` and `shift+n` resolve to `NextMatch` and `PreviousMatch` in the `LogSearch` context, making `key_consumed` true. While the search is still being edited, those actions are not handled, and this guard prevents the raw character from being appended, so common queries containing `n` cannot be entered. Navigation bindings should only consume these characters after editing is complete.

devenv-tui/src/app.rs
```diff
@@ -156,6 +163,18 @@ impl TuiApp {
         self
     }
 
+    pub fn with_preferences(mut self, preferences: TuiPreferences) -> Self {
+        self.config.max_log_lines_per_build = preferences.behavior.log_history_lines;
+        self.config.log_viewport_collapsed = preferences.behavior.log_preview_lines;
```
domenkozar, line 199: `log_preview_lines` is copied into `TuiConfig::log_viewport_collapsed`, but that field is never read when rendering previews; `ExpandedContentComponent::new` continues to use the fixed `LOG_VIEWPORT_COLLAPSED` constant. Consequently, every valid non-default `tui.behavior.log_preview_lines` setting is silently ignored.

domenkozar:
This also addresses #2793: users have requested a way to hide the persistent bottom/status bar (including because it interferes with terminal applications). The new `tui.statusline.enabled: false` setting provides that capability while still showing active search and confirmation prompts.

vimjoyer:
# Changes

- add a versioned personal YAML configuration file for TUI and interactive shell preferences
- load user configuration from $XDG_CONFIG_HOME/devenv/config.yaml, with a ~/.config/devenv/config.yaml fallback
- add --user-config and DEVENV_USER_CONFIG overrides without mixing personal preferences into project configuration
- add devenv config path, config validate, config show, and config schema commands that work without loading a project
- report unknown fields, duplicate keys, invalid ranges, unsupported formats, conflicting shortcuts, and ambiguous key sequences
- add a generated JSON Schema for editor completion and static validation
- keep the default inline viewport at the invocation position while preserving terminal content above it
- scroll the inline viewport only when growing content reaches the terminal bottom
- add a top-aligned viewport mode that consistently claims the visible terminal
- allow the statusline to remain inline, stick to the top or bottom, or be disabled
- preserve search and Ctrl-C confirmation prompts when the ordinary statusline is disabled
- add independent statusline layouts for the main view, fullscreen logs, search, and confirmation prompts
- arrange statusline components independently across left, center, and right sections
- expose activity counts, process state, profiles, project details, command, shell, elapsed time, selection, log position, retained logs, search state, pending keys, and key hints as components
- add custom text and typed statusline components with configurable formats
- compact, hide, or truncate lower-priority components as the terminal becomes narrower
- preserve required components and truncate safely by terminal display width
- add devenv, terminal-native, and unstyled theme presets
- support named terminal colors, ANSI palette colors, RGB colors, and reusable custom palette entries
- customize the whole statusline, separators, or individual components with foreground, background, and text modifiers
- make TUI keybindings configurable independently for the main view, process search, fullscreen logs, log search, and prompts
- support disabling default actions with empty binding lists
- support multi-key sequences with a configurable timeout and visible pending-key state
- derive displayed key hints from the resolved configuration
- make interactive shell shortcuts configurable for Bash, Fish, Nushell, and Zsh
- preserve shell-specific reload defaults while allowing them to be overridden or disabled
- forward configured shell shortcuts to applications while they own the alternate screen
- allow mouse capture to be enabled or completely disabled
- configure the initial stopped-process filter and fullscreen log-following behavior
- configure collapsed log preview size and retained log history limits
- apply configured preview limits consistently to rendering and terminal-fit calculations

PR #3120, opened 2026-08-25
feat(claude): add claude.code.skills

paulbertin:
Small addition to the Claude Code integration: skills, alongside the existing commands and agents.

My motivation is that most `.claude/` is generated in our projects, but skills were the one part we still manage outside devenv.

They're generated to `.claude/skills/<name>/SKILL.md` through the same `files` mechanism `claude.code.commands` and `claude.code.agents` already use. `resources` places extra files next to `SKILL.md` and mirrors `files`, including `executable`. `copyMode` is passed through too, so `seed` lets you iterate on a skill in place before moving the prose back into `devenv.nix`.

Smaller note: I emit the description with `builtins.toJSON` so it stays valid YAML. `claude.code.agents` does `description: ${agent.description}` unquoted, so `description: Reviews code: checks quality` is invalid, and a ` #` in a description silently truncates it. Happy to fold this one-line fix along here.

Thanks!

### DIFF
--- a/src/modules/integrations/claude.nix
+++ b/src/modules/integrations/claude.nix
@@ -40,6 +40,29 @@ let
     };
   };
 
+  # A file bundled next to a skill's SKILL.md. A bare path is the common case;
+  # the attribute set form exists because a bundled script cannot be made
+  # executable after the fact: the symlink target is read-only in the store, and
+  # a copied file is rewritten on every shell entry.
+  skillResourceType = lib.types.coercedTo lib.types.path (source: { inherit source; }) (
+    lib.types.submodule {
+      options = {
+        source = lib.mkOption {
+          type = lib.types.path;
+          description = "File to place next to SKILL.md.";
+        };
+        executable = lib.mkOption {
+          type = lib.types.bool;
+          default = false;
+          description = ''
+            Make the file executable, for a resource that is run directly
+            rather than passed to an interpreter.
+          '';
+        };
+      };
+    }
+  );
+
   # Reserved keys that are not tool names (for backward compat detection)
   reservedPermissionKeys = [ "defaultMode" "disableBypassPermissionsMode" "additionalDirectories" "rules" ];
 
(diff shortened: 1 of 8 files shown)


src/modules/integrations/claude.nix
```diff
@@ -455,6 +478,108 @@ in
...
+              default = [ ];
+              internal = true;
+              visible = false;
+              description = "Assertions raised by this skill's submodule, collected into the top-level assertions.";
+            };
+          };
+
+          config.assertions = [
+            {
+              assertion = builtins.match "[a-z0-9]+(-[a-z0-9]+)*" name != null && builtins.stringLength name <= 64;
```
sandydoo, line 653: Looks like `synced` is a reserved directory name.

https://code.claude.com/docs/en/skills#where-skills-live

> The folder name synced is reserved in the enterprise, personal, and project skills locations, in any capitalization. Claude Code downloads the skills you enable on claude.ai into ~/.claude/skills/synced/ when CLAUDE_CODE_SYNC_SKILLS is set in non-interactive mode, and skips a skill you author at that name.

src/modules/integrations/claude.nix
```diff
@@ -455,6 +478,108 @@ in
...
+              '';
+            };
+            content = lib.mkOption {
+              type = lib.types.lines;
+              description = "The body of SKILL.md, loaded when the skill triggers.";
+            };
+            allowedTools = lib.mkOption {
+              type = lib.types.listOf lib.types.str;
+              default = [ ];
+              description = "Restrict the tools available while this skill is active. Empty means no restriction.";
```
sandydoo: Is this description accurate? I think it allows tool use without prompting, rather than restricting.

src/modules/integrations/claude.nix
```diff
@@ -455,6 +478,108 @@ in
...
+                What the skill covers and when Claude should load it.
+                Descriptions are the only part of a skill kept in context at all
+                times, so this is what decides whether the skill ever triggers.
+              '';
+            };
+            content = lib.mkOption {
+              type = lib.types.lines;
+              description = "The body of SKILL.md, loaded when the skill triggers.";
+            };
+            allowedTools = lib.mkOption {
```
sandydoo, line 497: What about `disallowedTools`? Any other missing fields?

src/modules/integrations/claude.nix
```diff
@@ -40,6 +40,29 @@ let
...
+          type = lib.types.bool;
+          default = false;
+          description = ''
+            Make the file executable, for a resource that is run directly
+            rather than passed to an interpreter.
+          '';
+        };
+      };
+    }
+  );
```
sandydoo: Might make sense to extract a reusable `fileSpecType` module from files.nix instead of reimplementing it. We need a way to access it and then import it in the submodule here.

sandydoo (changes requested):
Awesome! A few things to fix and we can merge.

src/modules/integrations/claude.nix
paulbertin: Good catch, I added an assertion for this https://github.com/cachix/devenv/pull/3120/changes/6a9203d0a607ae88435e4684e498a35aa44e4d9a

src/modules/integrations/claude.nix
paulbertin: Indeed! Fixed and added its `disallowedTools` counterpart

src/modules/integrations/claude.nix
paulbertin: I added 11 new options, so it exposes 14 of the 20 documented fields in the [frontmatter reference](https://code.claude.com/docs/en/skills#frontmatter-reference).

I intentionally omitted 6 of them (`hooks`, `metadata`, `license`, `compatibility`, `shell` and `paths`) so it's not too bloated. I tried to follow existing pattern with agents, curating the most relevant ones (and the one I could actually test working, e.g. `paths` looks a bit broken on CC 2.1.246). Happy to revisit

src/modules/integrations/claude.nix
paulbertin: Good call, I tried to refactor it in https://github.com/cachix/devenv/pull/3120/changes/82db93a229f482010a16f5805437f3035f5ed762. Let me know if that matches what you had in mind!

paulbertin:
Thanks for the great review @sandydoo, let me know if you'd like to reiterate on it

sandydoo (approved):
LGTM! Thanks!

PR #3131, opened 2026-08-29
Add FileChanged hook type and timeout option to claude.code.hooks

tarc:
## Summary
- Add `FileChanged` to `claude.code.hooks.<name>.hookType`, matching the documented Claude Code hook event that fires when a watched file changes on disk. For this event, `matcher` is a glob pattern (or `|`-separated globs) against project-relative paths rather than a tool-name regex.
- Add `claude.code.hooks.<name>.timeout` (seconds before Claude Code cancels the hook command), plumbed through into the generated `.claude/settings.json`, omitted when unset.
- Both verified against the official Claude Code hooks documentation before implementing.
- Update `docs/src/content/docs/integrations/claude-code.md` with the new hook type, the `timeout` field, and a `FileChanged` example.
- Add `tests/claude-hooks`, the first test coverage for `claude.code.hooks`, asserting the generated JSON via `jq` — no `claude` binary or live session required.

## Test plan
- [x] `devenv-run-tests run --only claude-hooks tests` passes
- [x] `nix-instantiate --parse` on the modified module


tarc:
For some context, I'm mapping [GSD](https://github.com/open-gsd/gsd-core) hooks - which it insists on saving in settings.local.json for non global configurations - into devenv's configuration. Those two pieces were missing: the `FileChanged` hook type and `timeout` property.

src/modules/integrations/claude.nix
```diff
@@ -133,10 +154,12 @@ let
         (hook: {
           matcher = hook.matcher or "";
           hooks = [
-            {
+            ({
               type = "command";
               command = hook.command;
-            }
+            } // lib.optionalAttrs (hook.timeout or null != null) {
```
sandydoo:
```suggestion
            } // lib.optionalAttrs (hook.timeout != null) {
```

src/modules/integrations/claude.nix
```diff
@@ -109,17 +110,37 @@ let
...
-        description = "Regex pattern to match against tool names (for PreToolUse/PostToolUse hooks).";
+        description = ''
+          For most hook types, a regex pattern to match against tool names
+          (for PreToolUse/PostToolUse hooks).
+
+          For `FileChanged`, this is instead a glob pattern (or `|`-separated
+          list of glob patterns) matched against file paths relative to the
+          project root, e.g. `.env` or `*.md` or `.env|.env.local`. Claude
+          Code expands the glob(s) into the set of files it watches, and
+          fires the hook whenever one of them changes.
```
sandydoo: https://code.claude.com/docs/en/hooks#filechanged is a little confusing, but iiuc it watches files by exact name, so `*.md` will look for a file named exactly that and not all markdown files. Needs looking into.

For regex to work you'd need to register `watchPaths` separately and then this would filter the event? 🤷

sandydoo (changes requested):

src/modules/integrations/claude.nix
tarc: Yes, good catch! That description doesn't fit at all the documentation, thank you!

That second item in the docs is really cryptic ("Filter which hooks run..."). I'll experiment with it and update the documentation.

tarc: So, @sandydoo , for `FileChanged` hooks the `matcher` serves those two roles as stated in the docs simultaneously:

> The matcher for this event serves two roles:
    - Build the watch list: the value is split on `|` and each segment is registered as a literal filename in the working directory, so `".envrc|.env"` watches exactly those two files. Regex patterns are not useful here: a value like `^\.env` would watch a file literally named `^\.env`.
    - Filter which hooks run: when a watched file changes, the same value filters which hook groups run using the standard [matcher rules](https://code.claude.com/docs/en/hooks#matcher-patterns) against the changed file’s basename.

So, for `FileChanged` hooks, `matcher` serves two purposes depending when they are used by Claude Code. The mental model is something like this:

- At setup time (once), every `FileChanged` hook contributes to form a watch set (Claude Code forms a union of all the `FileChanged` hooks matcher's). Here Claude is being strict about what each entry contributes to the watch set, allowing - at most - the `"|"`-split;
- At change time (every time a watched file actually changes), Claude Code re-evaluates every `FileChanged` entry's own matcher against that file basename. Here Claude is using the "matcher rules" (see above, or in the docs) to know if the entry matches the changed file.

Suppose, for instance, that we have only one `FileChanged` hook with matcher `".env|.env.local"`. When evaluated during _setup time_, according to the first item, Claude builds a two element set `{ ".env" ".env.local" }`. When one of those files is modified and Claude re-evaluates the matcher against this very file's basename, the matcher's `".env|.env.local"` expression is also valid according to the "matcher rules" and indeed matches the file (see the second item in the table of [Matcher patterns](https://code.claude.com/docs/en/hooks#matcher-patterns): "Exact string, or list of exact strings separated by | or , with optional surrounding whitespace").

So, similarly, we can set one `FileChanged` hook 
- with a generic matcher: `".env|.env.local|.env.production"`
- and another one, more specific: `"^\.env\..*$"`

But also:
- the generic one: `".env|.env.local|.env.production"`
- and the specific: `".env.production"`

Claude Code just handles the matchers at setup time as plain filenames (no regex, no glob etc.) and at change time "using the standard [matcher rules](https://code.claude.com/docs/en/hooks#matcher-patterns) against the changed file’s basename."

I tried to convey this in the description of the matcher and also added more examples (in the ./examples dir).

Please, let me know if this is clearer.

PR #3135, opened 2026-08-31
shell: preserve application Ctrl-Alt input in alternate screen

### LINKED ISSUE #3107: Ctrl Input grab by devenv shell with reload=true
**Describe the bug**
when I'm in a devenv shell with the reload active
my nvim input based on ctrl stop working
(i think this bug is not specific to nvim but i've not in mind an other cli that use ctrl)
after some test this only occured when the reload is on

**To reproduce**
```
mkdir /tmp/test && cd /tmp/test && devenv init && devenv shell && nvim
```

**Version**
devenv 2.2.2 (aarch64-darwin)

Arshgill01:
Fixes #3107

## Summary

- Keep reload-shell Ctrl-Alt shortcuts active at the shell prompt.
- Forward the same byte sequences to full-screen applications while they own the terminal's alternate screen, so editors such as Neovim receive their input.
- Add unit, session PTY, and real `devenv shell` regression coverage for forwarding `ESC Ctrl-D` unchanged.

## Root cause

The reload shell classified legacy Ctrl-Alt byte sequences before writing stdin to the child PTY. That classification remained active after a full-screen application entered the alternate screen, so application keybindings that produced the same bytes were consumed as devenv shortcuts.

## Validation

- [x] Repository pre-commit hooks (`nixfmt`, `rustfmt`)
- [x] `shellcheck tests/reload-alternate-screen-input/alternate-screen-reader.sh tests/reload-alternate-screen-input/.test.sh`
- [x] `nix build --accept-flake-config .#devenv --no-link --print-out-paths`
- [x] Targeted session PTY regression
- [x] Nix-built `devenv-run-tests run tests --only reload-alternate-screen-input` (real reload shell, exact `1b 04` bytes)
- [x] Remaining `devenv-shell` suite with two unrelated timing assertions skipped (188 library tests and 18 PTY tests passed)

The unfiltered suite also ran. The new regression passed; two existing status-line timing assertions remain sensitive to zero/one-millisecond elapsed values (`test_build_lifecycle_status_line` and `status_line_cache_invalidates_for_state_width_and_spinner_changes`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Bug Fixes
 
+- Fixed reload-shell shortcuts intercepting Ctrl-Alt input from full-screen terminal applications such as Neovim ([#3107](https://github.com/cachix/devenv/issues/3107)).
 - Fixed `showOutput = true` / `--show-output` being ignored when AI-agent auto-quiet (or `--quiet`) is active. Explicit task output now streams even at Quiet, without treating stdout as an error ([#3038](https://github.com/cachix/devenv/issues/3038)).
 - Fixed `devenv tasks run` skipping a process that a task depends on via `@completed` when that process has `start.enable = false`. The process now runs to completion as the dependency requires; `devenv up` still does not auto-start it ([#3005](https://github.com/cachix/devenv/issues/3005)).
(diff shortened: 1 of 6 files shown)


Arshgill01:
Friendly review ping — this keeps reload-shell Ctrl-Alt shortcuts at the prompt while forwarding the same byte sequences to full-screen apps on the alternate screen (e.g. Neovim), fixing #3107.

PR is mergeable against `main`. Happy to address review feedback.

PR #3141, opened 2026-09-01
feat: add opt-in localhost process proxy

domenkozar:
Adds an opt-in shared HTTP proxy for process URLs such as `http://web.demo.localhost`. Enable it with `process.proxy.enable = true`; routes use the project name, process name, and declared ports. The TUI shows each process's URL beside its port, and both processes and individual ports support hostname overrides.

The proxy reconciles each project's routes through a private Unix socket, supports IPv4 and IPv6 loopback servers, and shuts down promptly for restarts. Linux port-80 access uses the existing capability broker with terminal handoff for sudo; macOS uses a loopback-scoped prebound listener. The proxy is packaged alongside the CLI. This version serves HTTP and keeps routes in memory.

Validation:

- Cargo proxy tests cover IPv4/IPv6 fallback, POST bodies, unavailable upstreams, and daemon restarts.
- Cargo CLI tests cover route generation, URL serialization, sudo terminal handoff, and waiting for a previous daemon to exit.
- Activity and TUI tests cover serialization and displayed hostnames after integration with the current port and readiness types on `main`.
- Verified `http://docs.devenv8.localhost/` serves HTTP 200 through the Linux port-80 proxy.
- Native macOS port-80 behavior has not been tested locally.


venkatd:
Amazing!

One possible syntax proposal:
```
{ config, ... }: {
  process.proxy.enable = true;

  processes = {
    backend.ports.http.hostname = "api.stanfordtax--${config.git.branch}.localhost";
    frontend.ports.http.hostname = "app.stanfordtax--${config.git.branch}.localhost";
    clients-frontend.ports.http.hostname = "clients.stanfordtax--${config.git.branch}.localhost";
  };
}
```

I feel like an API that lets you explicitly set the hostname would be less magic (you can read what the host would look like) and more flexible.

For example, in our case, we mirror subdomains to match the structure of our production and staging applications as much as possible.

In this case, I think `config.git.branch` might need to be exposed?

Curious on your thoughts both good/bad

domenkozar:
Yeah, overriding the default should be definitely possible.

RafaelKr:
@domenkozar also multiple Domains per process pointing to the same unix domain socket using the PROXY protocol would be helpful. We got a multi-tenant application where the tenant is determined by domain.

venkatd:
@RafaelKr could you share an example of how this would look? Like what is an example setup of hosts including interpolation that you'd use?

My thinking is that we should just be able to write anything we want as long as the "TLD" (such as `.localhost`) is supported.

Then could have a simple way to support all hosts where you just need to derive the `.hostname` property and devenv resolves it

harrysbaraini:
I upvote for multiple hostnames for the same process and port. Hostname accepting a list of strings would already be helpful.

When replacing my Docker-based setup, one of my pain-points was getting a dynamic reverse proxy to do what Traefik was doing. I came up with hackish module that runs a single Caddy instance, which I import in my projects (https://github.com/vansdevcode/devenv-modules).

In a project's devenv.nix, I do something similar to caddy's module:

```nix
services.sharedProxy = {
  enable = true;
  virtualHosts."${config.env.APP_DOMAIN}.localhost" = {
    serverAliases = [ "tenant-a.${config.env.APP_DOMAIN}.localhost", "tenant-b.${config.env.APP_DOMAIN}.localhost" ];
    extraConfig = ''
        reverse_proxy 127.0.0.1:${toString config.processes.frankenphp.ports.http.value}
    '';
  };
};
```

domenkozar:
Could someone create a new issue?

With use case explained

harrysbaraini:
https://github.com/cachix/devenv/issues/3164

PR #3147, opened 2026-09-02
fix(tasks): run start.enable=false processes pulled in as @completed deps

### LINKED ISSUE #3005: process with `start.enable = false` process is skipped if ran via @completed task dependency
**Describe the bug**
When a task depends on a process via `@completed`, and that process has `start.enable = false`, running the task headlessly with `devenv tasks run` does **not** execute the process — yet the task reports success (exit 0). The process's `exec` never runs and no error is surfaced. Setting `start.enable = true` makes it work.

Per the option's own documentation, `start.enable` should only affect `devenv up` autostart — not whether a process runs when pulled in as a task dependency.

**Version**

`devenv 2.1.2+ea3d94a (aarch64-darwin)`

## To reproduce

https://gist.github.com/aflatter/5093ec95b1e645edc16b1a2ee6b22e39

```sh
rm -f process-ran.txt task-ran.txt
devenv tasks run --no-tui repro:build
ls process-ran.txt task-ran.txt
```

### Expected

Both `process-ran.txt` and `task-ran.txt` exist — the `@completed` dependency
runs the process to completion, then the task body runs.

### Actual

Only `task-ran.txt` exists. `process-ran.txt` is **missing** — the process's
`exec` never ran. The command prints `{}` and exits `0` (success).

Arshgill01:
Fixes #3005.

`start.enable = false` is documented as only controlling whether a process autostarts with `devenv up`. When a task depended on such a process via `@completed`, `devenv tasks run` skipped the process `exec` and still treated the dependency as satisfied, so the task exited 0 without the process ever running.

The Nix module already generated process tasks for disabled processes. The skip happened at launch time: a transient `devenv tasks run` registered those processes as waiting with `start.enable` still false, `launch_waiting` turned them into `NotStarted`, and `@completed` is vacuously true for `NotStarted`.

This matches the existing attach / `devenv processes start` path, which already overrides `start.enable` for explicitly requested starts (`start_with_deps`).

## Changes

- In the transient task runner (`devenv tasks run`), force-start processes that were **pulled in as dependencies**, even when `start.enable` is false, so `@completed` / `@ready` / `@started` wait for real execution.
- `devenv up` (process mode) still honours `start.enable` and leaves those processes stopped.
- Processes that are themselves roots of `devenv-tasks run` (flake-compat / containers / `config.ci`, which name every process as a root) still register as `NotStarted` when disabled.
- Document that disabled processes still run as task dependencies.
- Unit tests for the dependency, root, and process-mode paths, plus a Nix integration test matching the [repro gist](https://gist.github.com/aflatter/5093ec95b1e645edc16b1a2ee6b22e39).

## Test plan

- `cargo test -p devenv-tasks --lib -- task_run_starts_auto_start_off_process_for_completed_dep task_run_does_not_start_auto_start_off_process_when_it_is_the_root process_mode_does_not_start_auto_start_off_process`
- Integration test `tests/tasks-completed-start-enable-off`: `devenv tasks run --no-tui repro:build` creates both `process-ran.txt` and `task-ran.txt`


Arshgill01:
Friendly review ping — `devenv tasks run` now force-starts a `start.enable = false` process when it is pulled in as an `@completed` dependency (Fixes #3005). `devenv up` still leaves those processes stopped.

Happy to address review feedback.

Arshgill01:
Rebased onto latest `main` (`30058c52`). The only conflict was `CHANGELOG.md`: kept this PR’s #3005 entry and main’s #3149 fix. Task/process logic and tests applied cleanly.

PR #3150, opened 2026-09-03
fix: typo in "compared to flakes" table

quasi-coherent:
The guide with a "devenv vs. nix flake" comparison table mixes up which one operates in a pure evaluation mode.

### DIFF
--- a/docs/src/content/docs/guides/using-with-flakes.md
+++ b/docs/src/content/docs/guides/using-with-flakes.md
@@ -45,7 +45,7 @@ Consider using the Flake integration when:
 | Protection from garbage-collection | ✓ | ✗ |
 | Faster evaluation (lazy trees) | ✓ | ✗ |
 | Evaluation caching | ✓ | ✗ |
-| Pure evaluation | ✓ | ✗ (`impure` by default) |
+| Pure evaluation | ✗ (`impure` by default) | ✓ |
 | Cross-project references | ✓ | ✓ |
 | secretspec.dev | ✓ | ✗ |
 | Running processes when testing | ✓ | ✗ |


domenkozar:
Thanks for catching that the current wording is ambiguous. However, swapping the columns is not quite correct: the devenv CLI is pure by default (`impure` defaults to `false`, and the backend enables `pure-eval`).

Nix flakes also support pure evaluation, but devenv's flake integration normally requires `--no-pure-eval` so it can discover the project directory. Pure evaluation is possible when `devenv.root` is explicitly set to an absolute path, at the cost of portability.

I suggest:

```markdown
| Pure evaluation | ✓ (default) | ✓ (requires an absolute `devenv.root`) |
```

quasi-coherent:
Oh, I guess I'm confused by what this section of the docs is meant to show.  

The heading is "Choosing between devenv and Nix Flakes," which I took to mean "devenv versus Nix Flakes (on its own)" not "devenv versus devenv integration with Nix Flakes."  In the former case, it's evaluated in pure evaluation mode by default (like what the official docs [say](https://wiki.nixos.org/wiki/Flakes#Making_your_evaluations_pure)); in the latter interpretation, it's not, but because of `devenv.lib.mkShell` needing to know the working directory.  

I opened the PR only because I thought it was strictly comparing devenv versus plain Nix Flakes, since later in the page there's the note

> :::note[Why do I need to use the --no-pure-eval flag?] Flakes use "pure evaluation" by default, which prevents devenv from figuring out the environment its running in: for example, querying the working directory. The --no-pure-eval flag relaxes this restriction.
>
> An alternative, and less flexible, workaround is to override the devenv.root option to the absolute path to your project directory. This makes the flake non-portable between machines, but does allow the shell to be evaluated in pure mode. :::

which seems to directly contradict the row of the table we're talking about.

In that case, I don't know how this page should change, or if it should change at all.  So I can close this if you'd like.

PR #3146, opened 2026-09-02
Honor showOutput under Quiet / AI-agent auto-quiet

Arshgill01:
Fixes #3038

## Summary

- `showOutput = true` and `--show-output` now stream task stdout even when AI-agent auto-quiet (or `--quiet`) sets `VerbosityLevel::Quiet`.
- Quiet without `show_output` still suppresses successful task stdout.
- Stdout is not reclassified as `Error`; the console renderer treats `show_output` as an explicit per-task force flag.

## Root cause

AI-agent detection (`CLAUDECODE` and similar) selects `VerbosityLevel::Quiet`. `show_output` only promoted task logs from `Debug` to `Info`, and Quiet admits only `Error`, so an explicit `showOutput = true` was still filtered. Task stderr already used `Error`, which is why the original report looked like “stdout missing, stderr shown”.

This is the remaining #3038 item after maintainers narrowed it: CI without agent env is unaffected; failing-task replay on `main` was already fixed by ab87ccb3.

## Relationship to #3115 / #3145

#3115 is the Quiet `--no-tui` **lifecycle** gap (no running/succeeded lines until exit). #3145 has merged; this branch is rebased onto that `main`. Both live in `devenv/src/console.rs` and now compose: #3145’s `show_lifecycle` for start/end, this PR’s `force_output` for `show_output` logs. Quiet without `show_output` still suppresses task stdout.

## What changed

- `ConsoleOutput` stores `force_output` from `TaskInfo.show_output` alongside `show_lifecycle`.
- Task log lines with `force_output` bypass the verbosity gate via `write_visible`, still rendered as ordinary indented logs (not Error).
- Unit tests in `devenv/src/console.rs`:
  - `quiet_show_output_streams_task_stdout`
  - `quiet_without_show_output_suppresses_task_stdout`
  - `quiet_show_output_no_duplicate_on_fail`
- Integration coverage in `tests/tasks/.test.sh` for `--quiet` and `CLAUDECODE=1`.

## Validation

- [x] Confirmed no existing open PR already fixes #3038 (searched cachix/devenv for 3038 / showOutput + quiet)
- [x] Rebased onto current `cachix/devenv` main (`b537c844`)
- [x] `git diff --check`
- [ ] `cargo test -p devenv --lib quiet_show_output` — blocked here: this environment has no `nix-flake-c.pc` / Nix. CI on this PR should run the new unit tests and `tests/tasks`.

[...]

PR #3152, opened 2026-09-03
fix(tasks): reuse running process manager for process deps

### LINKED ISSUE #3137: Tasks starts processes even when they are already running
A task that has dependencies on processes always attempts to start them, even when they are already running via `devenv up` (or otherwise, potentially a different task). I would expect that the running process manager is detected and the state of the running processes is respected and the task is immediately run.

**To reproduce**

```nix
{

  tasks."test:repro" = {
    exec = "true";
    after = ["devenv:processes:postgres"];
  };
}
```

Run `devenv up -d; devenv tasks run test:repro`. This will lead to output roughly as follows:

```
✓ Validating lock 1 files                                                                                                                                                                                                                                                                                                                                                                                                                                                                          11ms
✓ Configuring cachix                                                                                                                                                                                                                                                                                                                                                                                                                                                                                5ms
✓ Configuring shell                                                                                                                                                                                                                                                                                                                                                                                                                                                                               657ms
  └ ✓ Evaluating shell 1849 files                                                                                                                                                                                                                                                                                                                                                                                                                                                                 657ms
✓ Loading tasks                                                                                                                                                                                                                                                                                                                                                                                                                                                                                     2ms
  └ ✓ Evaluating devenv.config.task.config                                                                                                                                                                                                                                                                                                                                                                                                                                                          1ms
✗ Running tasks 2/2                                                                                                                                                                                                                                                                                                                                                                                                                                                                               182ms
  └ ✗ test:repro failed                                                                                                                                                                                                                                                                                                                                                                                                                                                                             0ms
  └ ✗ postgres gave up (crash loop)
    │ Process exited (Failure), restarting
    │ Restarted (attempt 1)
    │ Process exited (Failure), restarting
    │ Restarted (attempt 2)
    │ Process exited (Failure), restarting
    │ Restarted (attempt 3)
    │ Process exited (Failure), restarting
    │ Restarted (attempt 4)
    │ initdb: error: cannot be run as root
    │ initdb: hint: Please log in (using, e.g., "su") as the (unprivileged) user that will own the server process.
    │ initdb: error: cannot be run as root
    │ initdb: hint: Please log in (using, e.g., "su") as the (unprivileged) user that will own the server process.
    │ initdb: error: cannot be run as root
    │ initdb: hint: Please log in (using, e.g., "su") as the (unprivileged) user that will own the server process.
    │ initdb: error: cannot be run as root
    │ initdb: hint: Please log in (using, e.g., "su") as the (unprivileged) user that will own the server process.
    │ Process exited (Failure), restarting
    │ Restarted (attempt 5)
    │ process exit: restart rate limit exceeded
  × Some tasks failed
```

**Version**

This is the case as of devenv 2.2.2.

Arshgill01:
Fixes #3137.

`devenv tasks run` started `devenv:processes:*` dependencies in a fresh local runner even when those processes were already healthy under `devenv up`. That second copy crash-looped services such as PostgreSQL (`initdb` as root).

When a process manager is already running, `tasks run` now:

1. Collects the process names the task graph would schedule
2. Asks the native manager to start only missing names (`Start` / `start_with_deps` skips Ready/Starting/Waiting)
3. Waits until those processes are ready
4. Sets `ignore_process_deps` so the local runner does not launch them again
5. Does not `stop_all` on exit, so the existing manager keeps ownership

If no manager is running, behavior is unchanged: process deps start locally and are stopped when the task graph finishes.

## Test plan

- [x] `cargo test -p devenv-tasks --lib scheduled_process_names`
- [x] `cargo fmt --all -- --check`
- [ ] Integration test `tests/tasks-run-attach-running` (`devenv up -d` then `devenv tasks run test:repro` must keep the dummy process pid and `starts == 1`)
- [ ] Manual: with `services.postgres.enable = true` and `after = ["devenv:processes:postgres"]`, run `devenv up -d; devenv tasks run test:repro` without a postgres crash loop


Arshgill01:
Rebased onto latest `main` (includes #3146 and #3135). CHANGELOG keeps this PR’s #3137 entry plus main’s newer entries. `behind_by` 0, mergeable.

Arshgill01:
Rebased this branch onto latest `cachix/devenv` `main` (`6e830506`). Conflicts were in CHANGELOG (moved the #3137 entry under 2.3.2 unreleased, kept newer main entries) and `docs/src/content/docs/tasks.mdx` (docs were converted from `.md`). Same head branch; no new PR.

PR #3156, opened 2026-09-04
fix: serialize cold eval/task cache init across processes

### LINKED ISSUE #3133: Concurrent `devenv shell` calls result in "Failed to initialize eval cache database: database is locked"
## Summary

`devenv shell`, as far as I can see, does not guard against other concurrent `devenv` processes also starting in the same directory. When two entries hit a project that has no `.devenv/` yet, one of them reliably exits non-zero:

```
× Failed to initialize eval cache database: Database error: error returned from database: (code: 5) database is locked
```

Once `.devenv/` exists, concurrent entries are fine — the race is at first-time creation.

## Reproduce

I've pushed a quick repro script at https://github.com/gaggle/devenv/commit/204d3f1226e2016e70c6f98afd23e72e18e0aa0f

```bash
% cd repro-cold-concurrent-entry 

% ./repro.sh 
cold .devenv — entering 2 shells at once
  shell 1   exit=0
  shell 2   exit=1

── shell 1 ──
hello from devenv
git version 2.55.0
── shell 2 ──
  × Failed to initialize eval cache database: Database error: error returned from database: (code: 5) database is locked

FAIL: 1 of 2 entries died; every entry should succeed

% devenv --version
devenv 2.2.2 (aarch64-darwin)
```

Notes:

- The eval cache isn't the only contended state: with 3+ concurrent entries I also saw `× Failed to initialize task cache`, and concurrent entry against a warm `.devenv/` in another project once produced `× Failed to remove existing GC root: No such file or directory (os error 2)`.
- The SQLite error varies, I've also seen `(code: 5898) disk I/O error`.
- A longer evaluation (e.g. `languages.elixir.enable = true;`) widens the window — more entries fail.

## Why this matters

I start multiple processes at the same time to run different commands inside the devenv environment, so they all enter the same project concurrently, and they have to get through the initial cold boot for that to work. Against a warm `.devenv/` everything is fine, it's the first launch in a fresh checkout where a process dies at startup.

## Expected

All concurrent entries succeed regardless of whether `.devenv/` exists yet, e.g. initialization takes a lock, or SQLite is opened with a busy timeout/WAL settings that tolerate a concurrent creator.

Arshgill01:
Fixes #3133.

Concurrent `devenv shell` entries into a project that does not yet have `.devenv/` raced on eval-cache (and task-cache) SQLite initialization. One process created the database; another failed with `Failed to initialize eval cache database: Database error: (code: 5) database is locked`. Under more contention the same window also produced task-cache init failures, `SQLITE_IOERR_DELETE_NOENT` (code 5898) after a loser deleted the DB, and `Failed to remove existing GC root: No such file or directory`. Warm `.devenv/` was already fine.

`busy_timeout(10s)` and WAL with DELETE fallback on `SQLITE_IOERR_SHMMAP` were not enough: `sqlite3_open(CREATE)` / `PRAGMA journal_mode=WAL` can return `SQLITE_BUSY` before the timeout is installed, and a migration error used to delete the database out from under the process that held it.

## Changes

- Take an exclusive flock on a sibling of the SQLite file (`*.db.init.lock`, not the `.db` itself) around create + migrate in `Database::new`, so eval cache and task cache share the same cold-init serialization.
- Hold that lock on a dedicated OS thread so waiters do not stall the tokio runtime.
- Never delete the database on a busy/locked migration error; retry instead.
- Treat a missing GC-root symlink as success when replacing it, so two shells creating the same root no longer lose the TOCTOU race.
- Regression tests: concurrent cold/warm `Database::new`, concurrent cold `TaskCache::new`, and `tests/cold-concurrent-shell` (two concurrent `devenv shell -q -- true` against a missing `.devenv/`, then the same against a warm one).

## Test plan

- [x] `cargo test -p devenv-cache-core` (includes `concurrent_cold_open_succeeds` / `concurrent_warm_open_succeeds`)
- [x] `cargo test -p devenv-tasks --lib -- test_cold_cache_allows_concurrent_open test_warm_cache_allows_concurrent_open`
- [x] `cargo clippy -p devenv-cache-core --all-targets -- -D warnings`
- [ ] `devenv-run-tests run tests --only cold-concurrent-shell` (needs a full devenv/Nix environment; not run in this agent VM)


Arshgill01:
Rebased onto latest `main` (`behind_by: 0`, mergeable). Kept the #3133 changelog entry with main’s newer 2.3.0/2.3.1 notes. Preserved flock around `Database::new` and GC-root `NotFound`-as-success on `ensure_gc_root`.

Arshgill01:
Rebased onto latest `main` (`behind_by: 0`, mergeable). Kept the #3133 changelog entry with main’s newer 2.3.1/2.3.2 notes. Preserved flock around `Database::new` and GC-root `NotFound`-as-success on `ensure_gc_root`.

PR #3166, opened 2026-09-08
fix: keep PATH exports on their own line when assembling shell-env.sh

### LINKED ISSUE #3165: shell-env.sh glues venv task exports onto `eval "${shellHook:-}"` — PATH values with spaces word-split (`      bash: export: ... not a valid identifier`)
**Describe the bug**
When a devenv project enables the Python venv (so activation has to re-prepend `venv/bin` to `PATH`) **and** the parent shell's `PATH` contains any directory with a space (fairly common on macOS: `~/Applications/Home Manager Apps/...`, `.../Ghostty.app/Contents/MacOS`, etc.), starting a devenv shell prints a batch of `bash: export: ... not a valid identifier` errors and `PATH` inside the shell is **truncated at the first space**, silently dropping system directories.

At that point `PATH` is corrupted for the whole devenv shell session.

**To reproduce**
```console
$ cat devenv.yaml            # minimal
inputs:
  nixpkgs:
$ cat devenv.nix
{ pkgs, ... }: {
  languages.python = {
    enable = true;
    package = pkgs.python312;
    venv.enable = true;
    uv = { enable = true; sync.enable = true; };
  };
}
```

Then launch with a space in PATH (any space-containing dir reproduces; below uses a fake one so it's machine-independent):

```console
$ mkdir -p "/tmp/Some App/bin"
$ PATH="/tmp/Some App/bin:$PATH" devenv shell
...
bash: export: `Apps/:/etc/profiles/per-user/<user>/bin:/run/current-system/sw/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin:/Users/<user>/Applications/Home': not a valid identifier
bash: export: `Apps/Some App/bin': not a valid identifier
```

Note the artifacts reference the real-world path with spaces (`.../Applications/Home Manager Apps/...`). Inside the shell, `echo "$PATH"` shows everything from the first space onward is gone.

Without a venv (e.g. a devenv that only has `packages`), the same shell starts cleanly — the errors only appear when the environment includes Python venv activation.

## Root cause

`.devenv/shell-env.sh` (regenerated on every devenv invocation) ends with the venv's final exports glued onto the last line of the environment script:

```bash
eval "${shellHook:-}"export PATH='/Users/<user>/.../.devenv/state/venv/bin:...:/Users/<user>/Applications/Home Manager Apps/Ghostty.app/Contents/MacOS'
export VIRTUAL_ENV=/Users/<user>/.../.devenv/state/venv
```

The first export is on the **same physical line** as `eval "${shellHook:-}"` with no newline. When bash parses that line, the single quotes around the `PATH=` value delimit the word for the *outer* parse — so by the time `eval` re-parses the text, the assignment is unquoted:

```bash
eval "…" export PATH=/Users/<user>/…/Home Manager Apps/…   # quotes already consumed
```

`eval` then word-splits the PATH value at the spaces → `export: 'Apps/…': not a valid identifier` and PATH is truncated at the first space.

In the assembly code (`devenv/src/devenv/mod.rs`, `prepare_shell` — the "interactive / non-bash" branch that writes `shell-env.sh`), the env script is trimmed and task exports are appended with no separator:

```rust
let shell_env = output_str
    .trim_end()
    .trim_end_matches("exec bash")
    .trim_end_matches("exec $SHELL");
// …
let mut env_content = shell_env;
env_content.push_str(&task_exports);   // first export lands on the eval line
env_content.push_str(&task_messages);
std::fs::write(&env_script_path, &env_content)
```

`format_shell_exports` itself quotes correctly (`export PATH='…'`), and the trailing exports (`export VIRTUAL_ENV=…`) appear on their own lines — it is only the **first** export, glued to the eval line with its quotes consumed by the outer parse, that breaks.

Two projects on the same devenv 2.2.2 make the trigger clear:

- Project A (no Python): env script ends `… \neval "${shellHook:-}"\n` — nothing appended → no error.
- Project B (Python + venv): env script ends `eval "${shellHook:-}"export PATH='…'` (glued) → error whenever PATH contains a space.

I checked v2.3's `devenv/src/devenv/mod.rs` — the assembly (trim + `push_str(&task_exports)` with no separator) is identical, so v2.3 likely still reproduces this.

**Version**
devenv 2.2.2 (aarch64-darwin, nix-darwin + home-manager); also verified the assembly is unchanged in v2.3 (2026-09-07), so likely still reproducible there.

Arshgill01:
Fixes #3165.

When Python venv (or any enterShell task) re-exports `PATH` **and** the parent `PATH` contains a directory with a space, `devenv shell` printed `bash: export: … not a valid identifier` and truncated `PATH` at the first space.

`prepare_shell` trims the Nix env script with `trim_end()`, which removes the newline after `eval "${shellHook:-}"`. Task exports were then appended with no separator, so the first `export PATH='…'` landed on the same physical line. Outer bash consumed the quotes; `eval` word-split the PATH value at spaces.

## Changes

- Add `push_shell_fragment`, which inserts a newline before appending a non-empty fragment.
- Use it wherever the env script is concatenated with task exports/messages (`prepare_shell`, PTY/reload `shell-env.sh`, direnv export).
- Empty fragments remain a no-op, so projects without venv/task exports are unchanged.

## Tests

- Unit: assemble a trimmed env script plus a PATH export containing `Some App`, assert the export is on its own line, then `bash` the result and check PATH is not truncated.
- Integration: `tests/shell-path-spaces` — PATH with a space + a venv-like enterShell PATH export; `--shell zsh` writes `.devenv/shell-env.sh` (the glued assembly) and `devenv shell --` covers the non-interactive path.

No competing open PR for #3165.

## Test plan

- [x] Standalone reproduction: glued `eval "${shellHook:-}"export PATH='…Some App…'` produces `not a valid identifier` and truncates PATH; with a newline separator, bash sources cleanly and PATH keeps the space-containing dir.
- [ ] `cargo nextest run -p devenv trimmed_shell_env push_shell_fragment`
- [ ] `devenv-run-tests run tests --only shell-path-spaces`
- [ ] Manual: `PATH="/tmp/Some App/bin:$PATH" devenv shell` in a project with `languages.python.venv.enable = true` — no export errors, `echo "$PATH"` still contains `Some App`.


Arshgill01:
Rebased onto latest `cachix/devenv` `main` (was ~13 commits behind). The `push_shell_fragment` / PATH-with-spaces fix is unchanged; the changelog entry now lives under **2.3.2 (unreleased)** so it stays with the current unreleased notes rather than the released 2.3.1 section.

Arshgill01:
### Still ready for review (bundled)

Gentle weekly nudge — still no maintainer feedback on the open Arshgill01 set. All MERGEABLE (rebasing onto latest `main` now if needed):

| PR | Summary |
| --- | --- |
| https://github.com/cachix/devenv/pull/3166 | PATH spaces in `shell-env.sh` (Fixes #3165) |
| https://github.com/cachix/devenv/pull/3156 | GC-root concurrent `NotFound` TOCTOU leftover from #3133 |
| https://github.com/cachix/devenv/pull/3152 | Reuse running process manager for process deps (Fixes #3137) |
| https://github.com/cachix/devenv/pull/2978 | process-compose `list` / `wait` lifecycle commands |

Happy to address review feedback on any of them.

PR #3167, opened 2026-09-08
fix(cli): load YAML from remote --from sources

rafaelfranca:
## Summary

- Materialize non-path `--from` sources before the final configuration load.
- Load remote `devenv.yaml`, `devenv.local.yaml`, YAML imports, YAML inputs, and Nix modules.
- Keep `path:` sources live, and keep target configuration at higher precedence.
- Use a virtual source-only lock, and retain the original source in the final lock.
- Add resolver unit tests and a network-free `git+file://` integration test.
- Add a dedicated out-of-tree devenvs guide and update related links.

## Use case

I keep the shared Rails configuration in [`rafaelfranca/devenv`](https://github.com/rafaelfranca/devenv/tree/main/rails/rails).

Each Rails checkout runs this command once:

```console
devenv --from 'github:rafaelfranca/devenv?dir=rails/rails' allow
```

Normal `devenv` commands then use the saved binding.
This change removes the local clone and temporary YAML copy from my current helper.

## Related work

[#2940](https://github.com/cachix/devenv/pull/2940) added persistent out-of-tree `--from` bindings.

Commit [`0497317e`](https://github.com/cachix/devenv/commit/0497317e622da8d1152285c69f42e3e2da489c8c) added source YAML support for live `path:` references.

Commit [`0999a140`](https://github.com/cachix/devenv/commit/0999a1409be16ec3333e6d7bd1f1458dc3b35319) improved out-of-tree source resolution and trust checks.

[#3055](https://github.com/cachix/devenv/pull/3055) adds remote YAML composition through input-style imports such as `shared-config/profiles/backend`.

PR #3055 still limits non-path `--from` sources to `devenv.nix`.
This pull request covers that separate entry point.

Both changes need virtual locking and source materialization.
The common resolver parts should be shared if both changes merge.

## Verification

- `cargo test -p devenv-nix-backend`: 56 passed.
- `cargo test -p devenv-core source_`: 4 passed.
- The focused source resolver tests: 4 passed.
- Direct and persisted `git+file://` source checks passed.
- The live `path:` update check passed.
- The public GitHub source and persisted binding checks passed.
- `cargo fmt --check`, `cargo clippy`, and the final debug build passed.
- `astro check`: no errors, warnings, or hints.
- `astro build`: 205 pages built, including `/guides/out-of-tree-devenvs/`.
- The documentation URL validator resolved all 201 legacy URLs.
- Browser checks confirmed the guide layout and HTTP 200 responses for its internal links.

## Known test failure

`proxy::tests::waits_for_previous_proxy_to_release_control_socket` failed twice with `EWOULDBLOCK`.
This pull request does not change proxy code.

### DIFF
--- a/devenv-nix-backend/src/source.rs
+++ b/devenv-nix-backend/src/source.rs
@@ -9,12 +9,28 @@ use serde_json::Value;
 use crate::anyhow_ext::AnyhowToMiette;
 use crate::{backend, lock};
 
+#[derive(Debug, PartialEq, Eq)]
+pub struct MaterializedSource {
+    repository_root: PathBuf,
+    directory: Option<PathBuf>,
+}
+
+impl MaterializedSource {
+    pub fn repository_root(&self) -> &Path {
+        &self.repository_root
+    }
+
+    pub fn directory(&self) -> &Path {
+        self.directory.as_deref().unwrap_or(&self.repository_root)
+    }
+}
+
 pub fn materialize_source(
     nix_settings: &NixSettings,
     root: &Path,
     lock_file_path: &Path,
     source_input: &Input,
-) -> Result<PathBuf> {
+) -> Result<MaterializedSource> {
     let store_settings = StoreSettings::default();
     let _gc_registration = backend::init_nix(nix_settings, &store_settings)
         .wrap_err("Failed to initialize Nix for source resolution")?;
@@ -22,8 +38,9 @@ pub fn materialize_source(
         .wrap_err("Failed to open the Nix store for source resolution")?;
     let (flake_settings, fetchers_settings) =
         backend::build_settings().wrap_err("Failed to build Nix source settings")?;
-    let mut eval_state = lock::build_eval_state(&store, root, &flake_settings)
-        .wrap_err("Failed to build the Nix source evaluator")?;
+    let mut eval_state =
+        lock::build_eval_state(&store, root, &flake_settings, nix_settings.refresh_fetchers)
+            .wrap_err("Failed to build the Nix source evaluator")?;
 
     let source_lock = crate::lock_source_input(
         &eval_state,
@@ -109,7 +126,7 @@ fn locked_source_attributes(lock_json: &str) -> Result<(Value, Option<String>)> 
     Ok((Value::Object(locked), dir))
 }
 
-fn select_source_directory(fetched_root: &Path, dir: Option<&str>) -> Result<PathBuf> {
+fn select_source_directory(fetched_root: &Path, dir: Option<&str>) -> Result<MaterializedSource> {
     let fetched_root = fs::canonicalize(fetched_root)
         .into_diagnostic()
         .wrap_err_with(|| {
@@ -119,7 +136,10 @@ fn select_source_directory(fetched_root: &Path, dir: Option<&str>) -> Result<Pat
             )
         })?;
     let Some(dir) = dir else {
-        return Ok(fetched_root);
+        return Ok(MaterializedSource {
+            repository_root: fetched_root,
+            directory: None,
+        });
     };
 
     let mut relative_dir = PathBuf::new();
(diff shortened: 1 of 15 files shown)


devenv-nix-backend/src/source.rs
```diff
@@ -0,0 +1,221 @@
...
+) -> Result<PathBuf> {
+    let store_settings = StoreSettings::default();
+    let _gc_registration = backend::init_nix(nix_settings, &store_settings)
+        .wrap_err("Failed to initialize Nix for source resolution")?;
+    let store = backend::open_store(&store_settings)
+        .wrap_err("Failed to open the Nix store for source resolution")?;
+    let (flake_settings, fetchers_settings) =
+        backend::build_settings().wrap_err("Failed to build Nix source settings")?;
+    let mut eval_state = lock::build_eval_state(&store, root, &flake_settings)
+        .wrap_err("Failed to build the Nix source evaluator")?;
```
domenkozar: [P2] Disable the fetcher TTL when materializing an updated source

When an upstream source changes within Nix's fetcher TTL, this evaluator still loads the cached revision: `refresh_fetchers` reaches the locker, but `lock::build_eval_state` never sets `tarball-ttl=0`. The final updater does disable that TTL, so it locks the new source using inputs read from the old YAML. Reproducing with an HTTP tarball left the old dependency URL in `devenv.lock` after `update from`; the next evaluation changed it again. Apply the refresh override to this evaluator's fetcher settings too.

devenv/src/main.rs
```diff
@@ -534,8 +536,36 @@ fn prepare_command(mut cli: Cli, shell_hint: Option<&str>) -> Result<PreparedCom
...
+        })
+    } else {
+        None
+    };
+
+    let mut config = if let Some(source_input) = remote_source_input.as_ref() {
+        let target_config = Config::load()?;
+        let mut preliminary_nix_settings =
+            NixSettings::resolve(nix_options.clone(), &target_config);
+        preliminary_nix_settings.refresh_fetchers = refresh_fetchers;
```
domenkozar: [P2] Keep from pinned when updating another input

For `devenv update nixpkgs`, this flag also unlocks `from` during source materialization, although the final updater leaves `from` pinned. With a fresh or expired fetcher cache, configuration therefore comes from an unrequested newer revision. This can change unrelated dependencies or prevent the update entirely: a newer source declaring `require_version: ">=999.0"` caused `update nixpkgs` to fail despite the locked source having no such requirement. Only unlock the source for a full update or an explicit update of `from`.

devenv/src/main.rs
```diff
@@ -534,8 +536,36 @@ fn prepare_command(mut cli: Cli, shell_hint: Option<&str>) -> Result<PreparedCom
...
+        let target_root = env::current_dir()
+            .into_diagnostic()
+            .wrap_err("Failed to resolve the target directory")?;
+        let source_path = devenv_nix_backend::source::materialize_source(
+            &preliminary_nix_settings,
+            &target_root,
+            &target_root.join("devenv.lock"),
+            source_input,
+        )?;
+        Config::load_with_source(Some(&source_path))?
```
domenkozar: [P2] Preserve the fetched repository root for YAML imports

Normal fetched Git trees lack `.git`, so `Config::load_with_source` cannot discover their repository root. A source selected with `?dir=profiles/rails` whose YAML imports `../common` consequently fails with “resolves outside the base directory,” even though the import stays inside the fetched repository; Git-root-relative imports fail too. Carry the fetched repository root alongside the selected directory and use it explicitly for import resolution and containment checks.

rafaelfranca: Fixed in 338c6778. Materialization now keeps the fetched root and the selected directory. Configuration loading uses that root for import resolution and containment.

devenv/src/main.rs
rafaelfranca: Fixed in 338c6778. Only a full update or `devenv update from` refreshes the source. The CLI regression covers an unrelated input update.

devenv-nix-backend/src/source.rs
rafaelfranca: Fixed in 338c6778. The source evaluator now applies `tarball-ttl=0` through its own `EvalStateBuilder`. A regression test verifies cache isolation between evaluators.

Issue #2465, opened 2026-02-04
Feature request: Cross-process task locking to prevent parallel execution races


schickling:
## Problem

When multiple terminals/processes trigger direnv simultaneously, the same devenv tasks can run in parallel across processes. This is common in setups like:

- **Conductor** running multiple Claude Code sessions that each trigger shell entry
- **Megarepo** setups where a parent repo and nested repos each have their own devenv
- Multiple terminal windows/tabs opening simultaneously

This causes several issues:
- **Duplicate network downloads** - Multiple `pnpm install` or `nix` fetches happening concurrently
- **Race conditions in file writes** - Concurrent writes to `node_modules`, lock files, or generated configs
- **Wasted CPU/bandwidth** - Redundant work that could be avoided

## Current Behavior

devenv tasks support parallelism **within** a single invocation (tasks can depend on each other and run concurrently where dependencies allow), but there is no coordination **across** multiple concurrent devenv invocations.

## Proposed Solution

Add optional task-level locking via a new attribute:

```nix
tasks."pnpm:install" = {
  exec = "pnpm install";
  lock = true;  # Acquire exclusive lock before running
};

# Or with named locks for grouping related tasks:
tasks."pnpm:install" = {
  exec = "pnpm install";
  lock = "package-manager";  # Named lock shared across tasks
};
```

When `lock` is enabled, devenv would:
1. Acquire an exclusive file lock (e.g., in `$DEVENV_STATE` or `/tmp`) before running the task
2. Other processes attempting the same task would either:
   - **Wait** for the lock to be released, then check if work is still needed
   - **Skip** if a flag indicates the task was just completed by another process

The lock file could be based on a hash of the devenv root + task name to ensure uniqueness.

## Workaround

Users can currently implement this manually using `flock` in task exec scripts:

```nix
tasks."pnpm:install" = {
  exec = ''
    exec 200>/tmp/devenv-pnpm-install.lock
    flock -x 200
    pnpm install
  '';
};
```

However, native support would be:
- Cleaner and more discoverable
- Consistent across all devenv users
- Able to integrate with devenv's existing state management

## Use Case

In a megarepo setup:
1. Parent repo has `devenv.nix` with tasks
2. Nested repos each have their own `devenv.nix`
3. Opening a shell in the parent can trigger direnv in nested repos
4. Running commands in nested repos can re-trigger the parent's direnv
5. Multiple terminals or automated tools (Conductor, CI) compound this

[...]

PR #3173, opened 2026-09-09
fix(files): order treefmt after devenv:files, copy atomically

### LINKED ISSUE #3172: `treefmt` sweep races `devenv:files` copy-mode rewrite: failed to stat
## Environment

- devenv 2.2.2, x86_64-linux (NixOS on WSL2)
- `treefmt.enable = true`
- a git-tracked `files.".qr-exclude.yaml".copyMode = "copy"` entry, matched by yamlfmt

## What happens

`devenv shell` intermittently fails during the enter-shell task wave with:

```text
devenv:treefmt:run failed → Error: failed to finalise formatting: failed to wait for formatters:
failed to stat <project>/.qr-exclude.yaml: no such file or directory
```

The failure is timing-dependent. In one failing run, `devenv:files` finished in 58 ms and
`devenv:treefmt:run` in 35 ms in the same task wave, so the sweep can stat the tree before
`devenv:files` finishes rewriting `.qr-exclude.yaml`. The project also runs a 4-second hot-reload
loop on shell entry, so the same race fires repeatedly in one session.

## Why

- `tasks."devenv:treefmt:run"` and `tasks."devenv:files"` both declare only
  `before = ["devenv:enterShell"]`, with no edge between them.
- devenv-tasks only adds a dependency edge for names listed in `after`/`before`, so the two
  edge-less tasks run concurrently as siblings.
- `copyMode = "copy"` materializes the destination as `rm -rf` followed by `cp -RL`, leaving the
  path briefly absent while it is rewritten.

## Minimal reproduction sketch

```nix
{
  treefmt.enable = true;
  treefmt.config.programs.yamlfmt.enable = true;
  files."managed.yaml" = { text = "a: 1\n"; copyMode = "copy"; };
}
```

Enter the shell repeatedly; failure is timing-dependent.

## Expected

`devenv:treefmt:run` should not observe a `copyMode = "copy"` file mid-rewrite, either by waiting
for `devenv:files` or by making the copy-mode write itself atomic.

## Workaround

Add to the project's own `devenv.nix`: `tasks."devenv:treefmt:run".after = ["devenv:files"];`

## Related

- #2387 gave `devenv:treefmt:run` its `before = ["devenv:enterShell"]`, making it a sibling of
  `devenv:files` in the schedule.
- #2948 added `copyMode = "copy"` along with the `rm -rf` + `cp -RL` sequence above.
- #2465 (cross-process task locking) is adjacent but not the same: a lock between separate
  `devenv` invocations, not an ordering edge between two tasks in one run.

mjwvolue:
## Problem

Interactive `devenv shell` intermittently fails when `treefmt.enable = true` and the project has
files managed by `files.<name>` with `copyMode = "copy"`:

```text
devenv:treefmt:run failed → Error: failed to finalise formatting: failed to wait for formatters:
failed to stat <project>/<file>: no such file or directory
```

`<file>` is a git-tracked, copy-mode file matched by a treefmt formatter (real case:
`.qr-exclude.yaml` under yamlfmt).

## Cause

- `tasks."devenv:treefmt:run"` (`src/modules/integrations/treefmt.nix`) and
  `tasks."devenv:files"` (`src/modules/files.nix`) both declared only
  `before = [ "devenv:enterShell" ]`, with no edge between them.
- `devenv-tasks` (`devenv-tasks/src/tasks.rs`) only adds a dependency edge for names listed in
  `after`/`before`; two edge-less tasks are spawned concurrently.
- `copyMode = "copy"` materialized a file as `rm -rf` followed by `cp -RL`, so the destination was
  briefly absent while the tree-wide treefmt sweep concurrently walked the project and tried to
  stat it.

## Fix

```text
CHANGELOG.md                            |  4 ++
docs/src/content/docs/creating-files.md |  2 +-
src/modules/files.nix                   | 68 ++++++++++++++++++++++-----
src/modules/integrations/treefmt.nix    |  6 +++
tests/files-treefmt-order/.gitignore    |  5 +++
tests/files-treefmt-order/.patch.sh     | 20 ++++++++++
tests/files-treefmt-order/devenv.nix    | 99 +++++++++++++++++++++++++++++++++++++++
tests/files-treefmt-order/devenv.yaml   |  6 +++
8 files changed, 197 insertions(+), 13 deletions(-)
```

- `src/modules/integrations/treefmt.nix`: add `after = [ "devenv:files" ]` to
  `tasks."devenv:treefmt:run"`. `devenv:files` is always declared (`files.nix` only wraps its
  *contents* in `optionalAttrs`), and an unresolved name in `after` is a hard error
  (`TasksNotFound`), so this is safe regardless of whether `files.*` is used.
- `src/modules/files.nix`: extract a `writeCopyScript` for copy-mode materialization; reword the
  `copyMode` option's own description to mention atomic replacement (regenerates
  `docs/src/data/options.json` via CI, not hand-edited here). A regular file is staged via
  `mktemp -d` beside the destination and `mv -f`'d into place: an existing file is replaced
  atomically, and a destination symlink is replaced rather than followed, including one
  pointing at a directory, which `mv` would otherwise resolve and drop the copy inside (that
  case's destination is briefly absent, never a plain file or an existing managed copy). Source
  directories keep the previous remove-then-copy path (`mv` onto an existing directory moves
  into it; `mv -T` is GNU-only).
- `docs/src/content/docs/creating-files.md`: note the atomic replacement for `copyMode = "copy"`.
- `CHANGELOG.md`: entry under `## 2.3.1 (unreleased)` / `### Bug Fixes`.
- `tests/files-treefmt-order/`: new `devenv-run-tests` case. `devenv.nix` declares three
  `copyMode = "copy"` files and asserts, via `$DEVENV_TASK_FILE`, that `devenv:treefmt:run` has
  exactly one `after` edge to `devenv:files`. `.patch.sh` pre-seeds three destinations: a plain
  file with a user edit (overwrite), a symlink elsewhere (replaced, not followed), and a symlink
  to a decoy directory (replaced; directory stays empty). The test checks no
  `.<name>.XXXXXX` staging directory is left behind for any of them.

## Behaviour changes / compat notes

- Log output: overwriting an existing `copyMode = "copy"` file now logs only `Overwriting <file>`,
  not also `Creating <file>` right after (old code removed the file first, then hit that branch).
- `copyMode = "seed"` over a *dangling* symlink now replaces the link instead of aborting: the old
  `cp -RL` wrote through the symlink and failed with "not writing through dangling symlink".
  `seed` shares `writeCopyScript` with `copy`, so it inherits the fix rather than regressing.
- Not verified on macOS/BSD: `mv -f` replacing a symlink, `mktemp -d TEMPLATE` beside the
  destination. Documented as portable, only exercised on Linux here.

### Notes

- No `trap` around the staging directory: a shell killed between `mktemp -d` and `rmdir` leaves a
  hidden `.<name>.XXXXXX/` dir behind. That matches the current signal-handling bar elsewhere in
  `devenv:files`, not a new gap.
- The new edge re-serializes `devenv:treefmt:run` behind `devenv:files`: a small, unavoidable loss
  of parallelism at shell entry.
- `devenv:gitnr:install` also writes into the project root with no edge to treefmt: same class of
  hazard, out of scope here.

## Testing

- `devenv-run-tests run tests --only files --only files-cleanup --only files-copy
  --only files-treefmt-order --only treefmt` → 5 passed (re-run after the symlink-to-directory
  fix and third test destination were added).
- Negative controls: reverting the `after` edge, and separately removing the symlink-to-directory
  unlink guard, both make `tests/files-treefmt-order` fail as expected.
- Separate 12-scenario old-vs-new harness: source mode (0444/0555) × destination state (missing,
  regular file, symlink-to-file, symlink-to-directory, dangling symlink, real directory): 0
  differences in resulting type, mode, or content.
- Stress: 300 replacements under a concurrent `test -e` poller: 0 "missing" observations with the
  new staged-rename sequence vs. 129 784 with the old `rm -rf` + `cp -RL` sequence.
- `$DEVENV_TASK_FILE` (read by the test via `jq`) is a real variable exported by
  `src/modules/tasks.nix` and read by `devenv-tasks/src/main.rs`, but undocumented elsewhere.
- `prek` (nixfmt, nixpkgs-fmt) passes on the changed files.

## Related

- Fixes: #3172
- #2387 ("treefmt: make treefmt run at enter shell") gave the treefmt task its
  `before = ["devenv:enterShell"]`, making it a scheduling sibling of `devenv:files`. This PR adds
  the missing edge between the two siblings.
- #2948 ("feat(files): add copy mode for editable files") introduced the non-atomic
  `rm -rf` + `cp -RL` sequence this PR replaces.
- `tasks."devenv:git-hooks:install"` already declares `after = [ "devenv:files" ]` for the same
  class of reason.
- #2465 (cross-process task locking) is an adjacent but distinct problem (a lock between separate
  `devenv` invocations, not an ordering edge between two tasks in the same run), and is not touched
  or duplicated here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 2.3.1 (unreleased)
 
+### Bug Fixes
+
+- Fixed `devenv shell` intermittently failing with a treefmt error such as `failed to stat <file>: no such file or directory` when `treefmt.enable = true` and the project has files managed by devenv. The tree-wide formatter now runs after devenv has written those files, and a file with `copyMode = "copy"` is replaced atomically, so tools that walk the project never see it missing.
+
 ## 2.3.0 (2026-09-07)
 
 ### Bug Fixes
--- a/docs/src/content/docs/creating-files.md
+++ b/docs/src/content/docs/creating-files.md
@@ -156,7 +156,7 @@ The `copyMode` attribute accepts:
 
 - `symlink` (default): symlink to the read-only file in the Nix store. Edits are not possible; devenv keeps the link pointed at the current contents.
 - `seed`: copy the file into place once, only if it does not already exist, and make it writable. Existing files are left untouched, so your edits are preserved. This is useful for seeding configuration files from templates that the user can then edit to fit their project.
-- `copy`: copy the file into place as a writable file, overwriting it with fresh contents on every shell entry. This is useful when a tool must write to the file in place but devenv should remain the source of truth.
+- `copy`: copy the file into place as a writable file, overwriting it with fresh contents on every shell entry. The file is replaced atomically, by renaming a temporary file in the same directory over it, so tools that walk the project never see it missing. This is useful when a tool must write to the file in place but devenv should remain the source of truth.
 
 :::tip[New in version 2.2]
 :::
--- a/src/modules/files.nix
+++ b/src/modules/files.nix
@@ -1,7 +1,7 @@
 { pkgs, lib, config, ... }:
 
 let
-  inherit (builtins) dirOf mapAttrs;
+  inherit (builtins) baseNameOf dirOf mapAttrs;
   inherit (lib) types optionalAttrs optionalString mkOption attrNames filter length mapAttrsToList concatStringsSep head assertMsg;
   inherit (types) attrsOf submodule;
 
@@ -101,7 +101,7 @@ let
 
         - `symlink` (default): symlink to the read-only file in the Nix store. Edits are not possible; devenv keeps the link pointed at the current contents.
         - `seed`: copy the file into place once, only if it does not already exist, and make it writable. Existing files are left untouched, so your edits are preserved. Useful for seeding configuration from templates the user then edits.
-        - `copy`: copy the file into place as a writable file, overwriting it with fresh contents on every shell entry. Useful when a tool must write to the file in place but devenv should remain the source of truth.
+        - `copy`: copy the file into place as a writable file, overwriting it with fresh contents on every shell entry. The file is replaced atomically, so tools that walk the project never see it missing. Useful when a tool must write to the file in place but devenv should remain the source of truth.
       '';
     };
   };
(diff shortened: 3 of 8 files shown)


mjwvolue:
@domenkozar Hi, any reason you've started separate work on this @ https://github.com/cachix/devenv/pull/3181 cherry-picking some parts from this PR?
Is there any technical issue with my PR, or did I fail some formal requirement for PR submission?
Would like to know what to do better/different in the future.
Best regards.

PR #2318, opened 2025-11-30
Update latest version to 1.11.2

onsails:
Probably fixes #2308 (see latest comments)

PR #3185, opened 2026-09-13
chore: bump latest-version to 2.3.1

zenoli:
v2.3.1 has been released (https://github.com/cachix/devenv/releases/tag/v2.3.1) but \`src/modules/latest-version\` was never bumped, so \`devenv update\` doesn't clear the "devenv X is newer than devenv input" warning even after syncing devenv.lock.

Bumps \`src/modules/latest-version\` from 2.2.2 to 2.3.1, following the same pattern as #1278 / #1183 / #2318.


domenkozar:
Bumped

